PR #255, opened 2026-01-06
feat: introduce reading resources in app template

cngonzalez:
### Description

This PR adds support for configuring data sources in Sanity apps. It introduces a new `sources` property in the CLI config that allows apps to specify project IDs and datasets they need to access. The runtime now automatically passes these sources to the app, eliminating the need for manual configuration in the App component.

### What to review

- The new `sources` schema in the CLI config
- The updated app entry module that passes sources to `renderSanityApp`
- The removal of manual SanityApp configuration in the basic-app example
- The updated CLI build process that extracts sources from the config

### Testing

The changes have been tested with the basic-app example, which now uses a real organization ID and project/dataset configuration. The example app has been updated to use the latest SDK packages from PR #704.


packages/@sanity/cli/src/actions/init/createAppCliConfig.ts
```diff
@@ -7,14 +7,23 @@ export default defineCliConfig({
   app: {
     organizationId: '%organizationId%',
     entry: '%entry%',
+    resources: {
```
cngonzalez: We now have[ a couple of implementations of this](https://github.com/sanity-io/cli/blob/1f1542639f433848064d9d4ce3dfa02381174ed0/packages/%40sanity/cli/src/actions/init/createAppCliConfig.ts), and I worry that this will cause a race condition / merge conflict when we release v3. I suggest we create a stack with _just_ the new CLI config type changes and config template as the base, and merge our PRs into it when the time comes).

cngonzalez (commented):
Mostly looks good! I had a couple of comments / architectural questions I'd like to hash out.

PR #612, opened 2026-03-11
feat(init): select project / dataset during app init

cngonzalez:


binoy14:
Closing in favor of #968

PR #760, opened 2026-03-23
feat(deploy): add --url flag and unattended mode support

rexxars:
## Summary

- Adds a `--url` flag to `sanity deploy` that works for both external and internal studio deploys. For external deploys, it passes the full URL. For internal deploys, it strips the `.sanity.studio` suffix if present, so users can pass either `my-studio` or `my-studio.sanity.studio`.
- The `--url` flag takes precedence over the `studioHost` config value, making it useful as a CLI override.
- When `--yes` (unattended mode) is used without `--url` or `studioHost` config, the command now errors with a helpful message (pointing to `--url`) instead of hanging on an interactive prompt.
- Covers both prompt sites: the `select()` prompt when multiple studios exist, and the `input()` prompt when no studio is configured.

FIXES SDK-1166

## Test plan

- [x] All 14 external deploy tests pass (7 new)
- [x] All 35 internal deploy tests pass (7 new)
- [x] Typecheck passes
- [x] Lint passes

### DIFF
--- a/packages/@sanity/cli/src/commands/__tests__/deploy.studio.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/deploy.studio.test.ts
@@ -1,4 +1,4 @@
-import {getCliTelemetry, studioWorkerTask} from '@sanity/cli-core'
+import {exitCodes, getCliTelemetry, studioWorkerTask} from '@sanity/cli-core'
 import {input, select} from '@sanity/cli-core/ux'
 import {mockApi, testCommand, testFixture} from '@sanity/cli-test'
 import nock from 'nock'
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/commands/__tests__/deploy.studio.test.ts
```diff
@@ -1208,6 +1208,479 @@ describe('#deploy studio', () => {
...
+        mocks: {
+          cliConfig: {
+            api: {projectId},
+          },
+        },
+      })
+
+      expect(error).toBeInstanceOf(Error)
+      expect(error?.message).toContain('Invalid studio hostname')
+      expect(error?.message).toContain('letters, numbers, and hyphens')
```
binoy14: should test for error code?

packages/@sanity/cli/src/commands/__tests__/deploy.studio.test.ts
```diff
@@ -1208,6 +1208,479 @@ describe('#deploy studio', () => {
...
+        mocks: {
+          cliConfig: {
+            api: {projectId},
+          },
+        },
+      })
+
+      expect(error).toBeInstanceOf(Error)
+      expect(error?.message).toContain('Cannot prompt for studio hostname in unattended mode')
+      expect(error?.message).toContain('Use --url to specify the studio hostname')
```
binoy14: should test for exit code?

PR #770, opened 2026-03-24
feat: workbench dev server

gu-stav:
### Description

> [!IMPORTANT]
> This will go into a feature branch for now and won't be released to `main` anytime soon. It's main purpose is to unblock development of `workbench` and to showcase how we imagine the setup to work.

Setup a workbench dev-server running alongside the studio/ app dev-server. It uses temporary runtime files written to `.sanity/workbench` as Vite `root`, similar to how runtime for studios works today.

Instead of navigating to the running application directly, users will be prompted to open `workbench` instead, which is why the `port` in the config now refers to the workbench `port`. This is something that we will likely have to discuss further at some point, because it mixes application and workbench configuration.

This needs the following studio PR to work: https://github.com/sanity-io/sanity/pull/12511, which exports `renderWorkbench` from `@sanity/workbench`.

### Testing

This can be tested using the studio and CLI pre-release together, but it will still fail, because we don't have a remote target yet.

It nevertheless shows that the code that is being run comes from `@sanity/workbench` which is what we expect at this point:

Screenshot 2026-03-26 at 14 48 23

### DIFF
--- a/packages/@sanity/cli/src/actions/dev/__tests__/startWorkbenchDevServer.test.ts
+++ b/packages/@sanity/cli/src/actions/dev/__tests__/startWorkbenchDevServer.test.ts
@@ -3,12 +3,15 @@ import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest'
 
 import {startWorkbenchDevServer} from '../startWorkbenchDevServer.js'
 
-const mockModuleResolve = vi.hoisted(() => vi.fn())
+const mockResolveLocalPackage = vi.hoisted(() => vi.fn())
 const mockCreateServer = vi.hoisted(() => vi.fn())
 const mockGetSharedServerConfig = vi.hoisted(() => vi.fn())
 const mockWriteWorkbenchRuntime = vi.hoisted(() => vi.fn())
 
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -1,7 +1,42 @@
...
+    ? await startAppDevServer(options)
+    : await startStudioDevServer(options)
+
+  // server is undefined only when startAppDevServer exits early (e.g. missing orgId);
+  // in that case the process is already exiting so no workbench needed.
+  if (!server) {
+    return {close: closeAppDevServer}
+  }
+
+  const workbenchPort = server.config.server.port + 1
```
joshuaellis: I think we should inverse this, we use the port provided (or not provided) and the app moves up one. so for instance, people still go to `localhost:3333` and they're greeted with their OS on their studio app (we'd need to construct the URL of course)

binoy14:
Whats the thinking around exposing workbench from `sanity` vs just using `@sanity/workbench` directly?

PR #887, opened 2026-04-06
feat: add support for user attributes

refiito:
**Add CLI support for User Attributes API**

**What**

Adds CLI commands for managing User Attributes — the foundation for parameterised role-based access control. User attributes are key-value pairs (e.g. location="UK", departments=["hr","sales"]) that can be referenced in role definitions and access filters.

> Note: The User Attributes API is currently in preview (vX). These commands are subject to change as the API stabilises.

**Commands**

**Attribute definitions** (org-level schema — requires sanity.organization.manage):

```
  sanity users attributes definitions list  --org-id <id> [--json]
  sanity users attributes definitions create --org-id <id> --key <key> --type <type> [--json]
  sanity users attributes definitions delete --org-id <id> <key>
```

**User attribute values** (list defaults to your own user; set/unset require sanity.organization.manage):

```
  sanity users attributes list  --org-id <id> [--user-id <id>] [--json]
  sanity users attributes set   --org-id <id> --user-id <id> --attributes '[{"key":"location","value":"UK"}]' [--json]
  sanity users attributes unset --org-id <id> --user-id <id> --key <key> [--key <key>...]

```
  When --org-id is omitted in an interactive terminal, the CLI prompts for an organization selection.

**Testing**

`pnpm test packages/@sanity/cli/src/commands/users/attributes`

---

> [!NOTE]
> **Medium Risk**
> New org-scoped commands can change user metadata and attribute schemas (access-control related), though behavior is gated by API permissions and the surface is preview (`vX`) with validation and tests.
> 
> **Overview**
> Adds **CLI support for the preview User Attributes API** (`vX`), enabling org admins to manage attribute schemas and per-user values used for parameterized access control.
> 
> New **`sanity users attributes`** commands cover **list / set / unset** on user values (with `--user-id`, JSON `--attributes` validation, and table or `--json` output) and **`definitions list / create / delete`** for org-level keys and types. A shared **`userAttributes` service** wraps the HTTP client; **`promptForOrganization`** plus a reusable **`--organization` / `-o`** flag (`getOrganizationFlag`) supply the org when omitted in interactive mode.
> 
> Ships broad **nock-based command tests**, a small **`formatAttributeValue`** helper, a **minor `@sanity/cli` changeset**, and a trivial **test import order** fix in `decorateIndexWithStagingScript.test.ts`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8547a24a72d6b08e5aa46805c3c0a31d79a739f9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/services/userAttributes.ts
```diff
@@ -0,0 +1,112 @@
...
+  return client.request({
+    method: 'DELETE',
+    uri: `/organizations/${orgId}/attribute-definitions/${encodeURIComponent(key)}`,
+  })
+}
+
+/**
+ * Get the authenticated user's own attributes within an organization
+ */
+export async function getMyAttributes(orgId: string): Promise<UserAttributesGetResponse> {
```
cngonzalez: "My" is a little idiosyncratic in this repo -- maybe `getCliUserAttributes` (to match the `getCliUser` function in `user.ts`

packages/@sanity/cli/src/actions/userAttributes/types.ts
```diff
@@ -0,0 +1,56 @@
+export type AttributeType =
```
cngonzalez: I'm a bit confused on why this and the constants file are in the actions directory when there's no corresponding actions. Should they be moved to `commands` where they're actually used? Or do we foresee a need for `actions` down the line (multi-step processes etc)

packages/@sanity/cli/src/actions/userAttributes/types.ts
```diff
@@ -0,0 +1,56 @@
...
+export interface AttributeDefinitionListResponse {
+  definitions: AttributeDefinition[]
+  hasMore: boolean
+
+  nextCursor?: string | null
+}
+
+export type AttributeValue = (number | string)[] | boolean | number | string
+
+export interface UserAttributeValues {
```
cngonzalez: It might be worth making a comment on this type -- I'm a little confused because it seems to be used as the value of an attribute as part of `UserAttribute`, which has an `activeSource` that indicates if it's `saml` or `sanity` but there's a possibility for an attribute to have a value that is not in line with with the `activeSource` on the attribute? Is it shared or updated perhaps?

packages/@sanity/cli/src/util/sharedFlags.ts
```diff
@@ -63,6 +63,35 @@ export function getProjectIdFlag(options: SharedFlagOptions) {
...
+ * Locked: flag name (`org-id`), char (`o`), `helpValue` (`<id>`), and parse (trims + validates non-empty).
+ */
+export function getOrgIdFlag(options: SharedFlagOptions) {
+  const {description: baseDescription, helpGroup, semantics, ...rest} = options
+  const isOverride = semantics === 'override'
+  const description =
+    (baseDescription ?? 'Organization ID to use') + (isOverride ? OVERRIDE_SUFFIX : '')
+
+  return {
+    'org-id': Flags.string({
```
cngonzalez: I'm not the final say on this, but I wanted to point out that elsewhere the CLI uses flags to specify an org, it's `--organization` (like in `init` and `projects create`). We may want to be consistent with the rest of the CLI

packages/@sanity/cli/src/commands/users/attributes/definitions/create.ts
```diff
@@ -0,0 +1,99 @@
...
+  'integer-array',
+  'number',
+  'number-array',
+  'boolean',
+]
+
+export class UserAttributeDefinitionsCreateCommand extends SanityCommand<
+  typeof UserAttributeDefinitionsCreateCommand
+> {
+  static override description = 'Create an attribute definition for an organization'
```
cngonzalez: Would "user attribute definition" be clearer here? I'm on the fence since the command is `sanity user attribute definition create`.

packages/@sanity/cli/src/commands/users/attributes/definitions/create.ts
```diff
@@ -0,0 +1,99 @@
...
+
+  static override examples = [
```
cngonzalez, line 26: Maybe not required for _every_ command, but it might be nice to provide examples that illustrate that the organization ID is not needed for these commands, since we're prompting for it.

(I think we also have a style guide rule for progressive complexity which you're doing a great job of here! Just think the "no org" example could be nice)

packages/@sanity/cli/src/commands/users/attributes/definitions/list.ts
```diff
@@ -0,0 +1,104 @@
...
+    let result: Awaited<ReturnType<typeof listAttributeDefinitions>>
+    try {
+      result = await listAttributeDefinitions(orgId)
+    } catch (err) {
+      debug('Error fetching attribute definitions', err)
+      this.error(`Failed to fetch attribute definitions:\n${getErrorMessage(err)}`, {exit: 1})
+    }
+
+    if (outputJson) {
+      this.log(JSON.stringify(result, null, 2))
```
cngonzalez: There is a `colorizeJson` helper available that might be nice for consistency

packages/@sanity/cli/src/commands/users/attributes/definitions/list.ts
```diff
@@ -0,0 +1,104 @@
...
+        sources: def.sources.join(', '),
+        type: def.type,
+      })
+    }
+
+    table.printTable()
+
+    if (result.hasMore) {
+      this.log(
+        '\nNote: Results are truncated. Use --json and the API directly with a cursor to fetch more.',
```
cngonzalez, line 105: Would it make sense for the user to run another CLI command to get a cursor or just hit the API directly (assuming it already gives a cursor?) It might also be nice to give them the URL to hit, if they need a token, etc.

packages/@sanity/cli/src/commands/users/attributes/definitions/delete.ts
```diff
@@ -0,0 +1,66 @@
...
+      description: 'Attribute key to delete',
+      required: true,
+    }),
+  }
+
+  static override description = 'Delete an attribute definition for an organization'
+
+  static override examples = [
+    {
+      command: '<%= config.bin %> <%= command.id %> --org-id o123 location',
```
cngonzalez: Generally, examples tend to position arguments before flags (though technically I think OCLIF allows either)

packages/@sanity/cli/src/commands/users/attributes/set.ts
```diff
@@ -0,0 +1,141 @@
...
+
+  static override examples = [
+    {
+      command:
+        '<%= config.bin %> <%= command.id %> --org-id o123 --user-id u456 --attributes \'[{"key":"location","value":"UK"}]\'',
+      description: 'Set a single attribute for a user',
+    },
+    {
+      command:
+        '<%= config.bin %> <%= command.id %> --org-id o123 --user-id u456 --attributes \'[{"key":"location","value":"UK"},{"key":"year_started","value":2020}]\'',
```
cngonzalez: I'm not sure if we have a standard or recommendations about passing JSON on the command line. @binoy14 might have more thoughts about this or has encountered this before.

cngonzalez (commented):
Functionally looks pretty good! Most comments are nits about style and guidelines

packages/@sanity/cli/src/services/userAttributes.ts
refiito: Renamed to `getCliUserAttributes`.

packages/@sanity/cli/src/actions/userAttributes/types.ts
refiito: Good point — there were no action files. I inlined the types and the `USER_ATTRIBUTES_API_VERSION` constant into `services/userAttributes.ts`, which matches the pattern in `services/organizations.ts` (types live in the service file that owns them). The `actions/userAttributes/` directory is gone.

packages/@sanity/cli/src/actions/userAttributes/types.ts
refiito: Added a JSDoc clarifying the distinction:

> Raw per-source values for a single attribute. Each entry holds the value received from that source (e.g. asserted in a SAML assertion, or set explicitly through Sanity). Both may be present at once; the value the API and access rules use is `UserAttribute.activeValue`, picked according to `UserAttribute.activeSource`.

To answer the question directly: yes, both `saml` and `sanity` can hold values simultaneously, which can differ. The resolved value is on the outer `UserAttribute` (`activeSource` + `activeValue`); `values` is the raw per-source state.

packages/@sanity/cli/src/util/sharedFlags.ts
refiito: Renamed to `--organization` everywhere: helper is now `getOrganizationFlag`, flag is `--organization` (kept the `-o` short form). All 5 callers + their tests updated. Verified `getOrgIdFlag` had no consumers outside this PR.

For now `--organization` accepts an org ID only (matching the API). `projects create` accepts slug-or-id; we can extend ours similarly in a follow-up if the API gains slug support.

packages/@sanity/cli/src/commands/users/attributes/definitions/create.ts
refiito: Updated all three definitions command descriptions (`create`, `list`, `delete`) to use "user attribute definition", and the corresponding success/info log lines too.

packages/@sanity/cli/src/commands/users/attributes/definitions/create.ts
refiito: Added no-org examples to all three definitions commands (create, list, delete) — they show that running without `--organization` will prompt for one in interactive mode.

packages/@sanity/cli/src/commands/users/attributes/definitions/list.ts
refiito: Switched all four `--json` output paths to `colorizeJson` (definitions `list`/`create` and attributes `list`/`set`). Color is only emitted in interactive TTY, so `--json | jq` still parses fine.

packages/@sanity/cli/src/commands/users/attributes/definitions/list.ts
refiito: Leaving the truncation note as-is for this PR — the User Attributes API is still in preview (vX) and the pagination contract may change. Once it's stable I'll come back and either thread the cursor through the CLI or document the API URL + token guidance in the help text, in a follow-up. Filed as a TODO for the User Attributes Followups project.

packages/@sanity/cli/src/commands/users/attributes/definitions/delete.ts
refiito: Reordered: example is now `sanity users attributes definitions delete location --organization o123` (positional arg first). Also added a no-org variant.

packages/@sanity/cli/src/commands/users/attributes/set.ts
refiito: Leaving the JSON-on-CLI shape as-is in this PR — happy to revise if @binoy14 has a stronger convention. Current shape is a JSON-array string passed to `--attributes`, which matches what we already do for other multi-value structured flags. Open to either: (1) a documented JSON-array stays, (2) per-attribute flag pairs like `--key location --value UK` (repeatable), (3) a `--attributes-file` for larger payloads. Tracking as a separate decision rather than blocking this PR.

refiito:
Round 3 of review feedback addressed in b9c5b4c5. Summary of disposition on the Claude bot review:

**1. `set.ts` key string check** — addressed. Validation now requires `typeof item.key === 'string'`, and the error message names the constraint.
**2. `userAttributes.ts` DELETE with body** — left as-is. The User Attributes API is in preview (vX), the Sanity API gateway is ours end-to-end, and the cli-core HTTP client preserves the body through to the request. If a real-world intermediary issue surfaces I'll revisit; the failure mode the comment describes is theoretical for our deployment topology.
**3. `unset.ts` success message uses flag value** — addressed. Now uses `result.sanityUserId` from the API response, mirroring `set.ts`.
**4. `promptForOrganization.ts` coverage** — addressed. New test file at `prompts/__tests__/promptForOrganization.test.ts` covers the NonInteractive throw, the empty-orgs case, the fetch-failure path with spinner state, the success path with formatted choices, and the spinner-succeed lifecycle.

Plus all of @cngonzalez's inline nits — see individual thread replies.

runeb:
@claude please re-review the PR

PR #890, opened 2026-04-06
fix(cli): rewrite update checker with background worker and project-aware caching

rexxars:
## Problem

The CLI update checker has several issues:

- **Wrong package in update command**: The checker compares `@sanity/cli` versions from npm but tells users to run `pnpm update sanity`. Since `@sanity/cli` and `sanity` are separate packages with independent version numbers, this is confusing and sometimes incorrect.
- **Premature update prompts**: A new `@sanity/cli` version can be published before `sanity` depends on it, causing the checker to suggest an update that isn't actually available through `sanity` yet.
- **Blocks CLI execution**: The version check makes a network request inline during CLI startup, adding latency to every command. If the npm registry is slow or unreachable, this directly delays the user.
- **Unncessary checks for temporary npx downloads**: Running `npx @sanity/cli` (without a local install) triggers update checks for a binary that will be discarded immediately.

## Solution

- **Project-aware package resolution**: Detects whether the local project depends on `sanity` or `@sanity/cli`, checks the correct package on npm, and shows the matching update command (e.g. `pnpm update sanity` vs `npm update @sanity/cli`).
- **Non-blocking background worker**: Version fetches run in a detached child process (same pattern as telemetry flush) that writes to a config cache. The main CLI process only reads from cache - it never waits on network. On the first run the cache is empty so no notification is shown; on subsequent runs the cached result is available instantly.
- **Project-isolated caching**: The cache stores only the latest npm version per package (globally valid). The installed version is resolved locally on each invocation, so switching between projects with different versions never shows stale or incorrect information.
- **npx temporary download detection**: Skips update checks when the CLI is run from a temporary npx cache (`/_npx/` in the binary path), while still checking when `npx sanity` resolves to a locally installed binary.

## Test plan

- [x] All 2464 tests pass (2 pre-existing snapshot failures in unrelated formatDocumentValidation.test.ts)
- [x] TypeScript type checking passes
- [x] Lint passes (no new warnings)
- [x] Build passes
- [ ] Manual test: run CLI in a project with `sanity` dep - should show `pnpm update sanity`
- [ ] Manual test: run CLI outside a sanity project - should show `npm update @sanity/cli`
- [ ] Manual test: switch between two projects with different versions - should show correct installed version for each


binoy14:
This was fixed by other PRs

PR #891, opened 2026-04-07
fix: make `@sanity/telemetry` a regular dep

stipsan:
### Description

Having `@sanity/telemetry` specified as a peer dependency of `@sanity/cli-core` is causing some problems. When `@sanity/telemetry` v0.9 dropped the `sanity` package itself updated faster than `@sanity/cli-core` updated its peer dependency range to allow it.
It can lead to odd branching in pnpm, with nasty workarounds: https://github.com/sanity-io/next-sanity/blob/370bbfb24ba82f50029f94d177c36a60c6a81969/.pnpmfile.cjs#L6-L13
`@sanity/cli` itself, like `sanity`, lists `@sanity/telemetry` as a direct dependency, not a peer.
It's far better to have it as a direct dependency, as deduping will happen when possible. And if deduping is not possible, instead of peer dependency warnings and unpredictable duplicated node_modules trees, it'll copy multiple versions of `@sanity/telemetry` in predictable and manageable ways.

### What to review

I ran an analysis with claude on the repo to see if there's any singleton patterns or equivalent that would explain why it's a peer, but none could be found so it looks like this change should be fine?

### Testing

CI should be sufficient.

### DIFF
--- a/packages/@sanity/cli-core/package.json
+++ b/packages/@sanity/cli-core/package.json
@@ -67,6 +67,7 @@
     "@oclif/core": "catalog:",
     "@rexxars/jiti": "^2.6.2",
     "@sanity/client": "catalog:",
+    "@sanity/telemetry": "catalog:",
     "babel-plugin-react-compiler": "^1.0.0",
     "boxen": "^8.0.1",
     "debug": "catalog:",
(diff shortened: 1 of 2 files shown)


binoy14:
@stipsan closing this since we removed the peer dependency in #902

PR #907, opened 2026-04-09
feat: add workbench under unstable flags

gu-stav:
### Description

Lands Workbench for the Sanity CLI: an opt-in way to develop and build federated applications (studios and SDK apps) behind a single dev server.

**The shape.** `sanity dev` for an opted-in app starts an extra dev server — the workbench shell — and loads the real app into it over Module Federation. Same delivery model as Dashboard: one host, many apps mounted as remotes. The app's own dev server registers itself against that shell as a federated remote; the shell renders it.

**One server, many apps.** Only one workbench runs per machine. The first opted-in `sanity dev` to start acquires a lockfile and hosts the shell; every other `sanity dev` finds it through the lock and registers its remote there. Run three apps locally and you get one workbench hosting three remotes.

**Opt-in is a single identity function.** Apps opt in by wrapping their config in `unstable_defineApp(...)` in `sanity.cli.ts`. That's the only signal — it stamps a brand symbol the CLI detects (`isWorkbenchApp`). Without it, dev/build/deploy take the exact paths they take on `main`. No flag, no env var, no behavioural change for anyone who doesn't call it.

**Build + deploy.** A branded app builds a Module Federation remote (`mf-manifest.json`, exposed `./App` / `./views/*` / `./services/*`) through a `sanity/federation` Vite plugin stack, instead of a plain SPA. `sanity deploy` validates the federation manifest instead of `index.html`. Studios and SDK apps share the path.

**Scaffolding.** `sanity init --unstable--workbench` scaffolds a federated app (wraps the config in `unstable_defineApp`). The flag is hidden and undocumented — internal-only while the API carries the `unstable_` prefix.

Experimental; should not surface in release notes.

### What to review

Three areas, in order of how much I'd want eyes on them.

**1. Nothing changes without opting in.** Every workbench branch is gated on `isWorkbenchApp()` (the brand) or the internal `SANITY_INTERNAL_IS_WORKBENCH_REMOTE` env var. Config parsing only forks after the brand check; `strictPort`, entry-module, Vite plugins and auto-updates all keep `main`'s logic on the non-workbench side; signal handlers attach only when there's a lock or registration to clean up; `init --unstable--workbench` defaults to off. The two new `app` fields (`group`, `priority`) are additive optionals. `getCliConfig.ts` and `getViteConfig.ts` sit on every command's hot path — worth a second read there.

**2. Lockfile + process handling.** The single-workbench model rests on `workbenchLock.ts` plus the registry. `O_EXCL` (`wx`) makes acquisition atomic, stale locks are pruned via liveness + OS start-time (PID reuse handled cross-platform, including Windows PowerShell), and the lock releases on normal exit / SIGINT / SIGTERM / throw-during-startup. Sharp edges to flag rather than block on:

- `updatePort()` writes the real Vite port back after the server listens, so there's a tiny window where a registrant reads the placeholder port — and that write isn't wrapped, so a failed write leaves a stale port silently.
- A short gap between the workbench server returning and the signal handlers attaching where a SIGINT can leak the lock (self-heals on next run via pruning).

### Testing

Coverage spans the new surface — Vite config, build artifacts, dev orchestration, the lock + registry (PID-reuse and Windows paths included), and deploy validation — plus a `federated-studio` fixture. Non-workbench paths stay on the existing suites, unchanged and still green.

---

> [!NOTE]
> **High Risk**
> Large cross-cutting change on every-command hot paths (`getCliConfig`, Vite config, `sanity dev` orchestration) plus new process/lock/registry semantics; regressions could affect non-opted-in projects if gating fails.
> 
> **Overview**
> Adds **Workbench** as an opt-in path for developing and building Sanity studios and SDK apps as Module Federation remotes behind a shared local shell.
> 
> **Opt-in.** Projects call `unstable_defineApp(...)` in `sanity.cli.ts`; cli-core detects the brand via `isWorkbenchApp` and parses workbench-specific config (views, services, `applicationType`) without touching legacy `app` schema. `sanity init --unstable--workbench` scaffolds the branded config; default init stays unchanged.
> 
> **New `@sanity/workbench-cli` package** (wired into cli, cli-build, CI previews) supplies federation Vite plugins, deploy helpers, and a dev-server registry.
> 
> **Dev (`sanity dev`).** `devAction` orchestrates a workbench host on the configured port (with lock/registry cleanup and signal handling), bumps the app/studio server to the next port when the host is up, registers remotes, watches manifests/interfaces, and **restarts** the app server when views/services change. A `SANITY_INTERNAL_IS_WORKBENCH_REMOTE` escape hatch runs the shell app standalone.
> 
> **Build.** Workbench builds skip the normal SPA runtime path and use `workbenchVitePlugins`; `resolveEntries` / entry modules handle dock-only apps (no `entry`) vs legacy SDK defaults. Vite `strictPort` relaxes for apps and workbench.
> 
> **Deploy.** Federated apps validate federation output and view payloads (stub persistence); studios reject `--external` for federated projects.
> 
> Also adds shared manifest schemas in cli-core, `getCliConfigUncached` for hot reload, `federated-studio` fixture, and broad unit/e2e coverage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb3bf21d95fd4cc924aaaadba3986990919ec82d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gu-stav:
The axios security warning will be resolved by https://github.com/module-federation/core/pull/4644 I think

gu-stav:
Rebased onto `main`

fixtures/federated-studio/sanity.config.ts
```diff
@@ -0,0 +1,17 @@
+import {defineConfig} from 'sanity'
+import {structureTool} from 'sanity/structure'
+
+import {schemaTypes} from './schemaTypes'
+
+export default defineConfig({
+  title: 'Basic Studio',
```
snocorp:
```suggestion
  title: 'Federated Studio',
```

packages/@sanity/cli-build/src/actions/build/writeSanityRuntime.ts
```diff
@@ -71,6 +74,44 @@ export async function writeSanityRuntime(options: RuntimeOptions): Promise<FSWat
...
+    runtimeDir,
+  })
+  const appJsContent = getEntryModule({
+    basePath,
+    entry: relativeEntry ?? undefined,
+    isApp,
+    reactStrictMode,
+    relativeConfigLocation,
+  })
+  await fs.writeFile(path.join(runtimeDir, 'app.js'), appJsContent)
```
snocorp, line 92: I know this isn't new in this PR but does this cause the watcher to trigger unnecessarily?

packages/@sanity/cli-core/src/config/cli/__tests__/workbenchApp.test.ts
```diff
@@ -0,0 +1,64 @@
+import {tmpdir} from 'node:os'
+import {join} from 'node:path'
+
+import {describe, expect, test} from 'vitest'
+
+import {parseWorkbenchCliConfig} from '../workbenchApp'
```
snocorp:
```suggestion
import {parseWorkbenchCliConfig} from '../workbenchApp.js'
```

packages/@sanity/cli-e2e/globalSetup.ts
```diff
@@ -31,22 +31,30 @@ export async function setup(project: TestProject): Promise<void> {
     console.log('Packing @sanity/cli-build...')
     const cliBuildTarball = packPackage('@sanity/cli-build')
 
+    // `@sanity/cli` and `@sanity/cli-build` depend on `@sanity/workbench-cli`,
+    // which isn't published to npm — pack it too so the installed binary can
+    // resolve it from the co-installed tarball instead of hitting the registry.
+    console.log('Packing @sanity/workbench-cli...')
+    const workbenchCliTarball = packPackage('@sanity/workbench-cli')
```
snocorp, line 39: Will this still be done after the new package is published to npm?

packages/@sanity/workbench-cli/src/actions/deploy/getWorkbench.ts
```diff
@@ -0,0 +1,74 @@
...
+            'Declare an entry or at least one view or service in the app config.',
+        )
+      }
+    },
+
+    async checkBuiltOutput(sourceDir) {
+      try {
+        const stats = await stat(sourceDir)
+        if (!stats.isDirectory()) {
+          throw new Error(`Directory ${sourceDir} is not a directory`)
```
snocorp: Kind of an awkward error message

packages/@sanity/cli/src/actions/dev/__tests__/devAction.unit.test.ts
snocorp, line 1: Looks like this test avoids fixtures (which is awesome!). I'd suggest renaming it to `devAction.unit.test.ts` to run with the rest of the tests that are like this.

packages/@sanity/cli/src/commands/__tests__/dev.test.ts
```diff
@@ -267,6 +272,50 @@ describe('#dev', {timeout: (platform() === 'win32' ? 60 : 30) * 1000}, () => {
     })
   })
 
+  describe('workbench-app', () => {
+    test('warns that --load-in-dashboard is ignored for workbench apps', async () => {
```
snocorp: This could be a unit test instead for better performance and consistency.

packages/@sanity/cli/src/actions/dev/registration/interfaceSetId.ts
```diff
@@ -0,0 +1,42 @@
+import {type DevServerInterface} from './deriveInterfaces.js'
+
+/**
+ * The identity of an app's declared interface set — an order-independent key
+ * over its forwarded Interface records (interface_type, name, entry_point).
+ * Two sets that differ only in declaration order share an id, so reordering
+ * `views`/`services` in `sanity.cli.ts` is not a change; adding, removing,
+ * renaming, or repointing a view/service is. `undefined` (project types that
+ * declare no interfaces, e.g. studios) ids to the empty set.
```
snocorp: This comment is a bit confusing. Looks like some edits got mangled somehow.

packages/@sanity/cli/src/actions/dev/registration/startDevManifestWatcher.ts
```diff
@@ -0,0 +1,160 @@
...
+  const watcher = watch(configDir, onEvent)
+
+  watcher.on('error', (err) => {
+    devDebug('Config watcher error: %O', err)
+    output.warn(`Manifest watcher error: ${err instanceof Error ? err.message : String(err)}`)
+  })
+
+  return {
+    close: async () => {
+      closed = true
```
snocorp, line 159: Should we check if close has already been called as well?

packages/@sanity/cli/src/actions/dev/registration/__tests__/startDevManifestWatcher.test.ts
```diff
@@ -0,0 +1,321 @@
...
+    fakeWatcher.emitChange('sanity.config.ts')
+    fakeWatcher.emitChange('sanity.config.ts')
+    fakeWatcher.emitChange('sanity.config.ts')
+
+    await vi.advanceTimersByTimeAsync(300)
+
+    expect(mockExtract).toHaveBeenCalledTimes(2)
+    expect(update).toHaveBeenCalledTimes(2)
+
+    await watcher.close()
```
snocorp: Should this be in a finally or afterEach?

packages/@sanity/cli/src/actions/dev/registration/__tests__/startDevManifestWatcher.test.ts
```diff
@@ -0,0 +1,321 @@
...
+    await vi.advanceTimersByTimeAsync(0)
+    expect(mockExtract).toHaveBeenCalledTimes(1)
+
+    await watcher.close()
+    expect(fakeWatcher.closed).toBe(true)
+
+    fakeWatcher.emitChange('sanity.config.ts')
+    await vi.advanceTimersByTimeAsync(300)
+
+    expect(mockExtract).toHaveBeenCalledTimes(1)
```
snocorp, line 271: Does this need to close the watcher as it does in other tests?

packages/@sanity/cli/src/actions/dev/servers/startAppDevServer.ts
```diff
@@ -0,0 +1,73 @@
...
+    const appTitle = cliConfig && 'app' in cliConfig ? cliConfig.app?.title : undefined
+    const {close, server} = await startDevServer({
+      ...config,
+      appTitle,
+      isApp: true,
+    })
+
+    const {port} = server.config.server
+
+    if (isWorkbenchApp(cliConfig?.app)) {
```
snocorp: This is called twice in this code path, potentially could store the result instead,

packages/@sanity/workbench-cli/src/actions/build/vite/plugin.ts
```diff
@@ -0,0 +1,145 @@
...
+  }
+
+  const generatedEntry = `./${RUNTIME_DIR}/${FEDERATION_FILE_NAME}.jsx`
+
+  function resolveEntryPath(entry: string) {
+    const resolvedPath = path.resolve(workDir, entry)
+
+    if (!resolvedPath) {
+      throw new Error(
+        `Could not resolve path for entry "${entry}". Resolved to "${resolvedPath}". Please check that the file exists and the path is correct.`,
```
snocorp: I don't think the second part of this is useful since the value is falsey anyway:

```suggestion
        `Could not resolve path for entry "${entry}". Please check that the file exists and the path is correct.`,
```

packages/@sanity/workbench-cli/src/actions/dev/__tests__/workbenchLock.test.ts
```diff
@@ -0,0 +1,196 @@
...
+  testDataDir = join(tmpdir(), `sanity-registry-test-${process.pid}-${Date.now()}`)
+  mkdirSync(testDataDir, {recursive: true})
+  // By default, return a start time matching "now" so our own PID passes isOurProcess
+  mockExecSync.mockReturnValue(new Date().toString())
+  __resetStartTimeCacheForTesting()
+})
+
+afterEach(() => {
+  vi.clearAllMocks()
+  vi.unstubAllEnvs()
```
snocorp: I don't think we need this line, but maybe I'm missing something

packages/@sanity/workbench-cli/src/actions/dev/__tests__/workbenchLock.test.ts
```diff
@@ -0,0 +1,196 @@
...
+})
+
+describe('acquireWorkbenchLock', () => {
+  test('returns lock object when lock is available', () => {
+    const lock = acquireWorkbenchLock({host: 'localhost', port: 3333})
+    expect(lock).toBeDefined()
+    expect(lock!.release).toBeTypeOf('function')
+    expect(lock!.updatePort).toBeTypeOf('function')
+    lock!.release()
```
snocorp: Should this be in a finally or afterEach?

packages/@sanity/cli/src/actions/dev/workbench/__tests__/writeWorkbenchRuntime.test.ts
```diff
@@ -0,0 +1,205 @@
...
+  beforeEach(() => {
+    tmpDir = mkdtempSync(join(tmpdir(), 'sanity-workbench-'))
+  })
+
+  afterEach(() => {
+    rmSync(tmpDir, {recursive: true})
+  })
+
+  test('returns the absolute path to the workbench directory', async () => {
+    const result = await writeWorkbenchRuntime({cwd: tmpDir, reactStrictMode: false})
```
snocorp: Not sure how long this test takes to run but you might be ale to run `writeWorkbenchRuntime({cwd: tmpDir, reactStrictMode: false})` once instead of once per test and still do all the checks

packages/@sanity/cli/src/actions/dev/workbench/__tests__/writeWorkbenchRuntime.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    })
+
+    test('includes a module script tag loading workbench.js', async () => {
+      await writeWorkbenchRuntime({cwd: tmpDir, reactStrictMode: false})
+
+      const content = await fs.readFile(join(tmpDir, '.sanity', 'workbench', 'index.html'), 'utf8')
+      expect(content).toContain('<script type="module" src="./workbench.js">')
+    })
+
+    test('is valid HTML with charset meta tag', async () => {
```
snocorp:
```suggestion
    test('starts with a DOCTYPE and has a charset meta tag', async () => {
```

packages/@sanity/workbench-cli/src/actions/dev/__tests__/registry.test.ts
```diff
@@ -0,0 +1,230 @@
...
+  testDataDir = join(tmpdir(), `sanity-registry-test-${process.pid}-${Date.now()}`)
+  mkdirSync(testDataDir, {recursive: true})
+  // By default, return a start time matching "now" so our own PID passes isOurProcess
+  mockExecSync.mockReturnValue(new Date().toString())
+  __resetStartTimeCacheForTesting()
+})
+
+afterEach(() => {
+  vi.clearAllMocks()
+  vi.unstubAllEnvs()
```
snocorp: I don't think we need this line

packages/@sanity/workbench-cli/src/actions/dev/__tests__/registry.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    const manifest = JSON.parse(readFileSync(filePath, 'utf8'))
+    expect(manifest.pid).toBe(process.pid)
+    expect(manifest.type).toBe('studio')
+    expect(manifest.port).toBe(3334)
+    expect(manifest.host).toBe('localhost')
+    expect(manifest.workDir).toBe('/tmp/project')
+    expect(manifest.startedAt).toBeDefined()
+    expect(manifest.version).toBe(1)
+
+    cleanup()
```
snocorp, line 72: Should this be in a finally clause?

packages/@sanity/workbench-cli/src/actions/dev/__tests__/registry.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    })
+
+    expect(existsSync(filePath)).toBe(false)
+  })
+})
+
+describe('watchRegistry', () => {
+  test('invokes callback when a manifest file is added', async () => {
+    const callback = vi.fn()
+    const watcher = watchRegistry(callback)
```
snocorp, line 185: Today I learned about using onTestFinished so you could add `t` to the params of this test callback and then use `t.onTestFinished(() => watcher.close())` instead of closing the watcher at the end of the test logic to ensure it always gets called.

packages/@sanity/workbench-cli/src/actions/dev/__tests__/processLiveness.test.ts
```diff
@@ -0,0 +1,376 @@
...
+  testDataDir = join(tmpdir(), `sanity-registry-test-${process.pid}-${Date.now()}`)
+  mkdirSync(testDataDir, {recursive: true})
+  // By default, return a start time matching "now" so our own PID passes isOurProcess
+  mockExecSync.mockReturnValue(new Date().toString())
+  __resetStartTimeCacheForTesting()
+})
+
+afterEach(() => {
+  vi.clearAllMocks()
+  vi.unstubAllEnvs()
```
snocorp: I don't think we need this line

snocorp (dismissed):
Some questions and minor suggestions but nothing blocking.

packages/@sanity/workbench-cli/src/actions/build/vite/plugins/plugin-module-federation.node.test.ts
```diff
@@ -0,0 +1,75 @@
+import {type Environment, type Plugin} from 'vite'
```
cngonzalez, line 1: Might be good to throw these in a _\_tests__ dir rather than colocate

packages/@sanity/workbench-cli/package.json
```diff
@@ -0,0 +1,82 @@
...
+    "lint": "eslint .",
+    "publint": "publint",
+    "test": "vitest run",
+    "posttest": "pnpm run lint",
+    "test:coverage": "vitest run --coverage",
+    "test:watch": "vitest",
+    "watch": "swc --delete-dir-on-start --strip-leading-paths --out-dir dist/ --watch src"
+  },
+  "dependencies": {
+    "@module-federation/vite": "1.16.0",
```
cngonzalez, line 60: Not vital, but guessing this is pinned to avoid skew. Maybe worth putting in the catalog with a note? NBD

packages/@sanity/workbench-cli/src/contract.ts
```diff
@@ -0,0 +1,104 @@
+import {z} from 'zod/mini'
+
+// The shared kernel of the module-federation extension contract: the supported
+// view/service types, the contract versions the helpers stamp, and the
+// declaration schemas an app author writes in `unstable_defineApp({views, services})`.
+//
+// Lives in `@sanity/cli-build` (alongside the vite plugin that consumes the same
```
cngonzalez: We're no longer in \@sanity/cli-build, right?

packages/@sanity/workbench-cli/src/actions/dev/registry.ts
```diff
@@ -0,0 +1,357 @@
...
+    return {
+      release() {
+        try {
+          unlinkSync(lockPath)
+        } catch {
+          // Already cleaned up
+        }
+      },
+      updatePort(port: number) {
+        writeFileSync(lockPath, JSON.stringify({...lockData, port}))
```
cngonzalez, line 335: Is it worth validating that we still have the lock here? How much can things drift in the meantime? (This would also be true of release)

cngonzalez (commented):
Mostly stylistic nits, one question about the locking process

packages/@sanity/cli-e2e/globalSetup.ts
gu-stav: Good catch. The comment was wrong - the pre-packing needs to stay though even after publishing I think.

packages/@sanity/cli-build/src/actions/build/writeSanityRuntime.ts
gu-stav: Nice find. I've created an issue about this and will address it as soon as this is merged https://linear.app/sanity/issue/SDK-1769/sanity-dev-runtime-watcher-does-a-redundant-initial-render-chokidar

gu-stav:
@claude re-review this PR

packages/@sanity/workbench-cli/package.json
gu-stav: Since its only referenced once I'd prefer to keep it out of the catalog.

gu-stav:
@claude re-review this PR

gu-stav:
@claude re-review. Focus on important problems/ blockers.

snocorp (dismissed):

joshuaellis (dismissed):

joshuaellis (dismissed):

mttdnt (approved):

PR #915, opened 2026-04-09
fix(telemetry): buffering events before consent is resolved

mttdnt:
### Description

- Early telemetry events (userProperties, trace.start) emitted before async consent resolution settled were silently dropped. Added initComplete flag and eventBuffer — events emitted during init are buffered, then flushed to disk if consent resolves as granted, or discarded otherwise.

### Testing

- The lifecycle test was working around a known limitation by creating a separate createTelemetryStore instance to verify file I/O. Two things masked the real issue: the buffering bug above, and a missing tag query param on the nock mock for the consent endpoint. Both are now fixed, allowing setupTelemetry itself to write files. The workaround store is removed.
- Added a failure path test — when the consent API returns a 500, buffered events are discarded and no files are written, but the telemetry infrastructure (getCliTelemetry()) is still set up correctly.
- Updated filename regex from [a-zA-Z0-9-]+ to [\w-]+ to match real TypeID session IDs (session_<ulid>), which contain an underscore.

### DIFF
--- a/.changeset/fix-telemetry-event-buffer.md
+++ b/.changeset/fix-telemetry-event-buffer.md
@@ -1,0 +1,5 @@
+---
+'@sanity/cli': patch
+---
+
+Added buffer for telemetry events captured before consent was done resolving
--- a/.changeset/fix-telemetry-event-buffer.md
+++ b/.changeset/fix-telemetry-event-buffer.md
@@ -2,4 +2,4 @@
 '@sanity/cli': patch
 ---
 
-Added buffer for telemetry events captured before consent was done resolving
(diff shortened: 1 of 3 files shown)


.changeset/fix-telemetry-event-buffer.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sanity/cli': patch
+---
+
+Fixed telemetry events being silently dropped before consent resolves
```
binoy14, line 5: 🤔 this does not need a user facing changelog does it?

binoy14 (approved):

PR #918, opened 2026-04-10
feat(federation): pass additional props to federation plugin

joshuaellis:
### Description

* Updates the federation plugin and provides additional props required to write runtime files

requires https://github.com/sanity-io/workbench/pull/64 to be merged and published first.

### DIFF
--- a/packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
+++ b/packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
@@ -27,6 +27,7 @@ export async function startWorkbenchDevServer(
   })
 
   if (!cliConfig?.federation?.enabled) {
+    devDebug('Federation not enabled, skipping workbench dev server')
     return {httpHost, workbenchAvailable: false, workbenchPort}
   }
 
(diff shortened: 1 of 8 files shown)


packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
```diff
@@ -27,6 +27,7 @@ export async function startWorkbenchDevServer(
   })
 
   if (!cliConfig?.federation?.enabled) {
+    devDebug('Federation not enabled, skipping workbench dev server')
```
gu-stav, line 30: Nit: I think this shouldn't be logged, because it will log for every application out there, once the feature branch is released.

gu-stav (approved):

packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
joshuaellis: it wont be logged unless someone enables `DEBUG=sanity:*` – feels important to understand the code branches? But we can definitely review what we need before final merge – agree there.

PR #923, opened 2026-04-10
fix(cli): default to "production" dataset in unattended mode

binoy14:
## Summary

- **Removes** the requirement for `--dataset` in unattended mode (`--yes`), which previously caused `sanity init --yes` to fail with _"`--dataset` must be specified in unattended mode"_
- **Defaults** to the `"production"` dataset with public visibility when `--yes` is used without `--dataset`, matching the documented behavior of `--dataset-default`
- **Respects** an explicit `--visibility` flag when provided (e.g., `--visibility=private` is not silently overridden)
- **Updates** tests to reflect the new default behavior, including coverage for when the "production" dataset already exists

Fixes SDK-1235

## Context

The `--yes` flag description says _"Unattended mode, answers 'yes' to any 'yes/no' prompt and **otherwise uses defaults**"_. However, running `sanity init --yes` (or `--mcp --yes`) would error out requiring `--dataset` to be explicitly specified. The expected behavior is that it should default to a public `"production"` dataset — the same as `--dataset-default`.

### DIFF
--- a/packages/@sanity/cli/src/commands/__tests__/init/init.setup.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/init/init.setup.test.ts
@@ -188,7 +188,6 @@ describe('#init: oclif command setup', () => {
 
     // Should not throw a --dataset validation error
     expect(error?.message ?? '').not.toContain('--dataset')
-    expect(error?.oclif?.exit).toBeUndefined()
   })
 
   test('throws error when `output-path` is not used in unattended mode with non-nextjs project', async () => {
--- a/packages/@sanity/cli/src/commands/__tests__/init/init.setup.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/init/init.setup.test.ts
@@ -175,30 +175,30 @@ describe('#init: oclif command setup', () => {
     expect(error?.oclif?.exit).toBe(1)
   })
 
-  test('throws error when in unattended mode and `dataset` is not set', async () => {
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/commands/__tests__/init/init.setup.test.ts
```diff
@@ -175,24 +175,13 @@ describe('#init: oclif command setup', () => {
     expect(error?.oclif?.exit).toBe(1)
   })
 
-  test('throws error when in unattended mode and `dataset` is not set', async () => {
```
binoy14, line 178: Instead of removing this test it should just change to asserting for if no dataset flag is provided it uses the default production

binoy14: Fixed. Changed the test to use `--bare` (bypasses the `--output-path` requirement) without `--dataset`, then asserts no `--dataset` validation error is thrown. This makes it distinct from the output-path test below and verifies dataset is no longer required at the validation layer. The "defaults to production" behavior is covered in `init.get-project-details.test.ts`.

mttdnt (approved):
LGTM! Just a thought, but not necessarily feedback. Would it be helpful for human/agents to print off the dataset name and visibility in the success message?

binoy14:
> Would it be helpful for human/agents to print off the dataset name and visibility in the success message?

possibly, I am not sure if we do that now or not but worth checking it out

PR #926, opened 2026-04-10
refactor(cli-core): replace tsx with jiti in getCliConfigSync

binoy14:
### Description

Replaces `tsx` with `@rexxars/jiti` for synchronous TypeScript config loading in `getCliConfigSync`:

- **`getCliConfigSync.ts`** — replaced the tsx `register()`/`unregister()` + `createRequire()` pattern with a simpler `createJiti()` sync call for loading `sanity.cli.ts` config files
- **`importModule.ts`** — removed stale "wrapper around tsx" doc comment (already uses jiti)
- **`pnpm-workspace.yaml`** — added `@rexxars/jiti` to catalog, updated `cli-core` to use `catalog:` reference

`execScript.ts` was not changed because jiti's register hook doesn't handle CJS/ESM interop with top-level await (`ERR_REQUIRE_ASYNC_MODULE` via `@oclif/core`), so that must stay on tsx.

### What to review

- `packages/@sanity/cli-core/src/config/cli/getCliConfigSync.ts` — sync jiti call replacing tsx register/unregister pattern
- `packages/@sanity/cli-core/package.json` — `@rexxars/jiti` now uses `catalog:` reference
- `pnpm-workspace.yaml` — `@rexxars/jiti` added to catalog

### Testing

Existing tests pass. The `getCliConfigSync` unit tests continue to pass as they mock `existsSync` and don't exercise the actual loader. The `exec` integration tests also pass since `execScript.ts` was not changed.

### DIFF
--- a/packages/@sanity/cli-core/package.json
+++ b/packages/@sanity/cli-core/package.json
@@ -65,7 +65,7 @@
   "dependencies": {
     "@inquirer/prompts": "^8.3.0",
     "@oclif/core": "catalog:",
-    "@rexxars/jiti": "^2.6.2",
+    "@rexxars/jiti": "catalog:",
     "@sanity/client": "catalog:",
     "babel-plugin-react-compiler": "^1.0.0",
     "boxen": "^8.0.1",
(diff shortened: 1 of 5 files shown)


binoy14:
Closing for now — jiti's synchronous mode (`jiti(path)`) uses `vm.runInThisContext` (CJS sandbox) which can't handle `import.meta` in transitive dependencies like vite. This causes the `sanity exec` integration tests to fail. Will revisit with a different approach.

PR #930, opened 2026-04-13
fix(dev): disable strict ports for applications

gu-stav:
### Description

While https://github.com/sanity-io/cli/pull/830 disabled strict ports for the _workbench_ dev-server, running `dev` often times still fails, because applications had `strictPort` set to true and therefore fail, if another application is already running on that port.


packages/@sanity/cli/src/actions/build/__tests__/decorateIndexWithStagingScript.test.ts
```diff
@@ -1,5 +1,7 @@
 import {afterEach, describe, expect, test, vi} from 'vitest'
 
+import {decorateIndexWithStagingScript} from '../decorateIndexWithStagingScript'
```
gu-stav, line 3: Running `pnpm format` introduced this 🤷🏼

joshuaellis (approved):
ah yeah i just ran into this.

PR #932, opened 2026-04-13
feat(workbench): support only running one workbench instance and finding local apps

joshuaellis:
### Description

* creates dev-server manifests at the root of the user's machine to coordinate which processes are running for which apps
* creates a lock file so there can only be one workbench instance at a time
* uses the HotClient API to forward the list of applications to the user's client app only on the workbench dev-server

### What to review

* I've opted for this over something like bonjour since it can largely be done natively, and we dont need to know whats going on in the network per say we just want to know whats happening on the local machine – does that make more sense? I think so?

### DIFF
--- a/packages/@sanity/cli/src/actions/dev/devAction.ts
+++ b/packages/@sanity/cli/src/actions/dev/devAction.ts
@@ -57,21 +57,26 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
 
   // Register the studio/app dev server in the registry (federated projects only)
   let cleanupManifest: () => void = syncNoop
+  let onSignal: (() => void) | undefined
   if (options.cliConfig?.federation?.enabled) {
     const addr = server.httpServer?.address()
     const appPort = typeof addr === 'object' && addr ? addr.port : server.config.server.port
     cleanupManifest = registerDevServer({
       host: httpHost || 'localhost',
       port: appPort,
-      type: options.isApp ? 'app' : 'studio',
+      type: options.isApp ? 'coreApp' : 'studio',
       workDir: options.workDir,
     })
 
-    // Ensure manifest is cleaned up on abrupt shutdown
-    const onSignal = () => {
+    // Ensure manifest and workbench lock are cleaned up on abrupt shutdown.
+    // closeWorkbenchServer() starts with synchronous calls (watcher.close,
+    // lock.release) that complete before the process exits; the trailing
+    // async server.close() is best-effort.
+    onSignal = () => {
       cleanupManifest()
-      process.off('SIGINT', onSignal)
-      process.off('SIGTERM', onSignal)
+      closeWorkbenchServer()
+      process.off('SIGINT', onSignal!)
+      process.off('SIGTERM', onSignal!)
     }
     process.on('SIGINT', onSignal)
     process.on('SIGTERM', onSignal)
@@ -88,6 +93,11 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
 
   return {
     close: async () => {
+      // Remove signal handlers to prevent double-cleanup and listener leaks
+      if (onSignal) {
+        process.off('SIGINT', onSignal)
+        process.off('SIGTERM', onSignal)
+      }
       cleanupManifest()
       // Run both closes independently — a failing workbench close must not prevent
       // the primary server from shutting down
--- a/packages/@sanity/cli/src/actions/dev/devAction.ts
+++ b/packages/@sanity/cli/src/actions/dev/devAction.ts
@@ -1,11 +1,15 @@
 import {styleText} from 'node:util'
 
+import {registerDevServer} from './devServerRegistry.js'
 import {startAppDevServer} from './startAppDevServer.js'
 import {startStudioDevServer} from './startStudioDevServer.js'
 import {startWorkbenchDevServer} from './startWorkbenchDevServer.js'
 import {type DevActionOptions} from './types.js'
 
-export async function devAction(options: DevActionOptions): Promise<{close?: () => Promise<void>}> {
+const noop = async () => {}
+const syncNoop = () => {}
+
+export async function devAction(options: DevActionOptions): Promise<{close: () => Promise<void>}> {
   const {output} = options
 
   const {
@@ -17,14 +21,13 @@ export async function devAction(options: DevActionOptions): Promise<{close?: () 
 
   // Start app/studio dev server: use workbenchPort + 1 if workbench feature is
   // available (reserves the configured port for it), otherwise use the original port
-  const desiredAppPort = closeWorkbenchServer === undefined ? workbenchPort : workbenchPort + 1
+  const desiredAppPort = workbenchAvailable ? workbenchPort + 1 : workbenchPort
 
   // When the workbench is running, point the remote's react-refresh preamble at
   // the workbench dev server so HMR updates flow through the host.
-  const reactRefreshHost =
-    closeWorkbenchServer === undefined
-      ? undefined
-      : `http://${httpHost || 'localhost'}:${workbenchPort}`
+  const reactRefreshHost = workbenchAvailable
+    ? `http://${httpHost || 'localhost'}:${workbenchPort}`
+    : undefined
 
   const appOptions: DevActionOptions = {
     ...options,
@@ -33,31 +36,50 @@ export async function devAction(options: DevActionOptions): Promise<{close?: () 
     workbenchAvailable,
   }
 
-  let closeAppDevServer: (() => Promise<void>) | undefined
+  let closeAppDevServer: () => Promise<void> = noop
   let server
   try {
     const result = options.isApp
       ? await startAppDevServer(appOptions)
       : await startStudioDevServer(appOptions)
-    closeAppDevServer = result.close
+    closeAppDevServer = result.close ?? noop
     server = result.server
   } catch (err) {
-    await closeWorkbenchServer?.()
+    await closeWorkbenchServer()
     throw err
   }
 
   // server is undefined only when startAppDevServer exits early (e.g. missing orgId);
   // in that case the process is already exiting so no workbench needed.
   if (!server) {
-    return {
-      close: async () => {
-        await closeWorkbenchServer?.()
-      },
-    }
+    return {close: closeWorkbenchServer}
   }
 
-  if (closeWorkbenchServer !== undefined) {
-    const appPort = server.config.server.port
+  // Register the studio/app dev server in the registry (federated projects only)
+  let cleanupManifest: () => void = syncNoop
+  if (options.cliConfig?.federation?.enabled) {
+    const addr = server.httpServer?.address()
+    const appPort = typeof addr === 'object' && addr ? addr.port : server.config.server.port
+    cleanupManifest = registerDevServer({
+      host: httpHost || 'localhost',
+      port: appPort,
+      type: options.isApp ? 'app' : 'studio',
+      workDir: options.workDir,
+    })
+
+    // Ensure manifest is cleaned up on abrupt shutdown
+    const onSignal = () => {
+      cleanupManifest()
+      process.off('SIGINT', onSignal)
+      process.off('SIGTERM', onSignal)
+    }
+    process.on('SIGINT', onSignal)
+    process.on('SIGTERM', onSignal)
+  }
+
+  if (workbenchAvailable) {
+    const addr = server.httpServer?.address()
+    const appPort = typeof addr === 'object' && addr ? addr.port : server.config.server.port
     const workbenchUrl = `http://${httpHost || 'localhost'}:${workbenchPort}`
     output.log(
       `Workbench dev server started at ${styleText(['blue', 'underline'], workbenchUrl)} (app on port ${appPort})`,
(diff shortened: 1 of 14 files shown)


packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -33,31 +36,50 @@ export async function devAction(options: DevActionOptions): Promise<{close?: ()
...
-    return {
-      close: async () => {
-        await closeWorkbenchServer?.()
-      },
+    return {close: closeWorkbenchServer}
+  }
+
+  // Register the studio/app dev server in the registry (federated projects only)
+  let cleanupManifest: () => void = syncNoop
+  if (options.cliConfig?.federation?.enabled) {
```
gu-stav, line 61: I think we shouldn't differentiate based on whether the feature is enabled, because it leads to a situation where running `sanity dev` behaves differently, based on that flag.

Imo a better way of thinking about it could be that workbench always starts, but applications that don't support federation, are running in iframes.

Wdyt?

packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -66,9 +88,10 @@ export async function devAction(options: DevActionOptions): Promise<{close?: ()
 
   return {
     close: async () => {
+      cleanupManifest()
```
gu-stav, line 101: I feel like this should be part of `closeWorkbenchServer()` and `closeAppDevServer()`

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+
+const workbenchLockSchema = z.object({
+  host: z.string(),
+  pid: z.number(),
+  port: z.number(),
+})
+
+const devServerManifestSchema = z.extend(workbenchLockSchema, {
+  startedAt: z.string(),
+  type: z.enum(['app', 'studio']),
```
gu-stav: Think it would be good if this follows the Brett conventions for `coreApp` and `studio`.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+
+import {devDebug} from './devDebug.js'
+
+const workbenchLockSchema = z.object({
+  host: z.string(),
+  pid: z.number(),
+  port: z.number(),
+})
+
+const devServerManifestSchema = z.extend(workbenchLockSchema, {
```
gu-stav, line 29: I'd add a `version` field, to make it possible to upgrade the schema at some point.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+ */
+export type DevServerManifest = z.infer<typeof devServerManifestSchema>
+
+/**
+ * Returns the path to the dev server registry directory.
+ * Respects `SANITY_INTERNAL_ENV=staging` to isolate staging instances.
+ */
+function getRegistryDir(): string {
+  const suffix = process.env.SANITY_INTERNAL_ENV === 'staging' ? '-staging' : ''
+  return join(homedir(), `.sanity${suffix}`, 'dev-servers')
```
gu-stav: Since the CLI already stores e.g. auth information there: is there any common tooling available already to write this location? I don't know if it can be changed for example.

packages/@sanity/cli/src/actions/dev/devAction.ts
joshuaellis: I think for now, we should retain that workbench is opt in. If we plan to hand this out as an experimental feature for customers we'd enable it in a stable version of the CLI as opposed to a tagged version. This also retains the same practice we do in `startWorkbenchServer` where if its not enabled, we don't start that dev-server.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: sure

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: I thought about this, but i didn't feel like the config store was appropriate for this since its not really a configuration of the CLI?

gu-stav: I'd recommend to also move this into the same folder structure as we have for the runtime: `~/.sanity/workbench`

packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
```diff
@@ -65,12 +96,15 @@ export async function startWorkbenchDevServer(
...
-    remoteUrl = URL.parse(process.env.SANITY_INTERNAL_WORKBENCH_REMOTE_URL || '')?.toString()
+    remoteUrl = new URL(process.env.SANITY_INTERNAL_WORKBENCH_REMOTE_URL || '').toString()
   } catch {
     // Ignore parsing errors, the variable might not be set or might be an invalid URL, in which case we just won't use it
   }
 
   const viteConfig: InlineConfig = {
+    // Define a custom cache directory so that sanity's vite cache
+    // does not conflict with any potential local vite projects
+    cacheDir: 'node_modules/.sanity/vite',
```
gu-stav, line 107: Why not use `.sanity/cache` here, instead of introducing a new tmp location? Or does this already exist?

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+    try {
+      raw = JSON.parse(readFileSync(filePath, 'utf8'))
+    } catch {
+      continue
+    }
+
+    const {data, success} = devServerManifestSchema.safeParse(raw)
+    if (!success) continue
+
+    if (isProcessAlive(data.pid)) {
```
gu-stav: I'm wondering if there is a danger that previously generated <pid>.json files don't get properly cleaned up (e.g. because the process wasn't killed properly aka battery empty), but a new process later would get the same PID assigned again.

In this case you could point to the wrong server. I'm not sure how to mitigate that - is there something else on a process you could compare?

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+  const registryDir = getRegistryDir()
+  mkdirSync(registryDir, {recursive: true})
+
+  const lockPath = join(registryDir, 'workbench.lock')
+  const lockData = {host: info.host, pid: process.pid, port: info.port}
+
+  devDebug('Acquiring workbench lock at %s', lockPath)
+
+  try {
+    writeFileSync(lockPath, JSON.stringify(lockData), {flag: 'wx'})
```
gu-stav, line 255: Why does this file need to be executable?

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
```diff
@@ -0,0 +1,265 @@
...
+    return {
+      release() {
+        try {
+          unlinkSync(lockPath)
+        } catch {
+          // Already cleaned up
+        }
+      },
+      updatePort(port: number) {
+        writeFileSync(lockPath, JSON.stringify({...lockData, port}))
```
gu-stav, line 266: Two processes could overwrite each other, if another process has update the lockfile between calling `acquireWorkbenchLock` and `release()`, I think?

Would it be more safe to always e.g. append, and cleanup the file when a process exits?

packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
```diff
@@ -96,18 +130,39 @@ export async function startWorkbenchDevServer(
...
 
+  // Update the lock file with the actual port so other processes can find us
+  workbenchLock.updatePort(actualPort)
+
+  // Respond to client requests for the current application list
+  server.ws.on('sanity:workbench:get-local-applications', (_, client) => {
+    client.send(
+      'sanity:workbench:local-applications',
+      toApplicationsPayload(getRegisteredServers()),
+    )
```
gu-stav, line 152: I'd suggest to pass on the version number here too, to be able to upgrade the payload one day. This might become relevant, if e.g. a user is running an old version of the CLI, but a latest workbench (which is always latest).

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
gu-stav: I agree and wouldn't put it in the existing JSON-file, but I would suggest to extract the path generation from https://github.com/sanity-io/cli/blob/6b48e93e249e85b117f1113d110aa3544264fb7a/packages/%40sanity/cli-core/src/services/cliUserConfig.ts#L158-L165 to avoid any drift in the future.

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Ok yeah I agree its ok for now, but I think if we move closer to production, we maybe need to decouple this, because I might want to opt-out of federation, but not the OS?

Not sure - should we track it as a ticket for product/ user-research?

packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
joshuaellis: it already exists.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis:
```
 * Attempt to acquire an exclusive lock for the workbench process.
 * Uses `O_EXCL` (the `wx` flag) which is atomic at the OS level — only one
 * process can create the file.
```

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: i don't think this is 1-1 with federation. Why would we make it unique to federation? Maybe 1-1 with workbench, but i think this makes sense as is, its a list of sanity dev-servers running on said machine.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: no, the CLI cleans up zombies on each run when `getRegisteredServers` is called. So on a hard crash, the system would check if the processes are alive and clear them if not.  

However,  yes,  if a new process had the same PID then we'd have an issue, i think the best way to work around this case would be to process the start time. and compare that probably to the current process start time? the first dev-server has to run workbench and everything would register against it.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: we're cleaning up when the CLI closes, but not handling other forms of clean up. I think we should fix this then.

packages/@sanity/cli/src/actions/dev/devAction.ts
joshuaellis: No lets not use linear as tracking for this sort of thing, otherwise the backlog just becomes a massive list of "could dos" instead of a clear actionable list. We can discuss this before production release if we commit to the route.

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: 👍🏼

packages/@sanity/cli/src/actions/dev/startWorkbenchDevServer.ts
joshuaellis: 👍🏼

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Might also become tricky if we allow more customization capabilities for workbench other than application filtering - which workspace config do we apply, if a workbench with a different configuration is already running?

joshuaellis: Again, something to figure out when we actually implement customisation

packages/@sanity/cli/src/actions/dev/devAction.ts
joshuaellis: We could, but it'd mean duplicating the federation gate and signal handler setup across both server functions. The current pattern in devAction keeps cleanup coordination in one place — simpler than three functions each doing partial cleanup no?

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis: its not really 1:1 but i've pushed some changes for this

gu-stav (approved):
I'm a little unsure if we should move forward with this, because the overall approach of ideally only running one workbenhc is not compatible with the customization idea, as I've highlighted in comments.

I am wondering if we should park it and accept the fact that one workbench is running per application, since its sufficient for demoing workbench and work with product before doing this?

Code-wise it is ok - since this is quite big and introduces an entirely new concept of a lockfile, I'd prefer if someone else with more CLI experience could have a look, before merging it.

Since it goes into the feature branch first, I'm gonna approve it.

joshuaellis:
> I'm a little unsure if we should move forward with this, because the overall approach of ideally only running one workbenhc is not compatible with the customization idea, as I've highlighted in comments.

Well, it depends on how we set something up, i think in a Blueprint world it makes more sense overall. I think this PR clearly illusrates some of the core intents we want to put forward. So I think we should merge for now.

PR #948, opened 2026-04-14
fix(cli): resolve correct package for update checker notifications

mttdnt:
### Description

The update checker compared `@sanity/cli` versions from npm but told users to run `pnpm update sanity`. Since `@sanity/cli` and `sanity` are separate packages with independent version numbers, this was confusing and sometimes incorrect.

Now resolves which package the project actually depends on (`sanity` vs `@sanity/cli`), fetches the latest version of *that* package, compares against the *installed* version, and shows the matching update command.

Relates to #890 (package resolution only, without the background worker changes).

### What to review

- `resolveUpdateTarget.ts` — new module that determines which package to check (`sanity` if declared+installed, otherwise `@sanity/cli`)
- `updateChecker.ts` — wires in `resolveUpdateTarget` to fetch/compare the correct package, uses per-package cache key (`latestVersion:sanity` vs `latestVersion:@sanity/cli`)
- `getUpdateCommand.ts` / `showNotificationUpdate.ts` — accept `packageName` param instead of hardcoded `cliPkgName`
- `detectPackages.ts` — `workspaceInfo`/`workspaceRoot` params made optional so `resolveUpdateTarget` can call them without full workspace context

### Testing

- 40 tests passing across 4 test files (checkForUpdates, resolveUpdateTarget, getUpdateCommand, detectPackages)
- New tests for `resolveUpdateTarget` covering: sanity declared+installed, sanity declared but not installed, sanity not declared
- New tests for per-package cache key isolation
- New test for `@sanity/cli` fallback path through the full update checker flow

Resolves SDK-1249 and SDK-1250

PR #949, opened 2026-04-14
refactor(cli): move update check to non-blocking background worker

mttdnt:
### Description

Moves the npm version fetch from an inline blocking call to a detached background worker process. The main CLI process now only reads from the config cache — it never makes network requests during startup.

When the cache is empty or expired, a detached child process is spawned to fetch the latest version from npm and write it to the cache. The notification is shown on the next CLI invocation when the cached result is available instantly.

Also fixes a Windows bug in the worker entry-point check by using `pathToFileURL` instead of string concatenation for `import.meta.url` comparison.

Stacked on #948.

### What to review

- `fetchUpdateInfo.ts` — core worker logic: resolves target, fetches latest from npm with timeout, writes to config cache
- `fetchUpdateInfo.worker.ts` — thin entry point spawned as a detached child process
- `updateChecker.ts` — rewritten to read from cache + spawn worker instead of inline fetch
- `fetchLatestVersion.ts` — deleted, replaced by the worker approach
- `checkForUpdates.test.ts` — rewritten for the new cache-read + worker-spawn architecture

### Testing

- 15 tests passing across 2 test files (checkForUpdates, fetchUpdateInfo.worker)
- Worker test covers: fetch + cache write, @sanity/cli cache key, fetch failure, fetch timeout
- Integration test covers: spawning worker on empty/expired cache, showing notification from cache, cache key isolation, early returns (CI, non-TTY)


mttdnt:
* **#951** <a href="https://app.graphite.com/github/pr/sanity-io/cli/951?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#949** <a href="https://app.graphite.com/github/pr/sanity-io/cli/949?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/sanity-io/cli/949?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#948** <a href="https://app.graphite.com/github/pr/sanity-io/cli/948?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

binoy14:
Don't think this needs to be `feat:` maybe `refactor`

packages/@sanity/cli/src/util/update/fetchUpdateInfo.ts
```diff
@@ -0,0 +1,47 @@
...
+
+const debug = subdebug('updateChecker')
+
+const FETCH_TIMEOUT = 15_000
+
+/**
+ * Fetch the latest version of the update target package and write it to the config cache.
+ * Designed to run in a detached child process so it never blocks the main CLI.
+ */
+export async function runFetchWorker(cwd: string, cliVersion: string): Promise<void> {
```
binoy14: nit: should the function name be same as the file name

packages/@sanity/cli/src/util/update/__tests__/fetchUpdateInfo.worker.test.ts
```diff
@@ -0,0 +1,98 @@
...
+const mockConfigStore = vi.hoisted(() => ({
+  delete: vi.fn(),
+  get: vi.fn(),
+  set: vi.fn(),
+}))
+
+vi.mock('@sanity/cli-core', async () => ({
+  ...(await vi.importActual('@sanity/cli-core')),
+  getUserConfig: vi.fn(() => mockConfigStore),
+  subdebug: vi.fn(() => vi.fn()),
```
binoy14: does debug need to be mocked?

packages/@sanity/cli/src/util/update/__tests__/fetchUpdateInfo.worker.test.ts
```diff
@@ -0,0 +1,98 @@
...
+
+vi.mock('../../promiseRaceWithTimeout.js', () => ({
+  promiseRaceWithTimeout: mockPromiseRaceWithTimeout,
+}))
+
+// Import from the non-worker module (fetchUpdateInfo.ts) rather than the worker bundle.
+// Worker files (*.worker.ts) are bundled by esbuild with internal deps inlined,
+// which prevents vi.mock() from intercepting them. The core logic lives in
+// fetchUpdateInfo.ts so we can mock its dependencies normally.
+const {runFetchWorker} = await import('../fetchUpdateInfo.js')
```
binoy14: Seems to me that the test file name is not actually testing that file. It should match the file name it is testing

binoy14 (commented):
comments around function and file name consistency but looks good otherwise

binoy14 (approved):

PR #954, opened 2026-04-15
feat(dataset copy): add --skip-content-releases option

tzhelyazkova:
### Description

Adds support for skipping content release documents when copying a dataset.

### DIFF
--- a/.changeset/pr-954.md
+++ b/.changeset/pr-954.md
@@ -1,6 +1,5 @@
-<!-- auto-generated -->
 ---
 '@sanity/cli': minor
 ---
 
-add --skip-content-releases option
+Add `--skip-content-releases` flag to `sanity datasets copy` for excluding content release documents from the target dataset.
--- a/.changeset/pr-954.md
+++ b/.changeset/pr-954.md
@@ -1,0 +1,7 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+Add `--skip-content-releases` flag to `sanity datasets copy` for excluding content release documents from the target dataset.
+
--- a/.changeset/pr-954.md
+++ b/.changeset/pr-954.md
@@ -3,5 +3,4 @@
 '@sanity/cli': minor
 ---
 
-Add `--skip-content-releases` flag to `sanity datasets copy` for excluding content release documents from the target dataset.
-
+add --skip-content-releases option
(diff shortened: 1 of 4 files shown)


packages/@sanity/cli/src/commands/datasets/copy.ts
```diff
@@ -105,6 +109,11 @@ export class CopyDatasetCommand extends SanityCommand<typeof CopyDatasetCommand>
       description: 'Start position in the list of jobs (default 0)',
       required: false,
     }),
+    'skip-content-releases': Flags.boolean({
```
binoy14, line 112: wondering if this is better as `--no-content-releases` to keep the standard of `no` based boolean flags?

tzhelyazkova: IMO would be better to keep the syntax this command follows. We already a have `--skip-history` flag. So just keeping the consistency here.

.changeset/pr-954.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+add --skip-content-releases option
```
binoy14: Can you update this based on Claude's feedback

packages/@sanity/cli/src/commands/datasets/__tests__/copy.test.ts
```diff
@@ -420,6 +420,29 @@ describe('#dataset:copy', () => {
...
+      mockListDatasets.mockResolvedValue([
+        createMockDataset('production'),
+        createMockDataset('staging'),
+      ])
+      mockApi({
+        apiVersion: DATASET_API_VERSION,
+        method: 'put',
+        projectId: testProjectId,
+        uri: `/datasets/production/copy`,
+      }).reply(200, {jobId: 'job-no-releases'})
```
binoy14: Would it better to assert the body here? Because this test does not seem to be really testing anything besides it works

binoy14 (changes requested):
Looks good overall, just some nitpicky comments

.changeset/pr-954.md
tzhelyazkova: err, yes sure. I didn't write this in the first place thought, what generates that?

binoy14: It's based on the title of the commit, if one is not present it is automatically generated

binoy14:
@claude update the changeset content to be 

```
Add `--skip-content-releases` flag to `sanity datasets copy` for excluding content release documents from the target dataset.
```

also remove the auto-generated comment in it so it doesn't get lost

binoy14 (approved):

PR #955, opened 2026-04-15
feat: use oauth config for mcp with cursor

mwritter:
### Description

For `cursor` we want to use the default OAuth config for the mcp instead of using the API token authentication.  Base on the [docs](https://www.sanity.io/docs/ai/mcp-server#e4d6a37648fb) we only need to remove the authorization header from the config to use the default OAuth config.

AIGRO-4651


rexxars:
What's the rationale? How is cursor different than other editors here?

mwritter:
> What's the rationale? How is cursor different than other editors here?

@rexxars - Ideally we want to use OAuth for all MCP connections for less friction. Cursor prompts the user to log in with OAuth straight from the chat. The other agents/editors currently need the user to auth manually (see findings [here](https://linear.app/sanity/document/mcp-auth-prompts-3760f9632311)). Since Cursor works well with OAuth, we want users to use OAuth instead of the API token.

jonahsnider (dismissed):

mwritter:
> #### 1. Type inconsistency between interface and usage (`writeMCPConfig.ts:24`, `editorConfigs.ts:37`)
> The `EditorConfig` interface declares `buildServerConfig: (token: string) => Record<string, unknown>` (token required), but `defaultHttpConfig` accepts `token?: string`. This mismatch forces an awkward workaround in `writeMCPConfig`:
> 
> ```ts
> // writeMCPConfig.ts:24 — calling with '' for oauthOnly or ! for others
> const serverConfig = oauthOnly ? buildServerConfig('') : buildServerConfig(token!)
> ```
> 
> The empty string `''` works at runtime (falsy, so no header is added), but `token!` suppresses a legitimate type error — if `oauthOnly` is false and `token` is `undefined`, this would silently produce a broken config.
> 
> Fix: update the interface and call site to be explicit:
> 
> ```ts
> // editorConfigs.ts interface
> buildServerConfig: (token?: string) => Record<string, unknown>
> 
> // writeMCPConfig.ts
> const serverConfig = buildServerConfig(oauthOnly ? undefined : token)
> ```
> 


This was considered but the refactor would have been more work that I think its worth.

binoy14 (approved):

PR #971, opened 2026-04-17
fix(workbench): externalize sanity and @sanity/workbench

gu-stav:
### Description

Exclude sanity/workbench (and its transitive dep @sanity/workbench) from dep pre-bundling so that `import.meta.hot` is available at runtime — pre-bundled modules do not receive Vite's HMR client injection, which causes the custom HMR events for local application discovery to silently not fire.

This solves the injection and applications are coming through now, but the list is empty. Going to fix this in a follow-up.

### DIFF
--- a/.changeset/fancy-cycles-strive.md
+++ b/.changeset/fancy-cycles-strive.md
@@ -1,6 +1,5 @@
-<!-- auto-generated -->
 ---
-'@sanity/cli': patch
+"@sanity/cli": patch
 ---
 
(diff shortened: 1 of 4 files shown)


gu-stav:
@claude fix the failing automated changeset

> @sanity/cli-monorepo@0.0.1 check:format /home/runner/work/cli/cli
> oxfmt --check

Checking formatting...

.github/workflows/snapshot-release.yml (14ms)

joshuaellis (approved):

PR #972, opened 2026-04-17
fix(workbench): use the same clock for workbench lock

gu-stav:
### Description

`registerDevServer` and `acquireWorkbenchLock` recorded `startedAt` as `new Date()` at call time. But `isOurProcess` validates a manifest/lock by comparing `startedAt` to the OS-reported process start time from `ps -o lstart=`, with a 2-second tolerance.

In real sanity dev runs, registration happens several seconds after the process starts (after the workbench + app Vite servers boot). The drift between call time and OS start time routinely exceeds 2s, so the fs-watcher re-read the file, declared it stale, and deleted it — leaving the workbench with an empty application list.

Fix: store `getProcessStartTime(process.pid)` (same source `isOurProcess` compares against) as startedAt, falling back to new Date() when ps is unavailable. Stored value and verification reference now come from the same clock.


gu-stav:
@claude fix the auto-generated changeset file formatting

.github/workflows/snapshot-release.yml
```diff
@@ -76,4 +76,4 @@ jobs:
         env:
           DIST_TAG: ${{ inputs.tag }}
           NODE_AUTH_TOKEN: ${{ secrets.NPM_PUBLISH_TOKEN }}
-          NPM_CONFIG_PROVENANCE: true
\ No newline at end of file
+          NPM_CONFIG_PROVENANCE: true
```
gu-stav, line 79: Unrelated, but fixes lint workflow

joshuaellis (approved):

PR #976, opened 2026-04-17
test(cli-test): add --ignore-workspace to fixture pnpm install

binoy14:
### Description

Fixes E2E test failures in registry mode (workflow_dispatch) where tests using `testFixture()` fail with "CLI config cannot be loaded".

**Root cause:** `setupTestFixtures` runs `pnpm install --no-lockfile` in fixture temp directories located inside the monorepo workspace (`packages/@sanity/cli-e2e/tmp/`). Pnpm detects the workspace root and runs a workspace-level install instead of installing dependencies in the fixture directory. This means fixtures never get their own `node_modules/sanity`, so the CLI fails when it tries to import `sanity.cli.ts` (which requires `sanity/cli` from `node_modules`).

**Fix:** Add an `ignoreWorkspace` option to `SetupTestFixturesOptions` that passes `--ignore-workspace` to pnpm install. The E2E globalSetup opts in (`ignoreWorkspace: true`), while the main test suite uses the default (`false`) to preserve existing behavior. This avoids the Windows Node 20 open-handle hang that occurs when pnpm installs full dependency trees into all 9 fixture directories.

### What to review

- `packages/@sanity/cli-test/src/test/setupFixtures.ts` — new `ignoreWorkspace` option on `SetupTestFixturesOptions`
- `packages/@sanity/cli-e2e/globalSetup.ts` — consolidated globalSetup that handles CLI pack/install and fixture setup with `ignoreWorkspace: true`
- `packages/@sanity/cli-e2e/vitest.config.ts` — removed separate `@sanity/cli-test/vitest` globalSetup entry (now handled by the consolidated globalSetup)

### Testing

- Reproduced the failure locally by setting up a fixture inside the workspace and running the published CLI against it — confirmed "CLI config cannot be loaded" error
- Verified the fix resolves the issue — CLI successfully loads config after `--ignore-workspace` is added
- Verified that the default path (no `--ignore-workspace`) preserves existing behavior for the main test suite
- Type checking passes for both `@sanity/cli-test` and `@sanity/cli-e2e`

### DIFF
--- a/.changeset/pr-976.md
+++ b/.changeset/pr-976.md
@@ -1,6 +1,0 @@
-<!-- auto-generated -->
----
-'@sanity/cli-test': patch
----
-
-add --ignore-workspace to fixture pnpm install
--- a/.changeset/pr-976.md
+++ b/.changeset/pr-976.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-test': patch
+---
+
+add --ignore-workspace to fixture pnpm install
--- a/packages/@sanity/cli-e2e/globalSetup.ts
+++ b/packages/@sanity/cli-e2e/globalSetup.ts
@@ -2,14 +2,16 @@ import {mkdtempSync, rmSync} from 'node:fs'
 import {tmpdir} from 'node:os'
 import {join} from 'node:path'
 
+import {setup as setupFixtures, teardown as teardownFixtures} from '@sanity/cli-test/vitest'
 import {config as loadDotenv} from 'dotenv'
+import {type TestProject} from 'vitest/node'
 
 import {installFromTarball, packCli} from './helpers/packCli.js'
 
 let cleanupDir: string | undefined
 let tarballPath: string | undefined
 
-export async function setup(): Promise<void> {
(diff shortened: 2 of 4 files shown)


mttdnt (dismissed):

.changeset/pr-976.md
ryanbonial, line 1: Is this file meant to be checked in?

ryanbonial (dismissed):

.changeset/pr-976.md
binoy14: not a big deal since cli-test is technically public

binoy14:
Updated the PR to make `--ignore-workspace` opt-in rather than unconditional.

The original change applied `--ignore-workspace` to all fixture setups, which caused vitest to hang indefinitely on Windows Node 20 — pnpm installed full dependency trees into all 9 fixture directories, leaving open handles that prevented the process from exiting.

Now `--ignore-workspace` is behind an `ignoreWorkspace` option on `SetupTestFixturesOptions`. Only the E2E globalSetup opts in (E2E only runs on Ubuntu so the Windows hang isn't a concern). The main test suite uses the default (`false`), preserving existing behavior.

mttdnt (approved):

PR #980, opened 2026-04-20
feat(cli): show runner-specific update command for npx/dlx/bunx

mttdnt:
### Description

Detects when the Sanity CLI is invoked via a package runner (npx, pnpm dlx, yarn dlx, bunx) and swaps the command shown in the update notification for one the user can actually run through that runner (e.g. `pnpm dlx sanity@latest` instead of `npm update sanity`). Resolves [SDK-1252](https://linear.app/sanity-io/issue/SDK-1252).

The runner is identified by inspecting `process.argv[1]` for known temp-cache layouts. The invoked package name (`sanity` vs `@sanity/cli`) is recovered by realpath'ing the `.bin/sanity` symlink into the runner's install and reading the nearest `package.json`, so the hint names whichever package the user typed.

> **Note:** while auditing this path I noticed an unrelated regression — the once-per-12h update-notification dedupe was silently dropped in #890 and not re-added in #949. That fix is in a stacked PR: `fix/restore-notification-dedupe`.

### What to review

- `packages/@sanity/cli/src/util/update/packageRunner.ts` — path patterns per runner, including false-positive guards (`/xfs-` + `/dlx-` for yarn, `/bunx-\d+-` for bun) so local dirs like `/projects/dlx-proxy/` don't match.
- `packages/@sanity/cli/src/util/update/resolveRunnerPackage.ts` — follows `argv[1]` through its symlink and walks up to find the real package.json; returns `{packageName, installedVersion}`.
- `packages/@sanity/cli/src/util/update/getRunnerUpdateCommand.ts` — maps runner + package name → exact update command shown. Uses a `BIN_NAMES` record so the yarn `-p` decision asks whether bin equals package name rather than hardcoding `'sanity'`.
- `packages/@sanity/cli/src/util/update/updateChecker.ts` — branches on runner detection; under a runner uses `config.version` + resolved package, otherwise keeps the existing cwd-walk.
- `packages/@sanity/cli/src/util/update/fetchUpdateInfo.{ts,worker.ts}` — worker is pinned to the main-resolved package name via `SANITY_UPDATE_CHECK_PACKAGE`, validated against the `SanityPackage` union at read time.
- `packages/@sanity/cli/src/util/update/showNotificationUpdate.ts` — accepts an optional runner and uses the runner command.

### Testing

Unit tests cover:

- every runner's temp path shape + false-positive guards (`dlx-proxy`, `bunx-demo`, partial `/xfs-` or `/dlx-` matches)
- the symlink → package.json walk with fixture dirs for `sanity` and `@sanity/cli`, plus walk-past-unrelated-package.json and non-string `version` fallback
- every runner → update command mapping (including the yarn-dlx with-`-p`/without-`-p` branches)
- the exhaustiveness guard in `getRunnerUpdateCommand`
- the `checkForUpdates` hook end-to-end via `test.each` across all four runners
- main/worker cache-key alignment by asserting the worker env var contains the resolved package name

End-to-end verified against a pkg.pr.new preview on macOS: `npx`, `pnpm dlx`, `yarn dlx`, and `bunx` each rendered the expected notification with the correct runner command.


.changeset/runner-specific-update-command.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sanity/cli': minor
+---
+
+Show a runner-specific update command in the update notification when the CLI is invoked via npx, pnpm dlx, yarn dlx, or bunx. Also restores the "only notify once per version" dedupe that was unintentionally dropped when the update checker was rewritten to use a background worker, so the notification no longer fires on every invocation while an update is available.
```
binoy14: claude's comment about this is valid? I am also wondering if this is a `feat` or `fix` ? 🤔

packages/@sanity/cli/src/util/update/packageRunner.ts
```diff
@@ -0,0 +1,14 @@
+export type PackageRunner = 'bunx' | 'npx' | 'pnpm-dlx' | 'yarn-dlx'
+
+export function detectPackageRunner(
+  binaryPath: string = process.argv[1] ?? '',
+): PackageRunner | null {
+  const normalized = binaryPath.replaceAll('\\', '/')
+
+  if (normalized.includes('/_npx/')) return 'npx'
+  if (normalized.includes('/pnpm/dlx/')) return 'pnpm-dlx'
+  if (normalized.includes('/xfs-') && normalized.includes('/dlx-')) return 'yarn-dlx'
```
binoy14, line 10: confirming if yarn 1 uses the same?

packages/@sanity/cli/src/util/update/resolveRunnerPackage.ts
```diff
@@ -0,0 +1,42 @@
...
+  packageName: SanityPackage
+}
+
+export async function resolveRunnerPackage(
+  binaryPath: string = process.argv[1] ?? '',
+  fallbackVersion = '',
+): Promise<RunnerPackage> {
+  try {
+    let dir = dirname(await realpath(binaryPath))
+    while (dir !== resolve(dir, '..')) {
```
binoy14: maybe add a comment here about what's happening?

packages/@sanity/cli/src/util/update/packageRunner.ts
mttdnt: I dont think yarn 1 has a temp package runner like this, there is `yarn create`, but in the case of yarn 1 with the cli it would be through a local or global install.

.changeset/runner-specific-update-command.md
mttdnt: Yeah I think I am trying to shove too much into this pr, I am going to pull it out as a separate fix

packages/@sanity/cli/src/util/update/resolveRunnerPackage.ts
```diff
@@ -0,0 +1,49 @@
...
+ */
+export async function resolveRunnerPackage(
+  binaryPath: string = process.argv[1] ?? '',
+  fallbackVersion = '',
+): Promise<RunnerPackage> {
+  try {
+    // Follow the runner's .bin/sanity symlink to the real bin file, then walk
+    // up until we hit a package.json for a known Sanity package.
+    let dir = dirname(await realpath(binaryPath))
+    while (dir !== resolve(dir, '..')) {
```
binoy14: Should this have a max iterations to avoid an infinite loop. Maybe we do this for findProjetRoot?

packages/@sanity/cli/src/util/update/resolveRunnerPackage.ts
```diff
@@ -0,0 +1,55 @@
...
+import {dirname, resolve} from 'node:path'
+
+import {subdebug} from '@sanity/cli-core'
+
+import {type SanityPackage} from '../packageManager/installationInfo/types.js'
+
+const debug = subdebug('updateChecker')
+
+const KNOWN_PACKAGES = new Set<SanityPackage>(['@sanity/cli', 'sanity'])
+const MAX_WALK_ITERATIONS = 50
```
binoy14: from 🤖 

```
A realistic, valid max here is ~20–30, with 25 being a good default.

Why:

You’re walking directory parents looking for a package.json that matches @sanity/cli or sanity.
In typical Node layouts, the .bin/sanity target lives somewhere like:
…/node_modules/.bin/sanity → …/node_modules/sanity/bin/sanity.js
or pnpm’s deeper store paths, but still usually well under ~15–20 parent hops before you reach the package root containing package.json.
The loop already also stops when it hits filesystem root (dir !== resolve(dir, '..')), so the cap is mainly a “don’t accidentally spin forever / do too much work if something is weird” guard.
What I’d pick:

25: “safe” across npm/yarn/pnpm + odd nesting without being huge.
50 (what you have) isn’t wrong, it’s just more than you likely need; worst-case it’s 50 filesystem reads/stat-like operations, which is still not enormous, but it’s unnecessary.
If you want it to be slightly more principled, you can also tie it to path depth as a guardrail (eg min(50, segments(binaryPath).length + 5)), but a fixed 25 is usually the simplest and still robust.
```

packages/@sanity/cli/src/util/update/__tests__/resolveRunnerPackage.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    const binLink = join(binDir, 'sanity')
+    await symlink(join(pkgDir, 'bin', 'sanity'), binLink)
+
+    expect(await resolveRunnerPackage(binLink, '9.9.9')).toEqual({
+      installedVersion: '9.9.9',
+      packageName: 'sanity',
+    })
+  })
+
+  test('walks past an unrelated package.json before finding a sanity package', async () => {
```
binoy14, line 101: There isn't a test for falling back at max iterations. Not sure how hard it is to mock/test but worth looking into

binoy14 (dismissed):
approved with comments, okay with a future PR for this

binoy14 (approved):

PR #987, opened 2026-04-21
feat(workbench): extend local application payloads

gu-stav:
### Description

Forward more configuration values from `sanity.cli.ts`:

- `deployment.appId` and `app.id`, with the former taking precedence if both are defined
- `app.title` and `app.icon`; the former only for SDK applications
- `server.hostname` if set + derived from the Vite configuration, to take user-provided and merged Vite configurations into account

This will allow us to enrich the display of local applications in dock, by displaying their actual title (if set) rather than a static localhost string.

We will be able to match local applications to remote applications in Brett via their id and group them together visually.

Enables https://github.com/sanity-io/workbench/pull/128

### DIFF
--- a/packages/@sanity/cli/src/actions/dev/devAction.ts
+++ b/packages/@sanity/cli/src/actions/dev/devAction.ts
@@ -1,6 +1,6 @@
 import {styleText} from 'node:util'
 
-import {normalizeAppId} from '../../util/appId.js'
+import {checkForDeprecatedAppId, getAppId} from '../../util/appId.js'
 import {readIconFromPath} from '../manifest/extractAppManifest.js'
 import {registerDevServer} from './devServerRegistry.js'
 import {startAppDevServer} from './startAppDevServer.js'
@@ -15,7 +15,7 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
   const {output} = options
 
   if (options.cliConfig) {
-    normalizeAppId({cliConfig: options.cliConfig, output})
+    checkForDeprecatedAppId({cliConfig: options.cliConfig, output})
   }
 
   const {
@@ -93,7 +93,7 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
     cleanupManifest = registerDevServer({
       host: appHost,
       icon,
-      id: options.cliConfig?.deployment?.appId,
+      id: getAppId(options.cliConfig),
       port: appPort,
       title: options.isApp ? options.cliConfig?.app?.title : undefined,
       type: options.isApp ? 'coreApp' : 'studio',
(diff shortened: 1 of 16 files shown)


joshuaellis:
Sorry, GH is being weird and now i see every commit/file change for the feature branch 🫠 can we rebase and then i can review? 😄

gu-stav:
@joshuaellis rebased ✅

packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -61,9 +68,34 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
...
+      } catch (err) {
+        output.warn(
+          `Could not inline app icon for workbench discovery: ${err instanceof Error ? err.message : String(err)}`,
+        )
+      }
+    }
+
     cleanupManifest = registerDevServer({
-      host: httpHost || 'localhost',
+      host: appHost || 'localhost',
```
joshuaellis: im not sure `appHost` is correct, it probably should be derived from `addr` which comes from the application dev-server not Workbench

packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -61,9 +68,34 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
...
+      icon,
+      id: options.cliConfig?.deployment?.appId,
       port: appPort,
+      title: options.cliConfig?.app?.title,
```
joshuaellis: studios won't have this and i think we specifically didn't enable this field to be pushed into writing `UserApplication.title` for studios, can we extract the title from the `sanity.config` maybe? 🤔

packages/@sanity/cli/src/actions/dev/devServerRegistry.ts
joshuaellis, line 1: i incorrectly wrote this in the other PR geting confused which one i was looking at, but i wonder if `icon` and `title` should be in a `manifest` so it matches the shape of `UserApplication` somewhat 🤔

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Discussed: derive from the actual applied vite config (from the dev-server) - as late as possible.

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Discussed: only do it for `isApp` this way and leave empty for studios for now, until we understand what the name of a studio is

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Its now derived from the server.

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Done - `type: studio` now sends `undefined`.

joshuaellis (approved):

PR #988, opened 2026-04-21
feat(init): add promt for federation

gu-stav:
### Description

Adds a new prompt about whether a user wants to enable federation for a given up, to the sanity init command, that is being triggered when running `npm create sanity@latest`. In unattended mode, it defaults to `true`.

The idea is that users will be able to run `npm create sanity@workbench` once this is merged, to ease getting started with module-federation internally.

### What to review

Should we stick to copy around "federation" or should we frame it as workbench, since "federation" is the technology, but not the value this promises?

### DIFF
--- a/.changeset/pr-988.md
+++ b/.changeset/pr-988.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+add promt for federation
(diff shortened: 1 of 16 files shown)


joshuaellis (approved):
seeing it in a prompt makes me wonder if we should have called it workbench lol

gu-stav:
@joshuaellis Agree. Counter argument imo is, that "workbench" has even less meaning. I guess this ties into the question about how we want to message this overall. "Run as part of the operation system?" doesn't have much meaning either. 🤔

PR #991, opened 2026-04-21
chore(cli): restore once-per-cache-refresh update-notification dedupe

mttdnt:
### Description

Restores the "only notify once per cache refresh" dedupe on the CLI update checker. The pre-worker `updateChecker` used `onCacheHit: () => { showNotificationUpdate = false }` to suppress notifications when the cache was warm, so a user saw at most one prompt per 12-hour cache window. That invariant was silently dropped in #890 (background worker rewrite) and not re-added in #949. The notification has been firing on every CLI invocation while an update is available.

Fix: store the last notified latest version under `notifiedAt:<package>` in the user config. The notification fires the first time a given latest version is seen and stays silent on subsequent invocations until the cache expires.

### What to review

- `packages/@sanity/cli/src/util/update/updateChecker.ts` — new `notifiedVersion:<package>` read/write around `showUpdateNotification`.
- `packages/@sanity/cli/src/hooks/init/__tests__/checkForUpdates.test.ts` — `test.each` covering both `sanity` and `@sanity/cli` packages for the once-per-version behavior; a follow-up test asserts the notification fires again when the cached latest bumps.

### Testing

Unit tests:

- `does not show notification twice for the same cache period` parameterised across both packages.
- `shows notification again when cache expires` asserts the gate releases on a version bump.

Would have failed on the pre-fix code.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to client-side update-notification gating using persisted config keys plus added unit tests; main risk is unintentionally suppressing or re-showing notifications due to key/`updatedAt` mismatches.
> 
> **Overview**
> Restores **once-per-cache-refresh** deduping for CLI update prompts by persisting a `notifiedAt:<package>` marker and skipping `showUpdateNotification` when the cached `latestVersion` entry’s `updatedAt` has already been notified.
> 
> Updates the cached-version reader to return `updatedAt` and adds tests to ensure notifications fire only once per cache cycle (for both `sanity` and `@sanity/cli`) and reappear after the cache is refreshed with a new `updatedAt`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d7690556a8833977ebf288aedf35fe59380c388a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/restore-notification-dedupe.md
+++ b/.changeset/restore-notification-dedupe.md
@@ -1,5 +1,0 @@
----
-'@sanity/cli': patch
----
-
-The update notification now fires at most once per cache refresh (~once per 12 hours) instead of on every CLI invocation while an update is available.
--- a/.changeset/restore-notification-dedupe.md
+++ b/.changeset/restore-notification-dedupe.md
@@ -2,4 +2,4 @@
 '@sanity/cli': patch
 ---
 
-Restore the once-per-version update-notification dedupe. The notification now fires once per new latest version instead of on every CLI invocation while an update is available.
+The update notification now fires at most once per cache refresh (~once per 12 hours) instead of on every CLI invocation while an update is available.
--- a/.changeset/restore-notification-dedupe.md
+++ b/.changeset/restore-notification-dedupe.md
@@ -1,0 +1,5 @@
+---
+'@sanity/cli': patch
+---
+
+Restore the once-per-version update-notification dedupe. The notification now fires once per new latest version instead of on every CLI invocation while an update is available.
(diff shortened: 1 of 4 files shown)


.changeset/restore-notification-dedupe.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sanity/cli': patch
+---
+
+The update notification now fires at most once per cache refresh (~once per 12 hours) instead of on every CLI invocation while an update is available.
```
binoy14: Is this needed if the PRs are never merged? Just a nitpick but we should get in the habit of being intentional about changesets

binoy14 (dismissed):

binoy14 (approved):

PR #997, opened 2026-04-22
feat(dev): extract manifests for local applications

gu-stav:
### Description

Extract the manifest for studio and core applications, when fedeation is enabled. For studios we extract the static manifest, like before. For core apps we fake a static manifest contract, because they never had static manifests.

The manifests extraction happens in the background, to not block the dev-server startup because it takes quite a while. This means initially it is possible that the list of applications in the dock does not have the full information available and it will be send once the extraction has finished, which imo is an acceptable tradeoff.

Counterpart in workbench https://github.com/sanity-io/workbench/pull/144

---

> [!NOTE]
> **Medium Risk**
> Touches dev-server lifecycle/registry IO and introduces async background extraction + file watching, which can affect workbench discovery and shutdown/cleanup behavior. Deploy-time manifest extraction was renamed/typed but remains optional, limiting production impact.
> 
> **Overview**
> When federation is enabled, `sanity dev` now registers studios and core apps in the local dev-server registry immediately *without* metadata, then **inlines a manifest asynchronously** and patches the registry to trigger workbench rebroadcasts. Studios use a new `startDevManifestWatcher` to run (and debounce) worker-based manifest extraction on `sanity.config.*` changes, while core apps run `extractCoreAppManifest` once in the background.
> 
> The dev-server registry format is updated to store `manifest` + `manifestUpdatedAt`, and `registerDevServer` now returns `{release, update}` with safeguards so late background updates don’t recreate deleted entries; workbench payloads are updated accordingly. Manifest extraction is refactored to accept explicit `workDir`/config `path` (used by the `manifest extract` command and dev watcher), core-app manifests are validated with a zod schema, and Vite cache dirs are unified under the new `SANITY_CACHE_DIR` constant (used for vendor builds, studio builds, and workbench).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ff3f547b36557b421b6d1e0b5f17fcd97e9ee84c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/manifest/types.ts
+++ b/packages/@sanity/cli/src/actions/manifest/types.ts
@@ -20,12 +20,28 @@ export interface CreateManifest {
   workspaces: ManifestWorkspaceFile[]
 }
 
-export interface AppManifest {
-  version: '1'
+/**
+ * Core-app application manifest. Mirrors the workbench's
+ * `CoreAppUserApplicationManifest` schema. Strictly validated via zod
+ * since the CLI produces the payload in full.
+ */
+export const coreAppManifestSchema = z.object({
+  icon: z.optional(z.string()),
+  title: z.optional(z.string()),
+  version: z.string(),
+})
 
-  icon?: string
-  title?: string
-}
+export type CoreAppManifest = z.infer<typeof coreAppManifestSchema>
+
+/**
+ * Studio application manifest (serialized `create-manifest.json`). Kept
+ * loose so the CLI isn't coupled to the workbench's evolving client-side
+ * schema — the workbench consumer is authoritative on the inner shape.
+ * See its `ClientManifest` for the fields clients expect.
+ */
+export const studioManifestSchema = z.record(z.string(), z.unknown())
+
+export type StudioManifest = z.infer<typeof studioManifestSchema>
 
 export interface ManifestWorkspaceFile extends Omit<CreateWorkspaceManifest, 'schema' | 'tools'> {
   schema: string // filename
--- a/.changeset/pr-997.md
+++ b/.changeset/pr-997.md
@@ -3,4 +3,4 @@
 '@sanity/cli': minor
 ---
 
-extract studio manifest and pass it for local applications
+extract manifests for local applications
--- a/.changeset/pr-997.md
+++ b/.changeset/pr-997.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+extract studio manifest and pass it for local applications
--- a/packages/@sanity/cli/src/actions/build/buildVendorDependencies.ts
+++ b/packages/@sanity/cli/src/actions/build/buildVendorDependencies.ts
@@ -195,7 +195,7 @@ export async function buildVendorDependencies({
     },
     // Define a custom cache directory so that sanity's vite cache
     // does not conflict with any potential local vite projects
-    cacheDir: path.join(SANITY_CACHE_DIR, 'vite-vendor'),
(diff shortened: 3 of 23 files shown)


packages/@sanity/cli/src/actions/manifest/types.ts
```diff
@@ -20,12 +20,28 @@ export interface CreateManifest {
...
+
+/**
+ * Studio application manifest (serialized `create-manifest.json`). Kept
+ * loose so the CLI isn't coupled to the workbench's evolving client-side
+ * schema — the workbench consumer is authoritative on the inner shape.
+ * See its `ClientManifest` for the fields clients expect.
+ */
+export const studioManifestSchema = z.record(z.string(), z.unknown())
+
+export type StudioManifest = z.infer<typeof studioManifestSchema>
```
joshuaellis, line 44: isnt this actually supposed to be CreateWorkspaceManifest? Maybe im misunderstanding the intent here.

joshuaellis (commented):
I do worry the extraction of `SANITY_CACHE_DIR` & change of appManifest -> coreAppManifest will make the main feature branch harder to reconcile. Would it be worth upstreaming these since they're not _strictly_ related to this feature branch changes? Probs an easy job for claude 🤖

packages/@sanity/cli/src/actions/manifest/types.ts
gu-stav: It is. My thinking about not typing it strictly is that the _remote_ shape may change one day which would break existing CLIs, because the validation no longer matches.

The manifest for core apps is hard-coded, but for studios it depends on the remote and I thought a loops coupling would be a lot less dangerous?

Right now I think there is no strict type for the studio manifest in the CLI, because of that.

gu-stav:
> Would it be worth upstreaming these since they're not strictly related to this feature branch changes? Probs an easy job for claude 🤖

@joshuaellis can do - good idea 👍🏼 . Should we try and get this merged first? I can take care of upstreaming the changes afterwards and do a rebase of the feature branch.

packages/@sanity/cli/src/actions/manifest/types.ts
joshuaellis: > Right now I think there is no strict type for the studio manifest in the CLI, because of that.

There is `CreateManifest` that is the strict type no?

>  that the remote shape may change one day which would break existing CLIs, because the validation no longer matches.

I don't understand what you mean, we're generating the manifest as the CLI can now so its types can be used here as well right?

I think i'm not getting what you're trying to explain here.

gu-stav: `CreateManifest` is just a TS-type, but not a zod schema, so during runtime there is no validation whether the shape of the manifest matches. The zod validation schema just checks, whether its a record for now.

We can also drop the zod validation for the studio manifest entirely - I just left in there, because then its the same for both application types.

joshuaellis: i see, sorry i misunderstood. okay lets keep as is.

joshuaellis (approved):

gu-stav:
I've extracted the changes for `main`: https://github.com/sanity-io/cli/pull/1008

packages/@sanity/cli/src/actions/manifest/types.ts
gu-stav: Yeah sorry. Re-reading my first answer I must admit it wasn't clear. Apologies!

PR #1006, opened 2026-04-22
fix(init): respect --project and --dataset flags for app-quickstart template

binoy14:
### Description

Fixes [SDK-1314](https://linear.app/sanity/issue/SDK-1314/yes-still-prompts-for-project-id-and-dataset) and the related interactive-mode regression.

Previously, running `sanity init --template app-quickstart --project <id> --dataset <name>` would still prompt the user to select a project (and an organization), ignoring the supplied flags. Unattended mode (`--yes`) also required `--organization` even when `--project` already supplied enough context to derive it.

**The fix** (in `packages/@sanity/cli/src/commands/init.ts`):

- `promptForAppTemplateSetup` now takes a flag-driven path whenever `--project` or `--project-name` is provided — in both interactive and unattended modes. This matches the behavior of the non-app template flow and stops the spurious "Configure a project for this app?" prompt.
- The organization is derived from the resolved project when not explicitly supplied, so the org prompt is skipped and `--organization` is no longer required in unattended mode when a project is specified.
- `getOrCreateProject` and `promptForProjectCreation` now return the `organizationId` of the resolved/created project so callers can propagate it.

### What to review

- `packages/@sanity/cli/src/commands/init.ts` — the early-return path in `promptForAppTemplateSetup`, the conditional skip of the org prompt in `getProjectDetails`, and the relaxed unattended validation in `checkFlagsInUnattendedMode`.
- Confirm the derived-org behavior is acceptable UX (we silently use the project's org rather than asking the user to confirm).

### Testing

Three new tests added in `init.get-project-details.test.ts`:

1. **Interactive + `--project` + `--dataset`** — neither the org selection nor the "Configure a project" prompt fires; existing dataset is reused.
2. **Unattended + `--project` + `--dataset` without `--organization`** (SDK-1314) — succeeds with no prompts, org derived from the project.
3. **Unattended without `--project` nor `--organization`** — still errors, but with an updated message mentioning `--project` / `--project-name` as alternatives.

All 93 `init` tests pass; the wider `@sanity/cli` suite has one pre-existing unrelated failure in `cliInstallationCheck.test.ts` (reproduced on `main`). Typecheck, lint, and build all green.

Manual testing not performed against a live Sanity project in this branch — would appreciate a reviewer sanity-checking the flow against a real org/project before merge.

### DIFF
--- a/.changeset/pr-1006.md
+++ b/.changeset/pr-1006.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+respect --project and --dataset flags for app-quickstart template
--- a/packages/@sanity/cli/src/commands/__tests__/init/init.get-project-details.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/init/init.get-project-details.test.ts
@@ -1179,4 +1179,117 @@ describe('#init: promptForAppTemplateSetup', () => {
     expect(mocks.listProjects).not.toHaveBeenCalled()
     expect(mocks.listDatasets).not.toHaveBeenCalled()
   })
+
+  test('interactive with --project and --dataset: skips the app-template selection prompt and the organization prompt', async () => {
+    // getOrCreateProject fetches projects and organizations in parallel
+    mocks.listProjects.mockResolvedValueOnce([
+      {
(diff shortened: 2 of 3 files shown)


binoy14:
#1015

PR #1009, opened 2026-04-23
feat(workbench): print app url

gu-stav:
### Description

Instead of printing the URL to the workspace, we now print the URL to the application that has been started. This avoids the surprise that when users click the link they land on the homepage of workbench, rather than the application, which is an extra click:

Screenshot 2026-04-23 at 15 43 39

### Testing

---

> [!NOTE]
> **Low Risk**
> Low risk: this is a CLI log output change (plus moving host resolution so it’s always available), with no impact on server behavior beyond what URL is printed to users.
> 
> **Overview**
> When running `sanity dev` with workbench enabled, the CLI now prints a direct **application URL** (workbench `local` route including `${appHost}-${appPort}`) instead of the generic workbench root URL, reducing the extra click to reach the started app.
> 
> The resolved Vite `host` used to build the app URL is computed unconditionally (not only in federation mode), ensuring consistent URL construction regardless of federation settings, and a changeset bumps `@sanity/cli` with a minor release note.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4f17a3c4f42b07cd1590d74d11b602a6be9542d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/dev/devAction.ts
+++ b/packages/@sanity/cli/src/actions/dev/devAction.ts
@@ -63,20 +63,20 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
   const addr = server.httpServer?.address()
   const appPort = typeof addr === 'object' && addr ? addr.port : server.config.server.port
 
+  // Read the applied host from the Vite dev server's resolved config —
+  // this reflects any user-supplied Vite config that may have overridden
+  // our defaults. `server.host` is `string | boolean | undefined`; non-string
+  // values (true/false/undefined → 0.0.0.0/localhost) aren't useful as a
+  // URL host, so fall back to 'localhost'.
+  const resolvedHost = server.config.server.host
+  const appHost = typeof resolvedHost === 'string' ? resolvedHost : 'localhost'
+
   // Register the studio/app dev server in the registry (federated projects only)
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -141,9 +141,8 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
   }
 
   if (workbenchAvailable) {
-    const workbenchUrl = `http://${httpHost || 'localhost'}:${workbenchPort}`
     output.log(
-      `Workbench dev server started at ${styleText(['blue', 'underline'], workbenchUrl)} (app on port ${appPort})`,
+      `Application available at ${styleText(['blue', 'underline'], `http://${httpHost || 'localhost'}:${workbenchPort}/local/${appHost}-${appPort}`)}`,
```
joshuaellis, line 145: We did this with the original dashboard architecture but the issues i see are:
* we have to keep this the same, forever
* its directly related to the host app but won't take into account and configurations

At minimum this probably needs to come from the `@sanity/workbench` package imo

PR #1028, opened 2026-04-29
fix(workbench): add `__mf__temp` directory to .gitignore

gu-stav:
### Description

Module federation still generates the `__mf__temp` directory in the root of the bootstrapped application when running `sanity init`, so we need to add it to the gitignore file.

I've added it unconditionally, because if a user would enable module-federation later, it is already ignored.

---

> [!NOTE]
> **Low Risk**
> Small, template-only change plus a matching test assertion; no runtime logic changes beyond the generated `.gitignore` contents.
> 
> **Overview**
> Bootstrapped projects from `sanity init` will now ignore the module federation temporary directory by adding `.__mf__temp/` to the shared `.gitignore` template.
> 
> The federation init test is updated to assert the generated `.gitignore` contains this entry, and a changeset is added to ship the CLI patch release.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f9f7345b1e596bf0902276634050da0b36c8d5af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1028.md
+++ b/.changeset/pr-1028.md
@@ -3,4 +3,4 @@
 '@sanity/cli': patch
 ---
 
-add `__mf__temp` directory to .gitignore
+fix(workbench): add `__mf__temp` directory to .gitignore
--- a/.changeset/pr-1028.md
+++ b/.changeset/pr-1028.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+add `__mf__temp` directory to .gitignore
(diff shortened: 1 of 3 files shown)


joshuaellis:
it'd be nice if we could avoid having this at the root of projects. iirc its set from the vite plugin, should we look into adding a way to change this filePath upstream do you think?

gu-stav:
@joshuaellis Turns out it wasn't hard to add it there: https://github.com/module-federation/vite/pull/670

Since it will take a while to get this merged and updated (the HMR implementation has since been done in the Vite plugin) I'm wondering if we should merge this and revert once it has landed?

joshuaellis:
> Since it will take a while to get this merged and updated (the HMR implementation has since been done in the Vite plugin) I'm wondering if we should merge this and revert once it has landed?

Sure, i think we should be very careful about introducing the HMR behaviour in and be extra vigilant it works as intended before shipping,

joshuaellis (approved):

joshuaellis:
ALso need to write a follow ticket to revert this change when we do the update

gu-stav:
@claude add an issue to the Dashboard 2.0 project, to revert the above change.

PR #1079, opened 2026-05-13
feat: configure agent skills as part of `sanity init`

jonahsnider:
### Description

Configure agent skills alongside the MCP server when running `sanity init`. The same logic is reused across both MCP setup and the skills setup for selecting what agents/editors, so that you only get asked once.

Internally this is just shelling out to the `skills` CLI, which is what Neon's `neonctl` CLI does as well.

### What to review

- Are the automated tests written well? ie. not too specific
- Additional questions
    - `--mcp` and `--no-mcp` are a little confusing, since they now also decide whether agent skills are added. Should we deprecate these flags and add new ones? Keep same?
    - Should we literally call `npx skills` or does it make sense to add `skills` as a dependency of the CLI? Should be a lot faster since we are no longer waiting for `npx` to download `skills`, but means we need to stay on top of `skills` updates via Renovate

### Testing

TODO: write a brief guide on how to test locally, including a note about Sanity internal env

### Todo list for myself

- [x] Update `sanity init` to
- [ ] Add `sanity skills` command for adding and updating agent skills in existing projects

---

> [!NOTE]
> **Medium Risk**
> Adds best-effort execution of an external `npx skills add` command during `sanity init`, which can affect init reliability/UX and write files into the new project directory. MCP setup logic is refactored to handle a new `cwd`/skills path and updated prompting/telemetry, increasing behavioral surface area.
>
> **Overview**
> `sanity init` now passes the new project `outputPath` into `setupMCP`, which in turn **installs Sanity agent skills** (via `npx skills add sanity-io/agent-toolkit`) for the selected/detected editors when a `cwd` is provided.
>
> MCP setup is updated to track `skillsCliAgent` mappings per editor, adjust the prompt copy, avoid writing MCP config for already-configured editors while still installing project-local skills during init, and surface skills-install failures separately without aborting MCP setup.
>
> Telemetry/test infrastructure is updated to record `installedSkillsCliAgents`, add unit tests for the new skills installer and MCP/skills interactions, and switch Vitest “agent” detection to `@vercel/detect-agent`.
>
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d3b61c207b6cf6d3807489ee032641b4cb6f872f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,6 +55,7 @@
     "@commitlint/types": "^20.4.0",
     "@eslint/compat": "catalog:",
     "@sanity/eslint-config-cli": "workspace:*",
+    "@vercel/detect-agent": "^1.2.3",
     "@vitest/coverage-istanbul": "catalog:",
     "eslint": "catalog:",
     "husky": "^9.1.7",
--- a/packages/@sanity/cli/src/actions/skills/setupSkills.ts
+++ b/packages/@sanity/cli/src/actions/skills/setupSkills.ts
@@ -20,7 +20,7 @@ export const SANITY_SKILLS_REPO = 'sanity-io/agent-toolkit'
 interface SetupSkillsOptions {
   /**
    * Editors that MCP was (or will be) configured for. Skills will be installed
-   * for each editor's mapped `skillsAgent`.
+   * for each editor's mapped `skillsCliAgent`.
    */
   editors: Editor[]
 
@@ -55,7 +55,7 @@ export async function setupSkills(options: SetupSkillsOptions): Promise<SetupSki
   }
 
   const agents = editors.flatMap((editor) => {
-    const agent = (EDITOR_CONFIGS[editor.name] as {skillsAgent?: string}).skillsAgent
+    const agent = (EDITOR_CONFIGS[editor.name] as {skillsCliAgent?: string}).skillsCliAgent
     return agent ? [agent] : []
   })
 
--- a/packages/@sanity/cli/src/actions/skills/setupSkills.ts
+++ b/packages/@sanity/cli/src/actions/skills/setupSkills.ts
@@ -19,6 +19,13 @@ export const SANITY_SKILLS_REPO = 'sanity-io/agent-toolkit'
 
 interface SetupSkillsOptions {
   /**
+   * Working directory in which to run `npx skills add`. Required so skills are
+   * always written into a concrete project directory rather than wherever the
+   * user happened to invoke the CLI from (e.g. `~/dev`).
+   */
+  cwd: string
+
+  /**
    * Editors that MCP was (or will be) configured for. Skills will be installed
    * for each editor's mapped `skillsCliAgent`.
    */
@@ -47,7 +54,7 @@ interface SetupSkillsResult {
  * best-effort and should never abort MCP setup or `sanity init`.
  */
 export async function setupSkills(options: SetupSkillsOptions): Promise<SetupSkillsResult> {
-  const {editors, mode = 'auto'} = options
+  const {cwd, editors, mode = 'auto'} = options
 
   if (mode === 'skip') {
     skillsDebug('Skipping skills installation (mode: skip)')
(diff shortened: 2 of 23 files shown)


jonahsnider:
* **#1079** <a href="https://app.graphite.com/github/pr/sanity-io/cli/1079?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/sanity-io/cli/1079?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

This stack of pull requests is managed by <a href="https://graphite.dev?utm-source=stack-comment"><b>Graphite</b></a>. Learn more about <a href="https://stacking.dev/?utm_source=stack-comment">stacking</a>.

runeb:
@jonahsnider could you address CI checks and bot review findings first?

package.json
```diff
@@ -55,6 +55,7 @@
     "@commitlint/types": "^20.4.0",
     "@eslint/compat": "catalog:",
     "@sanity/eslint-config-cli": "workspace:*",
+    "@vercel/detect-agent": "^1.2.3",
```
jwoods02, line 58: nit: I can only see this being used in the vitest config, do we need to add a new dep for this? (It's only dev dep so nit)

packages/@sanity/cli/src/actions/skills/setupSkills.ts
```diff
@@ -0,0 +1,97 @@
...
+    ...uniqueAgents.flatMap((agent) => ['-a', agent]),
+    '-y',
+  ]
+
+  skillsDebug('Running: npx %s (cwd: %s)', args.join(' '), cwd)
+
+  try {
+    // The cwd may not exist yet when called during `sanity init` (project
+    // bootstrap happens later). Create it so `npx` doesn't bail with ENOENT.
+    await fs.mkdir(cwd, {recursive: true})
```
jwoods02, line 80: If I understand this correctly, we're installing the skills "project" level not global and doing this work before pulling the template down?

If we create the directory here and add skills, could this conflict / break the template pulling down? I thought you got an error if the directory wasn't empty.

Apologies if tested or misunderstood but worth double checking this

I think we should probably invert this and ad the skills after the project has been pulled down and inited

jwoods02 (commented):
Nice job, I've got a couple of questions around this but mainly the UX:

The proposed flow of sanity init would still show something like this as I understand it:
? Configure Sanity MCP and agent skills for your editor?
  ◉ Cursor
  ◉ Claude Code
  ◯ VS Code (missing credentials)

A few issues:

a) The prompt conflates two things with very different scopes.
MCP writes to global config (~/.cursor/mcp.json, ~/.claude.json). Skills writes to the project dir. The user can't tell from "Configure Sanity MCP and agent skills for your editor?" that one of these is happening to their home directory and the other to their new project. No way to separate them either — it's one checkbox per editor.

b) Already-MCP-configured editors silently re-appear, with no explanation.
The "actionable" predicate now includes editors that already have MCP set up globally, because they still need project-local skills. Those editors show up in the prompt with no label distinguishing them from unconfigured ones — getEditorLabel only handles auth expired / missing credentials, not "MCP done, skills pending". So a user who already configured MCP last week sees Cursor in the list and doesn't know what selecting it will do.

c) No success/preview text explains what happened.
After install you see:
✓ MCP configured for Cursor
✓ Installed Sanity agent skills for cursor
Two lines, lowercase agent name on the second one, no indication that skills landed in ./my-project/ vs MCP landing in ~/.cursor/. Users probably puzzle out the global-vs-local distinction once they look around.

d) --mcp / --no-mcp now controls both, as you've already noted.
Doubles down on the same conflation.

I would split these out to 2 separate questions and make skills more generic, MCP questions stays as is and "Install Sanity skills for your editors" is a single yes or no question, if yes we install so they'll work in all detected editors (it's all project level anyway) not options to install skills for each individual editor.

Also to answer the other question in your PR description about adding npx skills as a dependency:
I think we should add skills as a dependency here. We are directly calling it, so we should pin it, manage supply-chain risk, avoid "cold start" issue of downloading it during init etc It is fairly chunky 433 kB addition to our current 3.87 MB bundle but yaml is already a dep of ours so it dedupes making it a little smaller.

If we need it we should add it as a proper dep though IMO.

Happy to discuss this sync or in Slack.

Also not pulled down and tested but will do when everything is green :)

jonahsnider:
Marking this a draft for now since I don't have time to follow up on the review feedback - @jwoods02 if you could create a followup PR with your implementation that'd be great (feel free to base it on this one if that's helpful)

jwoods02:
Closing in favour of https://github.com/sanity-io/cli/pull/1124

PR #1082, opened 2026-05-14
feat(login): add token login

runeb:
### Description

Adds `sanity login --with-token` for token-based CLI login from standard input. The token is validated through `users/me`; any token accepted by that endpoint is stored in the normal CLI auth config, while invalid or expired tokens fail before storage.

All login flows now use the same auth-token storage path, which clears cached telemetry consent and handles previous-token cleanup. When replacing an existing token, the CLI checks whether the previous token is a Sanity API token and skips `/auth/logout` for API tokens because the backend does not treat them as sessions. Expired previous tokens are also ignored.

Fixes AIGRO-4983

### What to review

Review the `sanity login --with-token < token.txt` flow, token validation, previous-token invalidation behavior, TTY/stdin handling, and telemetry handling in `packages/@sanity/cli/src/actions/auth/login/` and `packages/@sanity/cli/src/commands/login.ts`.

### Testing

- `CI=true mise exec -- pnpm test packages/@sanity/cli/src/commands/__tests__/login.test.ts packages/@sanity/cli/src/actions/auth/login/__tests__/login.vercel.test.ts`
- `mise exec -- pnpm check:types`
- `mise exec -- pnpm check:lint`
- `mise exec -- pnpm check:deps`
- `mise exec -- pnpm --filter @sanity/cli readme`

Latest affected login tests passed with 50 passed and 0 failed.

### DIFF
--- a/.changeset/login-token.md
+++ b/.changeset/login-token.md
@@ -1,0 +1,5 @@
+---
+"@sanity/cli": minor
+---
+
+Add `sanity login --with-token` for token-based CLI login from standard input.
--- a/packages/@sanity/cli/README.md
+++ b/packages/@sanity/cli/README.md
@@ -2489,13 +2489,14 @@ Log in to your Sanity account
 
 ```
 USAGE
-  $ sanity login [--open] [--provider <providerId> | --sso <slug>] [--sso-provider <name> ]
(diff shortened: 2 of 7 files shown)


rexxars (approved):
Looks good!

PR #1128, opened 2026-05-22
chore(ci): scope turbo cache by RUNNER_OS

gu-stav:
### Description

Follow-up to #1127. Scopes turbo cache entries by operating system to prevent cross-OS cache sharing.

#### Context

In #1127 we tracked down a CI flake where two `build.app.test.ts` / `build.studio.test.ts` cases crashed on Linux runners with `ENOENT` while statting symlinks under `packages/@sanity/cli-test/fixtures/basic-app/node_modules/`. The chain was:

1. The Windows build copied `node_modules` into the bundled fixtures dir (because of a path-splitting bug in `copy-fixtures.ts`).
2. `fs.cp` resolved those symlinks to absolute Windows paths.
3. `turbo.json` declares `fixtures/**` as a build output, so the Windows-absolute symlinks got cached.
4. Linux CI hit the same turbo cache (cache key wasn't OS-scoped — confirmed in the failing logs where the replay showed `C:\a\cli\cli\...` paths), and the symlinks became dangling on Linux.

#1127 fixes the root cause (the copy-fixtures filter) and adds defensive handling in `testFixture`. This PR addresses the deeper class of problem: build outputs can legitimately differ by OS (resolved symlinks, line endings, native binaries), so caches shouldn't be shared across operating systems.

#### Change

Add `RUNNER_OS` to `globalEnv` in `turbo.json`. Since cache keys hash declared env vars, this gives Linux/Windows/macOS runners separate cache entries.

This mirrors the existing `NODE_VERSION` entry — we already scope by Node version for the same reason.

Locally `RUNNER_OS` is unset, so the cache key gets the same empty value across machines and local sharing is unchanged.

### What to review

- [turbo.json](turbo.json) — single-line addition to `globalEnv`.

### Testing

- No test impact; this only changes how turbo hashes cache keys.
- After merge, the first CI run on each OS will be a cache miss (expected one-time cost).

---

> [!NOTE]
> **Low Risk**
> Low risk: a single `turbo.json` config change that only affects Turbo cache key hashing across OSes and doesn’t alter build/test logic.
> 
> **Overview**
> Scopes Turbo cache keys by operating system by adding `RUNNER_OS` to `globalEnv` in `turbo.json`, preventing cross-OS cache reuse for build outputs (e.g., cached `fixtures/**`). This will cause one-time cache misses per OS but avoids replaying incompatible artifacts between Linux/Windows/macOS runners.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4304a639690f5595b215be33e7db3e07cdf89a1b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jonahsnider (approved):
nice catch!

PR #1130, opened 2026-05-22
refactor: Move more build/schema logic to cli-build package

snocorp:
### Description

This PR moves build and schema logic into the cli-build package in preparation for allowing runtime-cli to execute build logic during deployment.

### Testing

Updated tests as needed.

---

> [!NOTE]
> **Medium Risk**
> Breaking change for any consumer of `@sanity/cli-build/_internal`; studio build, deploy, and schema flows are touched across many entry points, though behavior is largely a move with dependency injection for env vars.
> 
> **Overview**
> **`@sanity/cli-build`** no longer exposes a single `@sanity/cli-build/_internal` entry. Public internals are split into **`@sanity/cli-build/_internal/build`** (Vite config, runtime HTML, favicons, vendor deps, build telemetry) and **`@sanity/cli-build/_internal/extract`** (schema extraction, validation formatting, pattern matching, extract telemetry). The package picks up the build stack it actually uses (Vite, React, chokidar, etc.) and TSX/JSX build support.
> 
> **`@sanity/cli`** now consumes those subpaths instead of owning duplicate build/schema modules: studio/app build, dev/preview servers, deploy, and schema/manifest/graphql commands import from `cli-build`. Env injection for Vite is passed in via a **`getEnvironmentVariables`** callback on `getViteConfig` (CLI still supplies app vs studio env helpers). Some telemetry types moved next to the logic in `cli-build`. A few CLI-only deps were dropped from `@sanity/cli` where logic moved.
> 
> Tests were updated to mock the new import paths; shared **`createMockHttpServer`** / **`createMockWatcher`** helpers moved from CLI test utils into **`@sanity/cli-test`**. E2e/tsconfig path mapping uses `@sanity/cli-build/*` for the split exports. Changeset marks a **minor** bump for `@sanity/cli-build` due to the export break.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3b8e93c59312d9f005be573943150dce6a2fb5a2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


macdonst (dismissed):
LGTM Dave, it's mostly file moves anyway so it's not as terrible to review as the 67 changed files might indicate.

packages/@sanity/cli-build/package.json
```diff
@@ -27,12 +27,14 @@
...
-      "source": "./src/_exports/_internal.ts",
-      "default": "./dist/_exports/_internal.js"
+    "./_internal/build": {
+      "source": "./src/_exports/_internal/build.ts",
+      "default": "./dist/_exports/_internal/build.js"
+    },
+    "./_internal/extract": {
+      "source": "./src/_exports/_internal/extract.ts",
+      "default": "./dist/_exports/_internal/extract.js"
     },
```
rexxars, line 38: I'm perhaps being overly cautious here, but:

Given `@sanity/cli-build` is a dependency of `@sanity/cli`, and uses `^` (not pinned versions), there _could_ conceivably be a situation where an _old_ version of the CLI uses the _latest_ version of `@sanity/cli-build`. In that case, the removal of `@sanity/cli-build/_internal` would break.

Again, might be a bit overly defensive here - but _potentially_ this should be marked as a breaking to force a new major of the package, or we could use `./internal-extract` or some such and leave the old `_internal` in place

rexxars: …but if going with the latter approach, we'd need to make sure that the shape of `_internal` matches what it was (eg `generateWebManifest` was part of it previously but seems to be removed from the new)

rexxars (dismissed):
Approving, but see comment - _could_ be the safest approach would be to bump to `0.2.0`

packages/@sanity/cli-build/package.json
snocorp: Yes, you're correct. Good call. Added a changeset to do a minor bump. I don't think we need backward compat here because it's pre-v1. At least that was my intention. I'll make it v1 once it's finished and stable.

ryanbethel (approved):
Just reviewing the most recent changes since the other more in depth reviews previously. This looks good.

PR #1134, opened 2026-05-26
chore: only trigger agent workflows via label

mwritter:
### Description

Disable auto agent review workflow files. For now keep workflows and allow opt in triggering via label add for specific agents.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -2,15 +2,15 @@ name: Claude Code Review
 
 on:
   pull_request:
-    types: [opened, synchronize]
+    types: [labeled]
+
+concurrency:
+  group: claude-review-${{ github.event.pull_request.number }}-${{ github.event.label.name }}
+  cancel-in-progress: true
 
 jobs:
   claude-review:
-    if: >-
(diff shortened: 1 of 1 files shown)


.github/workflows/claude-code-review.yml
```diff
@@ -2,15 +2,18 @@ name: Claude Code Review
...
 
 jobs:
   claude-review:
     if: >-
-      !contains(github.event.pull_request.labels.*.name, 'skip-review') &&
+      github.event.label.name == 'claude:review' &&
       github.actor != 'squiggler[bot]' &&
-      github.actor != 'squiggler-app[bot]' &&
-      (github.actor != 'renovate[bot]' || contains(github.event.pull_request.labels.*.name, '⚠️ major'))
+      github.actor != 'squiggler-app[bot]'
```
runeb: Are these needed? They would just ignore this tagging by those agents afaik

runeb (dismissed):

.github/workflows/claude-code-review.yml
mwritter: Yeah since this only runs when the label is added I dont think we need any of the `github.actor` checks (they shouldn't mess with labels)

jwoods02 (approved):

PR #1138, opened 2026-05-27
feat(cli-core): expand isCi detection coverage via ci-info

runeb:
### Description

Replaces the hand-rolled three-env-var CI check with [`ci-info`](https://github.com/watson/ci-info), which detects ~80 CI providers (GitHub Actions, GitLab CI, CircleCI, Vercel, Buildkite, Azure Pipelines, etc.).

Resolves [AIGRO-5015](https://linear.app/sanity/issue/AIGRO-5015/detect-when-cli-is-running-in-ci).

### What to review

- `packages/@sanity/cli-core/src/util/isCi.ts` — now a 3-line pass-through to `ci-info`'s `isCI`.
- `packages/@sanity/cli-core/package.json` — adds `ci-info: ^4.3.1` as a direct dep.

`isCi()` is consumed by the update notifier, telemetry disclosure, telemetry consent resolution, and telemetry status — all already mock `isCi` directly in their tests, so the change is transparent to them.

### Testing

No new tests — `isCi` is now a trivial wrapper around `ci-info`, and testing it would only verify the mock returns what we set. The existing consumers continue to mock `isCi` directly.

Verified: `pnpm check:types`, `pnpm check:lint`, `pnpm check:deps` pass.

---

> [!NOTE]
> **Low Risk**
> Small utility swap; public `isCi()` API unchanged and downstream code already mocks it in tests.
> 
> **Overview**
> **`isCi()`** in `@sanity/cli-core` no longer checks a few env vars (`CI`, `CONTINUOUS_INTEGRATION`, `BUILD_NUMBER`) via a local **`isTrueish`** helper. It now returns **`ci-info`**’s **`isCI`**, which recognizes many more CI environments (e.g. GitHub Actions, Vercel, Buildkite).
> 
> The **`isTrueish`** module is removed. **`ci-info`** `^4.3.1` is added as a direct dependency, with a **minor** changeset entry.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 106e55990d6a9525c59d7a838986e46131edf48b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1138.md
+++ b/.changeset/pr-1138.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-core': minor
+---
+
+feat(cli-core): expand isCi detection coverage via ci-info
--- a/packages/@sanity/cli-core/package.json
+++ b/packages/@sanity/cli-core/package.json
@@ -68,6 +68,7 @@
     "@sanity/client": "catalog:",
     "babel-plugin-react-compiler": "^1.0.0",
     "boxen": "^8.0.1",
+    "ci-info": "^4.3.1",
     "debug": "catalog:",
     "get-it": "^8.7.0",
     "get-tsconfig": "catalog:",
--- a/packages/@sanity/cli-core/src/util/isCi.ts
+++ b/packages/@sanity/cli-core/src/util/isCi.ts
@@ -1,6 +1,3 @@
-import {isTrueish} from './isTrueish.js'
+import {isCI} from 'ci-info'
 
-export const isCi = (): boolean =>
-  isTrueish(process.env.CI) || // Travis CI, CircleCI, Gitlab CI, Appveyor, CodeShip
-  isTrueish(process.env.CONTINUOUS_INTEGRATION) || // Travis CI
-  isTrueish(process.env.BUILD_NUMBER) // Jenkins, TeamCity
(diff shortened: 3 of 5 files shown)


rexxars:
Were there CI providers that were missed that caused this change? We're trying to keep dependency count down both due to install time but also in light of recent wave of npm-related security issues. `eslint-plugin-unicorn` is only used in dev, so this would be a new _production_ dependency.

runeb:
> Were there CI providers that were missed that caused this change? We're trying to keep dependency count down both due to install time but also in light of recent wave of npm-related security issues. `eslint-plugin-unicorn` is only used in dev, so this would be a new _production_ dependency.

@rexxars 

Good catch, the dependency is a net add due to it only coming in via dev dependencies. Addressed. Also addressed the false finding that `isTrueish` is still used, there is a separate copy of that function which confused Claude.

On the addition; there is no report of a CI environment which is not reliably detected today. This PR attempts to strengthen CI env detection as we roll out more of the platform in the cli and also expect it to be exercised more in CI, by adding a widely used dependency (comparable clis ship it).

If you think `CI=true` is sufficient I am happy to close this PR in favor of less dependencies, though!

PR #1143, opened 2026-05-29
feat(workbench): application extension API + sanitized core-app manifest icons

gu-stav:
## Depends on

- sanity-io/workbench#226 — `unstable_defineApp` in `@sanity/federation`, consumed here via the released `0.1.0-alpha.9`

Required by: sanity-io/sanity#12911

## Why

Workbench application extension API on the CLI side:

1. **`unstable_defineApp` is the sole workbench opt-in.** The `federation: { enabled }` config flag is **removed entirely** — gating (workbench dev server, federation registration, the federation Vite build plugin) keys off the `isWorkbenchApp` identity (imported from `@sanity/federation`, not re-declared). The branded `app` bypasses the legacy `app` object schema so the brand survives config parsing.
2. **`applicationType` resolved at config load** — explicit wins, else inferred from a `sanity.config.*` (⇒ studio) else coreApp. `determineIsApp` reads it, so a studio that opts in via `unstable_defineApp` isn't misclassified as an SDK app.
3. **`@sanity/cli` re-exports `unstable_defineApp`** (+ `DefineAppInput`) so `sanity/cli` can surface it to app authors.
4. **`sanity init`** no longer prompts for / scaffolds federation — workbench is opt-in.

`@sanity/federation` is the released `0.1.0-alpha.9` (no preview override).

All CLI tests pass (cli-core config + dev/build suites updated to opt in via a branded `app`).

## Icon sanitization (folded from #1182)

The core-app manifest inlined the `unstable_defineApp` icon SVG verbatim, while the studio manifest sanitized its icon through a DOMPurify allowlist. Aligned: a shared `sanitizeIcon` runs the app icon through the same allowlist, lazy-loaded (`isomorphic-dompurify` pulls in jsdom) so only a deploy with an app icon pays the cost.

---

> [!NOTE]
> **Medium Risk**
> Breaking config removal (`federation.enabled`) and changed workbench detection affect dev/build/deploy paths; migration requires calling `unstable_defineApp` in `sanity.cli.ts`.
> 
> **Overview**
> Replaces the experimental **`federation: { enabled }`** CLI config with **`unstable_defineApp`** in `sanity.cli.ts` as the only workbench opt-in. **`sanity/cli`** re-exports `unstable_defineApp` (and `DefineAppInput`); dev, build, and Vite now gate workbench/federation on **`isWorkbenchApp`** instead of that flag.
> 
> **`@sanity/cli-core`** inlines the workbench brand check (`Symbol.for('sanity.workbench.defineApp')`) so it no longer depends on `@sanity/federation` on every command. Branded `app` values skip the legacy `app` schema; **`parseWorkbenchCliConfig`** validates the rest, clones the app, and sets **`applicationType`** (explicit or inferred from `sanity.config.*` → studio vs coreApp). **`determineIsApp`** uses that so workbench studios are not treated as SDK apps.
> 
> **`sanity init`** drops the federation prompt/flag, `--federation`, and federation scaffolding (including the workbench `sanity` dist-tag override). Core-app manifest icons are run through shared **`sanitizeIcon`** (same DOMPurify allowlist as studio), lazy-loaded on deploy.
> 
> **`@sanity/federation`** bumps to `0.1.0-alpha.9` in cli/cli-build; the federated-studio fixture opts in via `unstable_defineApp`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b5ee546ac9ebea2ef9c7f78a4c6e8cdd63e0665f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.changeset/cli-core-drop-federation.md
joshuaellis, line 1: do we need all three changesets? if so i think this PR is therefore too big.

packages/@sanity/cli-build/src/actions/build/getViteConfig.ts
```diff
@@ -72,6 +72,12 @@ interface ViteOptions extends Pick<CliConfig, 'federation' | 'schemaExtraction'>
 
   isApp?: boolean
 
+  /**
+   * Whether this is a workbench app (opted in via `unstable_defineApp`). Drives
+   * the module-federation build — replaces the old `federation.enabled` flag.
+   */
+  isWorkbench?: boolean
```
joshuaellis: Blocking: Workbench is also its own app/dev-server in the context of the CLI – is there a clearer name we can use internally?

packages/@sanity/cli-core/src/config/cli/workbenchApp.ts
joshuaellis, line 1: Feels really weird that we've branched loading the CLI config to handle workbench-ness, I would advise against it it's a lot to mentally manage in an already complicated cli.

packages/@sanity/cli/src/commands/init.ts
```diff
@@ -78,11 +78,6 @@ export class InitCommand extends SanityCommand<typeof InitCommand> {
         return input
       },
     }),
-    federation: Flags.boolean({
```
joshuaellis, line 81: so whats the replacement to get started?

joshuaellis (changes requested):
Left some inline feedback, my main gripe is the branching cli config parsing – it feels like we've duplicated a lot when we could just make a more powerful schema to manage this perhaps?

.changeset/cli-core-drop-federation.md
gu-stav: No, I can remove them, but we need to cleanup the feature branch anyways at some point.

packages/@sanity/cli/src/commands/init.ts
gu-stav: The new opt-in mechanism is to use `unstable_defineApp` (the identity function). If its being used, an app is opting into federation, otherwise it does not.

packages/@sanity/cli-build/src/actions/build/getViteConfig.ts
gu-stav: Do you have a suggestion?

packages/@sanity/cli-core/src/config/cli/workbenchApp.ts
gu-stav: I don't see how we could do this without a bigger refactoring? Happy to look into this once its in the `main` branch, but I've kept the feature branch intentionally small.

packages/@sanity/cli-build/src/actions/build/getViteConfig.ts
joshuaellis: i'd continue your trend of `isWorkbenchApp` more verbose but makes it clearer when it's passed around everywehre.,

packages/@sanity/cli/src/commands/init.ts
joshuaellis: sure, but should we add a `--workbench` flag instead that automatically adds the `unstable_defineApp` in? because we removed the quick starting feature here.,

gu-stav: Ah yeah, once this goes out for external users we should do that, but for now I'd leave it out of `sanity init` to keep as much of the interface private, so that no humand or agent accidentally enables this yet. Wdyt?

packages/@sanity/cli-core/src/config/cli/workbenchApp.ts
joshuaellis: IDM, we can do another cleanup refactor on the feature branch before we go to main. But i think the only thing blocking it is how you decide the app type which reads the FS at that point in time.

packages/@sanity/cli/src/commands/init.ts
joshuaellis: i would disagree because we want internal users to get it set up – i think the previous flag was very helpful for internal developers to try things out!

gu-stav: Added as follow-up to keep this PR cleaner https://github.com/sanity-io/cli/pull/1236

joshuaellis (approved):

PR #1147, opened 2026-06-01
refactor(cli): pass through reactStrictMode to let studio default take effect

jordanl17:
### Description

Defers the `reactStrictMode` default for `sanity dev` to the `sanity` package instead of having the CLI decide it. `getDevServerConfig` no longer forces a value when `reactStrictMode` is unset, and `getEntryModule` emits `reactStrictMode: undefined` so `renderStudio`'s own destructuring default applies. That default is off in Studio v5 and on in Studio v6, so the on-by-default behavior now lives entirely in `sanity@6` and reaches users only when they upgrade the Studio. Explicit config (`reactStrictMode: false` or `true`) and the `SANITY_STUDIO_REACT_STRICT_MODE` env var still take precedence. Production builds (`sanity build`) are unaffected. Part of [SAPP-3841](https://linear.app/sanity/issue/SAPP-3841).

This reverses the earlier approach on this branch. Previously the CLI flipped the default itself (`cliConfig?.reactStrictMode ?? true`), but because `sanity@5` depends on `@sanity/cli` via `^6.x`, a minor CLI release would have leaked StrictMode-on to existing `sanity@5` users running `sanity dev`. Making the CLI a pass-through keeps this change non-breaking for current `sanity@5` users and confines the breaking default to the `sanity@6` major (sanity-io/sanity#12916), which also covers the component audit.

### What to review

- `getDevServerConfig.ts` - drops the `?? true` fallback; the resolved value is now `boolean | undefined`.
- `getEntryModule.ts` - emits the literal `undefined` instead of coercing with `Boolean()` when `reactStrictMode` is unset, so the generated entry lets `renderStudio`'s default decide.
- `DevServerOptions` / `RuntimeOptions` - `reactStrictMode` widened to `boolean | undefined` (internal types, not part of the public export surface).
- Confirm the opt-out paths are intact: explicit config and the env var still emit concrete booleans, and `sanity build` still passes a literal `false`.

### Testing

Updated `getDevServerConfig` tests: the unset case now resolves to `undefined` (deferring to the studio), with an added explicit `reactStrictMode: true` config case; the env-var-true, env-var-false, and explicit-false cases are unchanged. Added a new `getEntryModule` test asserting the generated entry contains `reactStrictMode: undefined` when unset (for both the config and no-config templates), concrete `true`/`false` when set, and never references `reactStrictMode` for the app entry. All pass; types and format clean.

---

> [!NOTE]
> **Medium Risk**
> Changes default dev behavior when neither config nor env is set (now follows Studio major version), though explicit overrides remain; avoids turning StrictMode on for existing sanity@5 projects via a CLI-only default.
> 
> **Overview**
> For **`sanity dev`**, the CLI stops choosing a default for **`reactStrictMode`** when it is omitted from `sanity.cli.ts` and **`SANITY_STUDIO_REACT_STRICT_MODE`** is not set. **`getDevServerConfig`** now passes through `cliConfig?.reactStrictMode` (including `undefined`) instead of coercing with `Boolean()`, and **`getEntryModule`** writes the literal `reactStrictMode: undefined` in the generated runtime entry instead of `false`, so **`renderStudio`** in the installed **`sanity`** package applies its own default (off on v5, on on v6).
> 
> Explicit **`reactStrictMode: true|false`** in CLI config and the env var still resolve to concrete booleans. Internal types (`DevServerOptions`, `RuntimeOptions`, `getEntryModule`) accept **`boolean | undefined`**. **`CliConfig`** JSDoc now documents default **true** for the option when documented at the config layer. New tests cover dev resolution and generated entry output; a changeset records the **`@sanity/cli`** minor behavior note.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6bb3bedb2af2a669cc4eff544dc4628276b34cda. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/strict-mode-default-dev.md
+++ b/.changeset/strict-mode-default-dev.md
@@ -1,0 +1,5 @@
+---
+'@sanity/cli': minor
+---
+
+Enable React StrictMode by default in `sanity dev`. Opt out with `reactStrictMode: false` in `sanity.cli.ts` or by setting `SANITY_STUDIO_REACT_STRICT_MODE=false`.
--- a/.changeset/strict-mode-default-dev.md
+++ b/.changeset/strict-mode-default-dev.md
@@ -2,4 +2,4 @@
 '@sanity/cli': minor
 ---
 
-Enable React StrictMode by default in `sanity dev`. Opt out with `reactStrictMode: false` in `sanity.cli.ts` or by setting `SANITY_STUDIO_REACT_STRICT_MODE=false`.
+Defer the `reactStrictMode` default for `sanity dev` to the `sanity` package. When `reactStrictMode` is unset in `sanity.cli.ts` and the `SANITY_STUDIO_REACT_STRICT_MODE` env var is absent, the CLI no longer forces a value; `renderStudio`'s own default decides instead. That default is off in Studio v5 and on in Studio v6. Explicit config and the env var still take effect. Opt out with `reactStrictMode: false` in `sanity.cli.ts` or by setting `SANITY_STUDIO_REACT_STRICT_MODE=false`.
--- a/packages/@sanity/cli-build/src/actions/build/__tests__/getEntryModule.test.ts
+++ b/packages/@sanity/cli-build/src/actions/build/__tests__/getEntryModule.test.ts
@@ -1,0 +1,53 @@
+import {describe, expect, test} from 'vitest'
+
+import {getEntryModule} from '../getEntryModule'
+
+describe('getEntryModule', () => {
+  test('emits reactStrictMode: undefined when the option is undefined, deferring to the studio default', () => {
+    const output = getEntryModule({
+      reactStrictMode: undefined,
+      relativeConfigLocation: './sanity.config',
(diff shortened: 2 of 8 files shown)


jordanl17:
Re: rollout coordination - this CLI default flip is intended to land together with the companion Studio V6 work in sanity-io/sanity#12916, which contains the component StrictMode-safety fixes (render-purity, telemetry one-shot, manifest upload cancellation, a subscription-leak fix). Both ship as part of Studio V6.

If this CLI change were to ship before those fixes are in a published `sanity`, the worst case is transient extra dev-only double-invocation (StrictMode is a no-op in production builds), not a runtime regression - the audit found no crash or warning-level incompatibilities, and the dev studios have run with `reactStrictMode: true` for some time. Still, the two are meant to release together.

bjoerge (approved):
Looks great, thanks @jordanl17!

PR #1149, opened 2026-06-02
feat(workbench): build + register app view artifacts, re-synced live in dev (US3 + US5)

gu-stav:
CLI side of workbench US3 (panel views) + US5 (app views). Stacked on #1146.

Threads the app's declared `views` from `sanity.cli.ts` through build and dev into the `@sanity/federation/vite` plugin, so each view's `src` is built into a render artifact (exposed via module federation as `./views/<name>`).

On `sanity deploy`, the views are validated into the application-service payload and logged — the service that stores them doesn't exist yet, so nothing is sent, but a malformed view declaration fails the deploy before the bundle ships. Adds `app.views` to the CLI config type.

**App views (US5, folded in here).** The dev server forwards an app's navigable `entry` as an `app` interface alongside panels/workers, so the workbench can gate navigability on it; with no `entry` the app has no `app` view and isn't reachable as a full-page app. A studio that declares `entry` reaches the not-yet-implemented studio app-view path and is rejected with a clear error. (sanity-io/cli#1215 is closed in favor of this.)

Depends on the `unstable_defineView` / views + `app` interface surface in `@sanity/federation` (sanity-io/workbench#236); consumed via a pkg.pr.new preview (pinned to the development-free build from sanity-io/workbench#240).

## Live interface re-sync + remote rebuild (folded from #1218)

Adding or removing a `views` panel in `sanity.cli.ts` during `sanity dev` now takes effect without a restart. A shared `deriveInterfaces` re-derives the declared interfaces on each config edit; when the interface *set* changes, the app dev server is recreated — its module-federation exposes + codegen artifacts are computed once at startup, so `server.restart()` can't pick up a newly-declared view — and the workbench page reloads to re-fetch the rebuilt remote. A view *source* edit is unchanged: it stays on the HMR path. (FR-024)

---

> [!NOTE]
> **Medium Risk**
> Touches core `sanity dev`/`build` federation and deploy validation paths; mis-sync could break local workbench panels until reload, but behavior is heavily covered by new tests and deploy fails fast on bad view config.
> 
> **Overview**
> Workbench apps can declare **`views`** (panels) and optional **`entry`** (full-page app view) via `unstable_defineApp`. The PR threads **`views`** through **`sanity build` / `sanity dev`** into the federation Vite plugin so each view becomes a federated **`./views/<name>`** artifact, and reads them from the branded app config rather than the legacy `app` object.
> 
> **App views (no `entry`).** When a branded SDK app omits **`entry`**, runtime resolution uses **`null`** so federation skips **`./App`**, dev gets a placeholder page instead of a broken import, and **`deriveInterfaces`** does not emit an **`app`** interface. Studios that declare **`entry`** are rejected until studio app views exist.
> 
> **Dev live sync.** Declared interfaces are derived from config, stored on the dev-server registry (separate from manifest), and re-derived when **`sanity.cli.ts`** changes. If the interface *set* changes (add/remove/rename/repoint), the app dev server is torn down and restarted with a fresh config so federation exposes update; the workbench dev server issues a **full page reload** so clients do not keep a stale remote-entry. Manifest-only edits and view source edits still use soft reconcile / HMR.
> 
> **Deploy.** On **`sanity deploy`**, declared views are validated into an application-service payload (Zod); invalid views fail deploy. The payload is logged only until the backend exists.
> 
> Also adds **`@sanity/cli/runtime`** for browser-safe **`unstable_defineView`**, and pins **`@sanity/federation`** via pkg.pr.new for the workbench API surface.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f247cedd3f95dff244eacdd94e58ba92f84998ea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli-build/src/actions/build/getViteConfig.ts
```diff
@@ -37,7 +37,8 @@ interface ViteOptions extends Pick<CliConfig, 'schemaExtraction'> {
 
   entries: {
     relativeConfigLocation: string | null
-    relativeEntry: string
+    // `null` when a branded app declares no `entry` (US5) — no app view.
```
joshuaellis: need to remove all traces of US5 imho.

packages/@sanity/cli/src/exports/runtime.ts
```diff
@@ -0,0 +1,11 @@
...
+//
+// View/service src files bundle to the browser, so this entry re-exports ONLY
+// from `@sanity/federation` (a pure, dependency-light package) — it must never
+// import `@sanity/cli-core` or anything Node-only, or it would drag the CLI
+// into the frontend bundle. The `sanity` runtime package re-exports from here,
+// so the same constraint protects it.
+//
+// `unstable_defineApp` is config-time (Node) and stays on the main `@sanity/cli`
+// entry; only the runtime helpers live here.
+export {unstable_defineView} from '@sanity/federation'
```
joshuaellis, line 11: discussion: maybe this should have been exported from workbench instead. A view belongs to workbench not module federation no?

packages/@sanity/cli/src/actions/dev/devAction.ts
```diff
@@ -60,12 +69,30 @@ export async function devAction(options: DevActionOptions): Promise<{close: () =
...
 
+  // Adding/removing a view or service in `sanity.cli.ts` during dev requires
+  // rebuilding the federation remote: its module-federation `exposes` map +
+  // codegen artifacts are computed once at server start, so a newly-declared
+  // interface has no expose until the server is recreated. `server.restart()`
+  // can't do it — it re-uses the inline config — so we tear the app server down
+  // and bring it back up on the same port with a freshly-loaded config. The
+  // workbench page then reloads (driven by the registry watch in the workbench
+  // server) to re-fetch the rebuilt remote. A view/service *source* edit doesn't
+  // change the interface set, so it stays on the HMR path untouched.
```
joshuaellis, line 80: help me understand this a bit more, so declaring a new view would change the `expose` value of the federation config right? but that should just automatically trigger a server restart because the user changed the vite config?

packages/@sanity/cli/src/exports/runtime.ts
gu-stav: We could, but the downside of this would be that the extension API would then be split across packages and both packages would depend on each other (for types e.g.).

I think I'd leave it for now and discuss if we want to inline the whole API into the CLI and re-export from `sanity`, to get rid of one layer of re-exports.

joshuaellis: @claude write a follow up ticket for this that asks us to think about our package hierarchy for our public and internal APIS as they're spread across the following packages:

* `@sanity/federation`
* `@sanity/workbench`
* `@sanity/cli`
* `sanity`
* _maybe_ `@sanity/sdk`

joshuaellis: @claude I meant in linear for the SDK team.

packages/@sanity/cli/src/actions/dev/devAction.ts
gu-stav: Not quite. `sanity.cli.ts` isn't a Vite config file, so editing it doesn't trigger Vite's config-change restart. The federation exposes + the codegen artifacts are computed by the CLI from sanity.cli.ts and handed to the federation plugin when we create the server; they're baked in for the server's lifetime, not read from a file Vite watches.

And Vite's own restart wouldn't help anyway: server.restart() reuses the already-resolved inline config, so it doesn't re-load sanity.cli.ts or recompute the exposes.

So we watch sanity.cli.ts ourselve and when the interface set changes, tear the app server down and bring it back up on the same port with a freshly-loaded config (getCliConfigUncached) — that's what recomputes the exposes + regenerates the artifacts. The workbench page then reloads to re-fetch the rebuilt remote.

Does that make sense?

joshuaellis: yes it does. thanks.

joshuaellis (approved):

PR #1182, opened 2026-06-04
fix(workbench): sanitize unstable_defineApp icons via the shared manifest allowlist

gu-stav:
Follow-up to #1149.

`unstable_defineApp`'s `icon` is inlined into the core-app manifest verbatim, while the studio manifest runs its icon through a DOMPurify allowlist. This aligns them so every inlined icon passes the same trusted subset of SVG markup, with the existing studio manifest resolver as the source of truth.

- Extract the studio resolver's sanitization into a shared `sanitizeIcon` (the existing `purifyConfig` allowlist) and run the app icon (`readIconFromPath`) through it too.
- Lazy-load `sanitizeIcon` via `doImport`: `isomorphic-dompurify` drags in jsdom, and the core-app manifest runs in the CLI's main process — unlike the studio manifest, which is worker-isolated and already lazy-loads its icon machinery. An eager import would pull jsdom into the deploy command's startup graph; lazy-loading keeps it out, so only a deploy with an app icon pays the cost.

Builds on an in-progress change that started the `sanitizeIcon` extraction; completes it by aligning the load path with the studio resolver.

### DIFF
--- a/.changeset/pr-1146.md
+++ b/.changeset/pr-1146.md
@@ -4,4 +4,4 @@
 '@sanity/cli': minor
 ---
 
-feat(cli): merge app group + priority into the core-app manifest
+feat(workbench): merge app group + priority into the core-app manifest
--- a/.changeset/pr-1146.md
+++ b/.changeset/pr-1146.md
@@ -1,0 +1,7 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-core': minor
+'@sanity/cli': minor
+---
+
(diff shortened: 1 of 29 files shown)


gu-stav:
Folded into #1143 — the icon sanitization now lives at the lowest level, where extractCoreAppManifest first inlines the app icon (sanitizeIcon + the lazy-loaded sanitize call). Closing in favour of the foundational PR.

PR #1211, opened 2026-06-04
fix: Allow MCP and skills auto-install in all interactive environments, not just production

mwritter:
### Description

Previously, `sanity init` gated MCP server setup and agent skills auto-install behind `getSanityEnv() === 'production'`, which meant developers running in staging (or any non-production internal environment) never got the MCP/skills prompts. This PR removes that environment check so the only gates are `--no-mcp` / `--no-skills` flags and `resolveIsInteractive()` — the same rules production already uses.

### What to review

- **Environment gate removal in `init.ts`**: The `getSanityEnv() !== 'production'` check is removed from both the `mcpMode` and `skillsMode` resolution blocks. The remaining gates (`--no-mcp`, `--no-skills`, non-interactive detection, `--yes` for auto mode) are unchanged. Verify these are sufficient to keep CI/e2e tests safe from unintended auto-installs.
- **Prompt copy tweak in `setupMCP.ts`**: "agent skills" changed to "agent skill" (singular) in the combined MCP + skills prompt message.
- **Test updates**: Both `init.command.test.ts` and `init.staging-env.test.ts` now assert that staging environments receive `skillsMode: 'auto'` and `mode: 'prompt'` instead of `'skip'` for both.

### Testing

Existing unit tests updated to reflect the new behavior — staging environment tests now expect `skillsMode: 'auto'` and MCP `mode: 'prompt'` instead of both being skipped. See `init.command.test.ts` and `init.staging-env.test.ts`.

### DIFF
--- a/.changeset/pr-1211.md
+++ b/.changeset/pr-1211.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix: auto install in staging
--- a/.changeset/pr-1211.md
+++ b/.changeset/pr-1211.md
@@ -3,4 +3,4 @@
 '@sanity/cli': patch
 ---
 
-fix: auto install in staging
+Allow MCP and skills auto-install in all interactive environments, not just production
--- a/.changeset/pr-1211.md
+++ b/.changeset/pr-1211.md
@@ -3,4 +3,4 @@
 '@sanity/cli': patch
 ---
 
-fix: auto install in staging
+fix: Allow MCP and skills auto-install in all interactive environments, not just production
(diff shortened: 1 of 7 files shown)


jonahsnider (dismissed):

.changeset/pr-1211.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix: auto install in staging
```
jonahsnider: this changeset should be updated since it'll be user facing

jwoods02 (dismissed):

jwoods02 (approved):

PR #1215, opened 2026-06-05
feat(workbench): app-view interface from entry (US5)

gu-stav:
Stacked on #1176. The CLI side of US5 (app views) — the dev path.

An SDK app declares its navigable full-page view via `entry`. `sanity dev` now forwards it as an `app` interface on the dev-server registry (alongside panels and workers), so the workbench can gate navigability on it. An app with no `entry` forwards no `app` interface (not navigable). A studio that declares `entry` reaches the not-yet-implemented studio app-view path and is rejected with a clear error.

Pins the `@sanity/federation` preview carrying the `app` interface model.

Scope / follow-ups:

- The deploy payload (the Phase-1 stub — validated + logged, panel-only schema) does not yet include the app view; that follows when the app-service endpoint lands.
- Studio-extraction CI stays red until the federation `development` export is resolved. That is pre-existing and orthogonal: the SDK-app config path loads via jiti → dist and is unaffected; studios load via vite-node, which trips on federation source. US5 defers studios anyway.

Workbench side: sanity-io/workbench#239.

### DIFF
--- a/.changeset/pr-1176.md
+++ b/.changeset/pr-1176.md
@@ -1,0 +1,7 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-build': minor
+'@sanity/cli': minor
+---
+
+feat(workbench): thread workbench services through
--- a/.changeset/pr-1215.md
+++ b/.changeset/pr-1215.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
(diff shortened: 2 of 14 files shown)


gu-stav:
Folded into #1149 (US3): the dev server's app-interface-from-`entry` forwarding + the studio not-yet-implemented rejection now live there. Closing in favor of that restructure.

PR #1011, opened 2026-04-24
refactor(create-sanity): bundle init logic into single file

mttdnt:
### Description

Reapply of #759 onto current `main`. The original PR targeted `bundled-create` and could not be cleanly rebased due to the init refactor in #999 landing on main in the meantime.

The goal is to make `npm create sanity@latest` (and equivalent `yarn`/`pnpm`/`bun` commands) as fast as possible by minimizing the `create-sanity` bundle size. Previously `create-sanity` depended on `@sanity/cli` at runtime, pulling in oclif and hundreds of transitive dependencies just to run `sanity init`. This PR bundles only the init logic into a single Rollup-built file, tree-shaking away everything else.

Key changes to achieve this:
- **Decouple from oclif**: Extract shared flag definitions (`flags.ts`, `flagsToInitOptions.ts`) as plain objects so both oclif and `create-sanity` use the same schema without `create-sanity` needing oclif. `flagAdapter.ts` converts them back to oclif format for the CLI.
- **Replace oclif utilities**: Add lightweight `CLIError`/`CLIWarning` to `@sanity/cli-core/errors` and `stdout`/`warn` to `@sanity/cli-core/ux`, replacing direct imports from `@oclif/core` in init-reachable code paths.
- **Decouple dataset import**: Replace `ImportDatasetCommand` dynamic import with `execa` spawn, removing another oclif dependency from the init flow.
- **Bundle with Rollup**: New `rollup.config.ts` produces a single self-contained file for `create-sanity` with its own arg parser and help formatter.
- **Remove dead code**: Drop `--from-create` flag and the `fromCreate` option (no longer needed since `create-sanity` now runs init directly).

### What to review

- `packages/create-sanity/` — new bundled entry point, rollup config, parseArgs, help formatter, tests
- `packages/@sanity/cli-core/` — new error classes, UX utilities, export changes
- `packages/@sanity/cli/src/commands/init.ts` — slimmed down to use shared flag defs
- `packages/@sanity/cli/src/actions/init/` — extracted `flagsToInitOptions`, execa-based dataset import

### Testing

- TypeScript type checking passes across all packages
- All changed-file tests pass (`pnpm test --changed`)
- One pre-existing unrelated failure in `deploy.app.test.ts` (auth mock issue)

---

> [!NOTE]
> **Medium Risk**
[...]

PR #1237, opened 2026-06-09
fix: don't treat ignored build scripts as failure during dep install

rexxars:
### Description

SDK-1654

When running `npm create sanity`/`sanity init`, if selecting pnpm to use for installing dependencies you can get into this situation where pnpm installs dependencies, but exits with a non-zero exit code, and a message like:

```
[ERR_PNPM_IGNORED_BUILDS] Ignored build scripts: esbuild@0.28.0

Run "pnpm approve-builds" to pick which dependencies should be allowed to run scripts.
```

We need to allow this to happen - treating it as if this was a success (which it kind of is), instead of failing the onboarding process. Since esbuild works fine for the vast majority of cases without running the postinstall script, I decided to not show the "Run 'pnpm approve-builds'" text if that is the only dependency that needs it - but open to changing/removing that, just figured it would be less noise.

> [!NOTE]
> `pnpm create sanity` has a similar problem, but during the `pnpx` stage. This is harder to fix, but might be solved if/when https://github.com/sanity-io/cli/pull/1011 lands

### What to review

Does the change seem sensible?

### Testing

New tests added.

---

> [!NOTE]
> **Low Risk**
> Scoped to CLI package install error handling for pnpm; real install failures still exit with an error, with broader test coverage.
> 
> **Overview**
> Fixes **pnpm** dependency installs during Sanity CLI onboarding (`sanity init` / `npm create sanity`) when pnpm exits non-zero after **`ERR_PNPM_IGNORED_BUILDS`** even though packages were installed.
> 
> **`installPackages`** now runs package-manager commands with **`reject: false`**, detects ignored-build-script output (including wrapped/whitespace-separated lists), and treats that case as **success**. If only **esbuild** was skipped, no extra message is shown; for other packages (e.g. **sharp**, scoped names), it **warns** to run **`pnpm approve-builds`**. Real failures still fail the spinner but now log **stdout and stderr** so errors on stderr are visible. Behavior is unchanged for npm/yarn/bun.
> 
> Tests cover the new pnpm paths plus updated **`execa`** expectations; changeset is a **patch** for **`@sanity/cli`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3452a2f9763c48aa752a3f16770ed0d0afc7ee68. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1237.md
+++ b/.changeset/pr-1237.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix: don't treat ignored build scripts as failure during dep install
--- a/packages/@sanity/cli/src/util/packageManager/__tests__/installPackages.test.ts
+++ b/packages/@sanity/cli/src/util/packageManager/__tests__/installPackages.test.ts
@@ -67,6 +67,7 @@ describe('installDeclaredPackages', () => {
       cwd: workDir,
       encoding: 'utf8',
       env: {PATH: '/mock/path'},
+      reject: false,
       stdio: 'pipe',
     })
     expect(mockSpinnerInstance.start).toHaveBeenCalled()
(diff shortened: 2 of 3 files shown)


binoy14 (approved):
LGTM

PR #1243, opened 2026-06-10
feat(workbench): deploy federated builds

gu-stav:
`sanity deploy` died on every federated (`unstable_defineApp`) app with a generic "Error checking directory". Federated builds emit a module federation remote, but `checkDir` still required `index.html` — and the real error only showed under `DEBUG`.

## What this does

- New `checkWorkbenchAppDir` checks for a federation build: `mf-manifest.json` must exist as the marker that `sanity build` produced one. `checkDir` keeps its `index.html` contract untouched; the deploy paths pick the check based on the app shape.
- A federated app must export at least one interface. The rule reads what `unstable_defineApp` declares (`entry`, views, services) and fails before any prompts or API calls. Workbench studios always expose the studio config, so the rule only applies to the app path.
- Directory check failures surface verbatim instead of the generic message.

Verified live against staging with a panel-only workbench app.

---

> [!NOTE]
> **Medium Risk**
> Changes deploy preflight for hosted bundles (app and federated studio paths); behavior for non-workbench deploys is mostly limited to clearer directory-check error messages.
> 
> **Overview**
> **`sanity deploy` now supports federated (`unstable_defineApp`) builds** instead of failing directory validation that expected a static SPA (`index.html`).
> 
> For workbench apps, deploy runs **`checkWorkbenchApp`** up front so the config must declare at least one **`entry`**, **view**, or **service** (usage error before prompts/API). After build, **`checkWorkbenchAppDir`** validates the output via **`mf-manifest.json`**; classic apps/studios still use **`checkDir`**. App and studio deploy paths pick the check with **`isWorkbenchApp`**.
> 
> Directory validation errors are shown **verbatim** via **`getErrorMessage`** rather than a generic “Error checking directory”. Unit and deploy command tests cover federation dirs, empty-interface rejection, and the updated error text.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0fe7434952aed05d0131277dbc094305880c1895. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/actions/deploy/workbenchChecks.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @internal
+ * @param app - The branded `unstable_defineApp` result from the CLI config
+ * @returns void
+ */
+export function checkWorkbenchApp(app: WorkbenchAppInterfaces): void {
+  const {entry, services, views} = app
+  if (!entry && !views?.length && !services?.length) {
+    throw new Error(
+      'Nothing to deploy: `unstable_defineApp` declares no views or services. ' +
+        'Declare at least one view or service in the app config.',
```
joshuaellis: shouldn't this include that theres no entry either?

joshuaellis (approved):

packages/@sanity/cli/src/actions/deploy/workbenchChecks.ts
gu-stav: Good catch — it does check entry, the message just didn't say so. Updated: 0fe74349

PR #1244, opened 2026-06-10
fix(cli-core): release worker handles so studio commands exit cleanly

binoy14:
### Description

Worker threads spawned via `promisifyWorker` were terminated with a synchronous `setImmediate(() => worker.terminate())`. `terminate()` creates an internal ref'd `MessagePort` that keeps the parent process alive until the worker has fully shut down. With Vite 8+, Rolldown's native bindings can keep the worker thread's event loop alive indefinitely, so the CLI would hang after a command finished instead of exiting.

This changes the cleanup to:

- `unref()` the worker immediately so the parent process can exit without waiting for the worker to shut down.
- Schedule a deferred, `unref()`'d `terminate()` (5s) as a safety net to force-kill workers that never exit on their own (e.g. native addons holding handles), without keeping the process alive itself.
- Close the Vite server in `studioWorkerLoader.worker.ts`'s `finally` block to release handles — especially important under Vite 8+ where Rolldown's native bindings hold the event loop open.

### What to review

- `promisifyWorker.ts`: the new `cleanup()` logic — confirm the unref-first + deferred-terminate approach is sound and the timer is correctly unref'd.
- `studioWorkerLoader.worker.ts`: the `try/finally` around `executeId` and `server.close()`.

### Testing

Updated `promisifyWorker.test.ts` to assert the new behavior: `unref()` is called (instead of synchronous `terminate()`) after a message/error/messageerror, listeners are removed, and the deferred `terminate()` fires after 5s without double-firing alongside the timeout handler.

### DIFF
--- a/.changeset/pr-1244.md
+++ b/.changeset/pr-1244.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-core': patch
+---
+
+fix(cli-core): release worker handles so studio commands exit cleanly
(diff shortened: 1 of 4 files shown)


binoy14:

- #1244 ◀

PR #1276, opened 2026-06-11
feat(cli-build): preconnect and modulepreload the CDN sanity module for auto-update studios

jordanl17:
### Description

For auto-updating studios the main `sanity` module loads cross-origin from the CDN, and the browser only discovers it needs that module after downloading and executing the small entry script - a serial waterfall on every load.

This PR adds two resource hints inside the existing runtime import-map injector (`TIMESTAMPED_IMPORTMAP_INJECTOR_SCRIPT` in `addTimestampImportMapScriptToHtml.ts`): a `preconnect` to the CDN module origin and a `modulepreload` for the `sanity` module, so the connection is warmed and the download starts during head parse instead of after entry execution. Both hints set `crossorigin="anonymous"` to match the module's CORS-anonymous fetch, and the `modulepreload` href reuses the runtime `replaceTimestamp(...)` value so it matches the import-map entry and the largest chunk is not fetched twice.

Scoped to auto-update studios by construction: the injector runs only when an import map is present and hints emit only when an import resolves to a sanity-cdn host, so non-auto-update studios are unchanged. This is the one quick win from a wider Studio auth-ready p95 investigation; the expected saving is modest (roughly 0.25-0.8s per auto-update load) because the dominant pre-mount cost is CPU parse/execute, not download.

### What to review

- The only production change is the inline injector string in `addTimestampImportMapScriptToHtml.ts`. Confirm the no-importMap early return and the CSS-link loop are unchanged.
- Both hints are gated on the shared `isSanityCdnUrl` host check, so the bridge host `core.sanity-cdn.com` is excluded and only the module host is targeted (`sanity-cdn.com`, or `sanity-cdn.work` on staging).
- The `modulepreload` href uses the same timestamp as the import map, and both hints set `crossorigin="anonymous"`, so the warmed connection is reused and the module is not fetched twice.
- The `preconnect` has marginal value: the module URL 302-redirects from `sanity-cdn.com` to `modules.sanity-cdn.com`, where the chunks actually download, so the `preconnect` warms the redirector host rather than the asset host. Worth deciding whether to keep it.

### Testing

Four JSDOM runtime tests assert against the executed DOM: `preconnect` emitted with `crossorigin` to the module origin; `modulepreload` href carries the fresh timestamp and equals the import-map's resolved `sanity` entry (the no-double-fetch guarantee); no hints when no import resolves to a sanity-cdn host; and staging-host support. All tests in the file pass, and a mutation check confirmed the new tests fail when the hint code is removed.

Verified on a real build: `sanity build --auto-updates` on the `basic-studio` fixture, served and loaded in Chromium against the live CDN. The injector runs with no errors of its own, the `modulepreload` href matches the import-map `sanity` entry, and the `sanity` module entry URL is requested exactly once (the `modulepreload` and the entry-script import coalesce - no double fetch). The remaining deployed-preview DevTools check is now confirmatory rather than load-bearing. Build, type-check, lint, and format pass.

### DIFF
--- a/.changeset/pr-1276.md
+++ b/.changeset/pr-1276.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-build': minor
+---
+
+feat(cli-build): preconnect and modulepreload the CDN sanity module for auto-update studios
--- a/packages/@sanity/cli-build/src/actions/build/renderDocumentWorker/__tests__/addTimestampImportMapScriptToHtml.test.ts
+++ b/packages/@sanity/cli-build/src/actions/build/renderDocumentWorker/__tests__/addTimestampImportMapScriptToHtml.test.ts
@@ -215,6 +215,7 @@ describe('tests the runtime TIMESTAMPED_IMPORTMAP_INJECTOR_SCRIPT', () => {
     expect(preloadHref).toBe(
       'https://sanity-cdn.com/v1/modules/sanity/default/%5E3.2.0/t1700000000',
     )
+    expect(preloadLinks[0].getAttribute('crossorigin')).toBe('anonymous')
 
     const importMapEl = dom.window.document.querySelector('script[type="importmap"]')
     const runtimeImportMap = JSON.parse(importMapEl?.textContent || '{}') as {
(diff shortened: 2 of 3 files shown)


bjoerge (approved):
Nice!

PR #1277, opened 2026-06-12
test(cli): Convert more studio build tests to unit tests

snocorp:
### Description

This PR converts a few more tests into unit tests. These tests just verify the interactive and unattended mode logic and do not need to build a studio.

---

> [!NOTE]
> **Low Risk**
> Test-only refactor with no production code changes; behavior is preserved by equivalent unit coverage.
> 
> **Overview**
> Moves three **studio build** interaction scenarios out of the integration `build.studio.test.ts` suite into **`buildStudio.unit.test.ts`**, where `buildStudio` is exercised directly with heavier mocks (UX prompts, dependency comparison, upgrades, interactivity).
> 
> The relocated cases cover **skipping the version-mismatch prompt after disabling auto-updates for a prerelease**, and **skipping prompts with a warning** in **`--yes`** and **non-interactive** modes. Default test flags in the unit file now use **`yes: false`** so unattended behavior is asserted explicitly where needed.
> 
> The integration suite keeps full **`BuildCommand`** / fixture builds; only the redundant slow paths for this logic are removed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b3209abefc5b98899f4e6728fa0620871e4b92ab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
+++ b/packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
@@ -38,16 +38,20 @@ vi.mock('../getEnvironmentVariables.js', () => ({
   getStudioEnvironmentVariables: mockGetStudioEnvironmentVariables,
 }))
 
-vi.mock('@sanity/cli-build/_internal/build', () => ({
-  buildDebug: vi.fn(),
-  checkStudioDependencyVersions: vi.fn().mockResolvedValue(undefined),
-  resolveVendorBuildConfig: vi.fn().mockResolvedValue({
-    entries: {},
-    namesByChunkName: {},
-    specifiersByChunkName: {},
-  }),
-  StudioBuildTrace: {},
-}))
+vi.mock('@sanity/cli-build/_internal/build', async (importOriginal) => {
+  const original = await importOriginal<typeof import('@sanity/cli-core/ux')>()
+  return {
+    ...original,
+    buildDebug: vi.fn(),
+    checkStudioDependencyVersions: vi.fn().mockResolvedValue(undefined),
+    resolveVendorBuildConfig: vi.fn().mockResolvedValue({
+      entries: {},
+      namesByChunkName: {},
+      specifiersByChunkName: {},
+    }),
+    StudioBuildTrace: {},
+  }
+})
 
 vi.mock('@sanity/cli-core', async (importOriginal) => {
   const original = await importOriginal<typeof import('@sanity/cli-core')>()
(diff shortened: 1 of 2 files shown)


packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
```diff
@@ -1,16 +1,31 @@
 import {Output} from '@sanity/cli-core'
-import {describe, expect, test, vi} from 'vitest'
+import {spinner} from '@sanity/cli-core/ux'
```
filmaj: Is this import needed? LSP in my IDE says it's not used.

packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
```diff
@@ -65,4 +111,78 @@ describe('#buildStudio', () => {
...
+    // First select call is for prerelease prompt
+    mockedSelect.mockResolvedValue('disable-auto-updates')
+
+    await buildStudio({
+      autoUpdatesEnabled: true,
+      cliConfig: {},
+      flags: FLAGS,
+      outDir: '/tmp/dist',
+      output,
+      workDir: '/tmp',
```
macdonst, line 134: Will this work on Windows systems? Do we need to use something like:

```javascript
import {tmpdir} from 'os'
```

to make sure it works cross platform.

packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
```diff
@@ -65,4 +111,78 @@ describe('#buildStudio', () => {
...
+    })
+
+    await buildStudio({
+      autoUpdatesEnabled: true,
+      cliConfig: {},
+      flags: {...FLAGS, yes: true},
+      outDir: '/tmp/dist',
+      output,
+      workDir: '/tmp',
+    })
```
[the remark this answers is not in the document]
filmaj: This seems accurate?

filmaj (changes requested):
Please rebase 🙏 left a couple of comments. I think the robot is right, too.

packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
snocorp: It never actually writes or reads files so it should be fine. Would it be better to make that clear with comments?

packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
snocorp: Good catch, removing

packages/@sanity/cli/src/actions/build/__tests__/buildStudio.unit.test.ts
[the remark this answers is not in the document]
snocorp: Also fixed

filmaj (approved):
Very nice! LFG!

PR #1290, opened 2026-06-12
fix(workbench): exit the dev process after signal-triggered teardown

gu-stav:
Closing `sanity dev` often left the process running and holding its ports: the workbench signal handler traps SIGINT/SIGTERM (which disables Node's default exit), tears down, but never exits, so any handle that survives teardown — a keep-alive socket, an extraction worker mid-run — keeps the event loop alive. It now re-raises the signal once teardown settles (with an unref'd grace timer as a backstop), and along the way makes `close()` single-flight and refuses a rebuild requested during shutdown.

---

> [!NOTE]
> **Medium Risk**
> Touches dev-server lifecycle and process signal handling; mis-teardown could still leak ports or double-close, but scope is limited to workbench/registration dev paths with new unit coverage.
> 
> **Overview**
> Fixes **`sanity dev`** often staying alive after Ctrl+C when the workbench registers **SIGINT/SIGTERM** handlers: trapping the signal disabled Node’s default exit, so teardown could finish while the process still held ports.
> 
> **Signal shutdown** now runs **`close()`**, then **re-raises the same signal** so the process exits with normal semantics. A **5s unref’d grace timer** force re-raises if teardown never settles. **`close()`** is **single-flight** (duplicate signals or parallel **`close()`** calls share one teardown), **awaits an in-flight interface-set rebuild** so the replacement Vite server is closed, and **rejects rebuilds** once shutdown has started.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3347c385ad4a2143374c189cdbb2eb31a650542a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1290.md
+++ b/.changeset/pr-1290.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix(workbench): exit the dev process after signal-triggered teardown
--- a/packages/@sanity/cli/src/actions/dev/__tests__/devAction.unit.test.ts
+++ b/packages/@sanity/cli/src/actions/dev/__tests__/devAction.unit.test.ts
@@ -58,6 +58,11 @@ function passedInterfaceSetChange(): (() => Promise<unknown>) | undefined {
   return mockStartDevServerRegistration.mock.calls[0][0].onInterfaceSetChange
 }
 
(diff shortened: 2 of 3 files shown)


gu-stav:
@claude review

gu-stav:
@claude re-review

PR #1293, opened 2026-06-12
fix(workbench): bind a user-pinned port strictly instead of drifting

gu-stav:
### Description

Setting `server: {port: 5555}` in `sanity.cli.ts` doesn't pin the workbench dev server to that port — `createWorkbenchViteServer` hardcodes `strictPort: false`, so a busy port makes Vite silently drift to the next free one. A pinned port is a contract with external tooling (port allocators, proxies pointing at the URL), and the drift compounds: the app server stacks on `workbenchPort + 1`, so everything shifts off the allocated ports without a word.

`getSharedServerConfig` now reports whether the port was pinned (flag, env var, or `server.port` — the same chain that resolves the value), and the workbench server binds strictly when it was. A busy pinned port takes the existing fallback path: warn and run without a workbench rather than on the wrong port. The unconfigured 3333 default keeps the drift, same as before.

The app/studio dev server keeps its own non-strict behavior (`getViteConfig` decides that separately) — this only stops the workbench from squatting a neighboring port when the user asked for a specific one.

### What to review

The `httpPortConfigured` derivation in `getSharedServerConfig` (pinned means the flag/env/config chain produced a value — an explicit 3333 counts), and the `strictPort` plumbing through `devAction` into `createWorkbenchViteServer`.

### Testing

`getSharedServerConfig.test.ts` covers all 4 port sources, `startWorkbenchDevServer.test.ts` covers strict/non-strict vite config plus the busy-pinned-port fallback (lock released, warn, no workbench), and `devAction.test.ts` covers the plumbing both ways.

---

> [!NOTE]
> **High Risk**
> Large change to dev server orchestration, Vite config branching, and module-federation build/deploy; misconfigured ports or federation exposes could break local dev and deploys for workbench projects.
> 
> **Overview**
> Adds **Workbench** for projects that opt in via `unstable_defineApp` in `sanity.cli.ts`: the CLI loads a branded app config (`@sanity/cli-core`), builds federated remotes in `@sanity/cli-build` (`@module-federation/vite`, generated view/service artifacts, `./_internal/federation` APIs), and **`sanity dev`** runs a workbench host plus the app/studio Vite server on the next port with registry registration and **rebuild when views/services are added or removed** (`getCliConfigUncached` + `onInterfaceSetChange`).
> 
> **Build/deploy** paths gain `isWorkbenchApp`: federation skips the classic `.sanity/runtime` client bundle; workbench studios **error without `sanity.config`**; deploy validates `mf-manifest.json` and that the app exposes at least one interface.
> 
> **Port pinning:** `getSharedServerConfig` now exposes `httpPortConfigured` (flag, env, or `server.port`); `devAction` passes `strictPort` into the workbench server so a user-pinned port does not silently drift (which would also move the stacked app port).
> 
> Also includes a **federated-studio** fixture, shared `getSanityConfigDir` / `getSanityDataDir`, and keeps federated remote **dts generation off** to avoid tsconfig breakage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 77d06a740a1329ffec39fc37e3a10ae0a797fede. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joshuaellis:
what is the actual problem here, the description is quite waffle-y and im not getting it – is it that you set port X but if it isn't available it disregards it? if so why dont we just bail like the studio does if we feel that strongly about it adhering to the port the user has chosen?

joshuaellis (dismissed):

PR #1295, opened 2026-06-12
test(login): wait for callback server before simulating oauth callback

gu-stav:
### Description

`login.test.ts` fails regularly on CI, also on main ([example run](https://github.com/sanity-io/cli/actions/runs/27423213442/job/81054215544)). The root cause is a race: the OAuth callback helper slept a fixed 100ms and then fired at a hardcoded `localhost:4321`. On slow runners the login command hasn't bound its callback server within that window, so the callback gets ECONNREFUSED and the test fails. The abandoned login command then keeps running in the background — it squats on port 4321 and consumes nock mocks meant for later tests, which is where the cascading `pending mocks` and `Nock: No match` failures come from.

`startServerForTokenCallback` is now wrapped with a pass-through spy, so tests await the real server startup and read the port it actually bound (the fallback ports too, no more guessing 4321 vs 4000). `afterEach` closes any server still listening, so a single failure can't cascade through the rest of the file.

Side effects:

- The file runs in ~1.5s instead of ~17s, since the unconditional sleeps are gone.
- The port-fallback test asserts against the port the server really bound instead of assuming 4000.

### What to review

`waitForCallbackUrl()` and the `afterEach` server cleanup in `login.test.ts` — the rest is call sites losing their port argument.

### Testing

Test-only change. 48/48 pass, stable across repeated runs, including together with `login.vercel.test.ts`.

### DIFF
--- a/packages/@sanity/cli/src/commands/__tests__/login.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/login.test.ts
@@ -96,23 +96,38 @@ function mockStdin(input: string, options: {isTTY?: boolean} = {}) {
  * Waits until the login command's local callback server is listening, and
  * returns the callback URL it actually bound. The port can differ from 4321
  * when the server falls back to another port.
+ *
+ * Rejects with the command's error if it exits before the server starts.
  */
-async function waitForCallbackUrl(): Promise<URL> {
-  const {loginUrl} = await vi.waitFor(() => {
-    const [startup] = mockedStartServerForTokenCallback.mock.settledResults
-    if (startup?.type !== 'fulfilled') throw new Error('Callback server is not listening yet')
-    return startup.value
-  }, 9000)
+async function waitForCallbackUrl(commandPromise: Promise<{error?: Error}>): Promise<URL> {
+  const commandExit = commandPromise.then(({error}) => {
+    throw error ?? new Error('Login command exited before the callback server started')
+  })
 
-  return new URL(loginUrl.searchParams.get('origin')!)
+  const startup = await Promise.race([
+    vi.waitFor(() => {
+      const [result] = mockedStartServerForTokenCallback.mock.settledResults
+      if (!result || result.type === 'incomplete') {
+        throw new Error('Callback server has not started yet')
+      }
+      return result
+    }, 9000),
+    commandExit,
+  ])
+
+  if (startup.type === 'rejected') throw startup.value
+  return new URL(startup.value.loginUrl.searchParams.get('origin')!)
 }
 
 /**
  * Simulates the OAuth provider redirecting back to the local callback server.
  * Waits for the server to be listening, then makes an actual HTTP request to it.
  */
-async function simulateOAuthCallback(sessionId: string): Promise<number> {
-  const callbackUrl = await waitForCallbackUrl()
+async function simulateOAuthCallback(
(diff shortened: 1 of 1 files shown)


filmaj (dismissed):
Love this! Well done

packages/@sanity/cli/src/commands/__tests__/login.test.ts
rexxars: As far as I can tell this doesn't actually assert against the port - if port 4000 happens to be taken (eg if working concurrently), the server falls back to other ports, and this assertion falkes.

Maybe we can get the real URL from `waitForCallbackUrl()` before completing the callback, assert its port is not 4321, and assert that `loginUrl` contains that actual port?

packages/@sanity/cli/src/commands/__tests__/login.test.ts
```diff
@@ -81,20 +93,34 @@ function mockStdin(input: string, options: {isTTY?: boolean} = {}) {
...
-): Promise<number> {
-  await new Promise((resolve) => setTimeout(resolve, delay))
-
-  const url = `http://localhost:${port}/callback?url=${encodeURIComponent(
+async function waitForCallbackUrl(): Promise<URL> {
+  const {loginUrl} = await vi.waitFor(() => {
+    const [startup] = mockedStartServerForTokenCallback.mock.settledResults
+    if (startup?.type !== 'fulfilled') throw new Error('Callback server is not listening yet')
+    return startup.value
+  }, 9000)
```
rexxars: If the login command fails before reaching `startServerForTokenCallback` (mock mismatch or similar, the kind of regression this file is meant to catch), this waits the full 9 seconds and then fails with the unhelpful "Callback server is not listening yet". 

Can we check if `startup?.type === 'rejected'`, and rethrow `startup.reason` immediately?

Ideally we'd also bail out when the command promise has already settled with an error (needs the command promise passed in or tracked)

rexxars (changes requested):

packages/@sanity/cli/src/commands/__tests__/login.test.ts
```diff
@@ -81,20 +93,49 @@ function mockStdin(input: string, options: {isTTY?: boolean} = {}) {
...
+ * Waits until the login command's local callback server is listening, and
+ * returns the callback URL it actually bound. The port can differ from 4321
+ * when the server falls back to another port.
+ *
+ * Rejects with the command's error if it exits before the server starts.
+ */
+async function waitForCallbackUrl(commandPromise: Promise<{error?: Error}>): Promise<URL> {
+  const commandExit = commandPromise.then(({error}) => {
+    throw error ?? new Error('Login command exited before the callback server started')
+  })
```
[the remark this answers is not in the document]
gu-stav: False positive. `Promise.race` attaches handlers to both promises up front, so this rejection is always observed even when the server branch wins the race — it never goes unhandled. Verified with a standalone repro, and vitest would flag it otherwise: unhandled rejections are exactly what the original flake surfaced as.

gu-stav:
@rexxars Great feedback 🙏🏼 I've addressed both comments in f9fc3b2bb9fe5a29cde51cdbd02b417854a1517f

rexxars (approved):

rexxars:
Thanks for looking into this 🙏

PR #1298, opened 2026-06-12
docs: correct guidance on test filtering

rexxars:
Agent docs used an incorrect `--filter` flag, and also could use some help on better ways of running individual tests.

---

> [!NOTE]
> **Low Risk**
> Documentation-only changes to AGENTS.md with no runtime or test behavior impact.
> 
> **Overview**
> Updates **AGENTS.md** so agent-oriented test docs match how Vitest is actually invoked in this repo.
> 
> Package-scoped runs now use **`pnpm test --project=<pkg>`** instead of the incorrect **`--filter`** flag (pnpm `--filter` is for turbo/package scripts, not Vitest project selection). The quick reference clarifies that a test file argument is a **path substring match**, with a shorter example path.
> 
> The “Running tests” section adds practical guidance for **single-file** runs (unique path fragment) and **single-test** runs via **`-t` regex** or **line number with full path**. The test-failure workflow and Cursor Cloud notes are aligned to use **`--project`** for scoped package runs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 744bb57b250038c8a56f1ff37eab9d15af59c356. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -5,7 +5,7 @@ Read and respect rules and conventions from CONTRIBUTING.md. Additionally:
 All commands are run from the root of the repo.
 
 - `pnpm test` - run all unit tests
-- `pnpm test <test-file>` - run specific test file. Example: `pnpm test packages/@sanity/cli/src/commands/documents/__tests__/get.test.ts`
+- `pnpm test <test-file>` - run specific test file(s); the arg is a path substring match. Example: `pnpm test documents/__tests__/get`
 - `pnpm test --coverage` - run tests with coverage report (output in `coverage/`)
 - `pnpm check:types` - TypeScript type checking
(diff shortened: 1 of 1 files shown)


filmaj:
The split `test:unit` vs. `test:integration` run scripts now also impact this. `test` now invokes both via `pnpm test:unit && pnpm test:integration`, which means any flags appended to the `test` command only apply to `test:integration`. Probably OK in the short term, but long-term we will either need a script to wrap invocation of both suites, or make the contributing and agent docs more explicit about the two different test suites.

snocorp (approved):

PR #1331, opened 2026-06-19
fix(cli): stop init crashing on a stalled npm version lookup

gu-stav:
### Description

Sometimes `sanity init` dies halfway through with a cryptic Node crash — exit code 13, "unsettled top-level await". It happens when one of the npm version lookups hangs and never comes back: the process runs out of other work, so Node gives up mid-init. That's the flaky e2e failure in [this run](https://github.com/sanity-io/cli/actions/runs/27828788243/attempts/1).

Each version lookup now gets a 30s cap. If one stalls, we fall back to the `latest` tag (npm resolves it at install time anyway) and carry on, instead of letting a single stuck request freeze the whole command.

### What to review

The timer is deliberately not `unref`'d — keeping it on the event loop is the whole trick, since that's what stops Node exiting early. Real lookup errors still bubble up and fail init like before; only the "hangs forever" case falls back.

I tried `AbortSignal.timeout()` first since it'd be the tidier, more native option — but its internal timer is `unref`'d, so the loop can still drain and crash before it fires. It wouldn't actually fix this, hence the plain referenced `setTimeout`.

### Testing

One unit test reproduces the hang (a lookup that never resolves) and checks we fall back to `latest`. It fails on `main` and passes with this change.

### DIFF
--- a/.changeset/fix-init-version-lookup-hang.md
+++ b/.changeset/fix-init-version-lookup-hang.md
@@ -1,0 +1,5 @@
+---
+'@sanity/cli': patch
+---
+
+fix(cli): prevent `init` from crashing when an npm version lookup stalls
--- a/packages/@sanity/cli/src/util/__tests__/resolveLatestVersions.test.ts
+++ b/packages/@sanity/cli/src/util/__tests__/resolveLatestVersions.test.ts
@@ -11,6 +11,7 @@ vi.mock('get-latest-version', () => ({
 describe('resolveLatestVersions', () => {
   afterEach(() => {
     vi.clearAllMocks()
+    vi.useRealTimers()
   })
 
   test('passes through valid semver ranges without looking them up', async () => {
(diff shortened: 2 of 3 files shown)


filmaj (approved):
That's pretty cool! How did you figure out that it was this particular version-lookup routine that was the source of the E2E test failure you linked to?

gu-stav:
@filmaj The captured output from the job had:

```
✔ Bootstrapping files from template
- Resolving latest module versions      <- started, never got its completed tick
Warning: Detected unsettled top-level await at .../bin/run.js:26
```

I searched the repo for the string and found `bootstrapLocalTemplate.ts`, which had the `await resolveLatestVersions`, so I thought that must have been the promise in flight that died, when the process timed out. The analysis about the event loop and exit code 13 was all Claude.

filmaj:
Awesome! Thanks for sharing. I had missed that line. Every little improvement like this adds up, so thanks for taking the time 🙇 ❤️

snocorp (approved):

PR #1373, opened 2026-06-24
refactor: move logic for getting env vars into cli-build package

snocorp:
### Description

This PR moves the logic for getting environment variables for apps and studios into the build package where it will mainly be used. One small change in this refactor is that we now directly reference `vite` in the module rather than referencing through a sepatate `loadEnv` module file.

### Testing

Moved related test file.

---

> [!NOTE]
> **Low Risk**
> Internal package boundary refactor with preserved re-exports; behavior should match aside from the intentional `loadEnv` import change and test-only exec error output.
> 
> **Overview**
> Moves **`getAppEnvironmentVariables`** and **`getStudioEnvironmentVariables`** (and **`StudioEnvVariablesOptions`**) from `@sanity/cli` into **`@sanity/cli-build`**, exposed via a new **`@sanity/cli-build/_internal/env`** export. Build, dev server, and static file paths in the CLI now import those helpers from cli-build instead of a local module; **`sanity/cli`**’s internal re-export is updated to forward from cli-build so existing consumers keep the same API.
> 
> In cli-build, env loading now calls **`loadEnv` from `vite`** directly instead of a separate **`loadEnv`** utility module. Unit tests mock **`@sanity/cli-build/_internal/env`** instead of the old local path; one studio unit test fixes a mistaken **`importOriginal`** type for the build mock.
> 
> Exec integration fixtures log failures with **`console.error(error)`** instead of JSON error payloads, and several exec tests register **`onTestFailed`** hooks to print stdout/stderr when a run fails.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0dcd0add8b77559f9ed44b956bbfe5161a291497. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/test/integration/commands/exec-with-user-token.test.ts
```diff
@@ -71,26 +71,46 @@ describe('exec --with-user-token', {timeout: 15 * 1000}, () => {
...
     vi.stubEnv('SANITY_AUTH_TOKEN', '')
     vi.stubEnv('SANITY_CLI_CONFIG_PATH', join(tmpdir(), 'sanity-cli-nonexistent', 'config.json'))
 
-    const {error} = await testCommand(ExecCommand, [scriptPath, '--with-user-token'])
+    const {error, stderr, stdout} = await testCommand(ExecCommand, [
+      scriptPath,
+      '--with-user-token',
+    ])
+    // output stdout and stderr to help diagnose the failure
+    t.onTestFailed(() => {
```
filmaj, line 84: Do you still want this here?

filmaj (approved):
Looks fine to me. If you rebase on `main` I think the unit tests will pass.

packages/@sanity/cli/test/integration/commands/exec-with-user-token.test.ts
snocorp: I guess we could just add it manually locally in case of failure but I thought there's no harm in leaving it in place for the next time I break things.

PR #1379, opened 2026-06-24
fix(workbench): preview release @module-federation/vite to 1.16.11

gu-stav:
### Description

> [!NOTE]
>  Makes the upstream fix https://github.com/module-federation/vite/pull/854 available as preview release until 1.16.11 is released.


gioboa (commented):
Are those errors related to the new mfe/vite version?

gu-stav:
@gioboa Thanks for checking in :) They are not related.

gu-stav:
Superseded by #1389, which moves `@module-federation/vite` to the released 1.16.11 instead of the pkg.pr.new preview.

PR #1390, opened 2026-06-25
test(chore): unit test `SanityCommand` `getProjectId`, convert `dataset import` tests to unit

filmaj:
### Description

This PR adds a unit test file to the OCLIF custom base class `SanityCommand`, allowing for testing the methods on this class. That should allow us to test the methods in this class, which are shared across all `SanityCommand` child classes, without having to test them in each command.

Also:
- converts all of the `dataset import` tests to unit tests
- duplicate tests exercising `SanityCommand.getProjectId` logic is removed

### Testing

---

> [!NOTE]
> **Low Risk**
> Changes are primarily test structure and deduplication; the only runtime tweak is dataset import messaging/error handling via `this.output`, with no auth or data-path changes.
> 
> **Overview**
> Adds **`SanityCommand` unit tests** in `@sanity/cli-core` that exercise `getProjectId` end-to-end via OCLIF (`--project-id`, deprecated `--project`/`-p`, CLI config fallback, custom fallback, and the “unable to determine” error).
> 
> **`dataset import` tests** are rewritten as unit tests using `createMockSanityCommand` (mocked `SanityCommand`, direct `ImportDatasetCommand.run`, assertions on `output.log/warn/error`) instead of `testCommand` integration-style harnesses.
> 
> **Duplicate per-command tests** for project ID resolution (`--project-id`, missing project ID, etc.) are removed from copy, list, documents, graphql undeploy, and media import suites—those behaviors are now covered once on the base class.
> 
> The **`import` command** is aligned with the shared output API: `this.warn`/`this.log`/`this.error` become `this.output.*`, with **`return`** on error paths so failures exit cleanly in unit tests.
> 
> Vitest **`testTimeout: 10_000`** is set for `@sanity/cli` and `@sanity/cli-core` unit configs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47be2f292e8f8e6d4fa436818199379a1b929199. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli-core/src/__tests__/SanityCommand.test.ts
```diff
@@ -0,0 +1,118 @@
...
+describe('SanityCommand', () => {
+  afterEach(() => {
+    vi.clearAllMocks()
+  })
+  describe('getProjectId', () => {
+    test('returns --project-id value if provided', async () => {
+      const cmdClass = createMockedRunCommand({
+        run: async function () {
+          const id = await this.getProjectId()
+          expect(id).toEqual('torment nexus')
```
snocorp: I know this is unlikely but I think this test could pass if the run function was never invoked. Would it make sense to move the expect after the awaited call below and store the id in a variable accessible to both scopes? I think we could also add an `assertRunCalled` function to the mock if that's easier

filmaj: Ya i was thinking of that too... I was wondering if vitest had a like, `count` assertion, where each test requires a certain number of assertions to execute in order to pass. I know tap has that. I could do the closure variable trick too. But ya, good call

snocorp (approved):

PR #1395, opened 2026-06-26
refactor: move compareDependencyVersions out of internal build logic

snocorp:
### Description

In order to avoid moving compareDependencyVersions into the cli-build package and then exporting it to cli for the dev server and to make it easier to mock the logic from tests in the cli package, I've moved the call out of the internal logic into the cli package logic.

Note that we will eventually have to duplicate the return type into cli-build here as part of the options properties. The only alternative is to push this util into cli-core but that would then make mocking difficult again so I went with this option.

### Testing

No functionality changes here just adjusting the dependency chain.

---

> [!NOTE]
> **Low Risk**
> Refactor-only wiring with no change to comparison logic or build output; risk is limited to a mistaken callback binding, which existing unit tests already mock.
> 
> **Overview**
> **Dependency version checks** for auto-updates are no longer invoked directly inside `internalBuildApp` and `internalBuildStudio`. Both internal builders now take a **`compareDependencyVersions` callback** on their options; the public `buildApp` and `buildStudio` entry points supply the real implementation bound to `workDir` and `appId`.
> 
> `CompareDependencyVersionsResult` is **exported** from `compareDependencyVersions.ts` so the injected callback can be typed on `InternalBuildOptions`. Behavior of the version check is unchanged—only **where the call is wired** and **how tests can substitute** the logic.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6b2147a44e310b210432f431b63618cd1b3d6f39. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/build/buildApp.ts
+++ b/packages/@sanity/cli/src/actions/build/buildApp.ts
@@ -25,7 +25,10 @@ import {resolveWorkbenchApp} from '@sanity/workbench-cli/build'
 import {parse as semverParse} from 'semver'
 
 import {getAppId} from '../../util/appId.js'
-import {compareDependencyVersions} from '../../util/compareDependencyVersions.js'
+import {
+  compareDependencyVersions,
+  CompareDependencyVersionsResult,
+} from '../../util/compareDependencyVersions.js'
 import {formatModuleSizes, sortModulesBySize} from '../../util/moduleFormatUtils.js'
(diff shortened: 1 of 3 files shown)


filmaj (approved):
Dependency injection, always a 👍 from me!

PR #1396, opened 2026-06-26
refactor(cli): move missing appid warning out of internal logic

snocorp:
### Description

Another dependency refactor. Since the blueprints deployment won't use this logic, we can keep it in the cli package instead.

### Testing

No logic changes here, just moving the dependency chain.

---

> [!NOTE]
> **Low Risk**
> Refactor-only with the same warning conditions preserved at the public build wrappers; no auth, data, or deploy behavior changes beyond call-site organization.
> 
> **Overview**
> Refactors **auto-updates missing `appId` warnings** so they no longer live inside `internalBuildApp` / `internalBuildStudio`.
> 
> `InternalBuildOptions` drops **`calledFromDeploy`** (and studio drops **`projectId`** used only for that warning). In their place, callers pass **`checkAppId`**, which `buildApp` and `buildStudio` implement with the same rules as before: warn via `warnAboutMissingAppId` when auto-updates are on and there is no `appId`, unless **`calledFromDeploy`** is set on the outer build options. The internal path only invokes **`options.checkAppId()`** when auto-updates are enabled.
> 
> Behavior is intended to be unchanged; the split keeps deploy-specific skip logic and warning helpers at the CLI entry layer rather than in shared internal build code (e.g. for paths like blueprints that won’t use this warning chain).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ded37b9cf3b943fa199f691d331955e61b04dd19. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/build/buildApp.ts
+++ b/packages/@sanity/cli/src/actions/build/buildApp.ts
@@ -40,7 +40,7 @@ interface InternalBuildOptions {
   appId: string | undefined
   appTitle: string | undefined
   autoUpdatesEnabled: boolean
-  calledFromDeploy: boolean | undefined
+  checkAppId: () => void
   compareDependencyVersions: (
     packages: {name: string; version: string}[],
   ) => Promise<CompareDependencyVersionsResult>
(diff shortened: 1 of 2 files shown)


macdonst (approved):
LGTM

PR #1398, opened 2026-06-26
test(chore): convert `mcp configure` tests to unit

filmaj:
### Description

Converting more integration tests to unit. This time for the `mcp configure` code path.

- Added a unit test for the `mcp configure` Command
- Many of the 'editor' test cases already existed as unit tests in `src/actions/mcp/__tests__/editorConfigs.test.ts`
- Any integration tests missing from `setupMCP` unit tests were added.

---

> [!NOTE]
> **Low Risk**
> Production changes are limited to configure command control flow and error reporting; MCP setup behavior is unchanged aside from tests and a trivial fs import.
> 
> **Overview**
> Replaces the large **`mcp configure` integration suite** with **narrow unit tests** that mock `ensureAuthenticated` and `setupMCP`, and extends **`setupMCP`** / **`detectAvailableEditors`** coverage for scenarios that used to live in integration tests (no editors, all configured, token creation failure, parse/skip behavior).
> 
> **`ConfigureMcpCommand`** now picks **`auto` vs `prompt`** via **`this.isUnattended()`** instead of `isInteractive()`, and reports auth/setup failures through **`this.output.error` with `return`** rather than `this.error`. **`mockSanityCommand`** gains a **telemetry** stub so configure tests can run **`run()`** without real tracing.
> 
> **`detectAvailableEditors`** only switches to a **named `readFile` import** (no behavior change intended).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e40dc40ba4f448cf80c240a719993f28686540df. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/commands/mcp/configure.ts
```diff
@@ -30,19 +30,21 @@ export class ConfigureMcpCommand extends SanityCommand<typeof ConfigureMcpComman
...
+      return this.output.error(`Failed to check authentication: ${getErrorMessage(error)}`, {
+        exit: 1,
+      })
     }
 
     try {
       const mcpResult = await setupMCP({
         explicit: true,
-        mode: isInteractive() ? 'prompt' : 'auto',
+        mode: this.isUnattended() ? 'auto' : 'prompt',
```
snocorp, line 47: I think these are slightly different. Is interactive means there's a tty available and is unattended means the `--yes` flag was used

runeb: Hopefully isInteractive also reflects `-y` usage, but yes, it is "interactive" you want to test for here not just unattended

runeb (changes requested):
Thanks! Check for "interactive" instead of "unattended" and needs a rebase

packages/@sanity/cli/src/commands/mcp/configure.ts
filmaj: Interesting. I wonder if that should be folded into the `SanityCommand` interface. Both seem quite related? One is provided as part of the extended OCLIF `SanityCommand` while the other is imported.

filmaj:
I am going to close this PR, as the unit test approach in this PR was an early attempt that is no longer super applicable.

PR #1412, opened 2026-06-30
refactor: move internal build logic to cli-build package

snocorp:
### Description

This is the last PR needed before integrating the build logic into the runtime CLI. The logic for building studios and apps has been ported to the cli-build package.

> [!NOTE]
> This is a breaking change for the cli-build package as it removed some un used exports. The cli package is the only dependent of the cli-build package so this is a safe bump. 

### Testing

Tests were ported or updated as needed. Note that I did not port the buildStaticFiles test as it requires a fixture that is not available in the cli-build package and would add a lot of complexity just for that one test suite if ported.

---

> [!NOTE]
> **Medium Risk**
> Large refactor on the primary `sanity build` path and auto-updates/version prompts, but behavior is largely relocated with tests ported; risk is mainly regression in build/deploy and breaking internal cli-build consumers.
> 
> **Overview**
> Moves the full **Sanity app and Studio build implementations** from `@sanity/cli` into `@sanity/cli-build`, so the runtime CLI can depend on one package for build behavior. `@sanity/cli` **build commands become thin wrappers** that resolve config/flags/workbench and call `buildApp` / `buildStudio` with injected hooks (`compareDependencyVersions`, `checkAppId`, `upgradePackages`, `determineBasePath`).
> 
> The cli-build APIs take a **flag-free `BuildOptions` surface** (e.g. `unattendedMode`, `minify`, `sourceMap`) instead of reading CLI flags directly. **Tests for build flows move** into cli-build (with dependency comparison injected in tests); CLI tests focus on wiring such as missing `appId` warnings.
> 
> **`@sanity/cli-build/_internal/build` export changes** (major bump): exports `buildApp` and `buildStudio`; re-exports compare-version types; drops several previously exported build helpers (telemetry traces, module size formatters, full auto-updates import-map exports).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cafb038cb115d45eb9363c8808a5460e51d2d46a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli-build/src/actions/build/__tests__/buildApp.test.ts
```diff
@@ -1,44 +1,58 @@
 import {Output} from '@sanity/cli-core'
```
filmaj: I think this is a `type`, right?

packages/@sanity/cli-build/src/actions/build/__tests__/buildApp.test.ts
```diff
@@ -1,44 +1,58 @@
...
 
-const FLAGS = {
-  'auto-updates': true,
-  json: false,
-  minify: true,
-  'source-maps': true,
-  stats: true,
-  yes: false,
-} as const
+import {BuildOptions} from '../buildApp.js'
```
filmaj: This one's also a `type` - I thought we had a lint rule for this? I think that difference in use with `import` is important to prevent bloating the import graph. `import` on its own imports the runtime code, `import type` only brings in the type definitions.

packages/@sanity/cli-build/src/actions/build/__tests__/buildStudio.test.ts
```diff
@@ -0,0 +1,460 @@
+import {Output} from '@sanity/cli-core'
```
filmaj: Same comments here, let's use `import type` where appropriate.

packages/@sanity/cli-build/src/actions/build/__tests__/buildStudio.test.ts
```diff
@@ -0,0 +1,460 @@
...
+    },
+    void
+  >,
+}))
+
+vi.mock(import('../getEnvironmentVariables.js'), () => ({
+  getStudioEnvironmentVariables: mockGetStudioEnvironmentVariables,
+}))
+
+vi.mock('@sanity/cli-core', async (importOriginal) => {
```
filmaj: If you pass in the first parameter wrapped in `import()`, then you can drop the use of the generic in the call to `importOriginal`. Also, if possible, please avoid using `importOriginal`, as that imports the full weight of the code. So calling `importOriginal()` imports _all_ of cli-core, which will e.g. import the `jsdom` dependency which automatically bloats the test import time by ~250ms on a modern M4 Macbook Pro (and obvs much worse in CI). I understand that can't be done easily, esp. for e.g. all of `cli-core`, but I just ask we keep that in mind, and I think it's more doable with e.g. the `cli-core/ux` import below.

packages/@sanity/cli-build/src/actions/build/__tests__/buildStudio.test.ts
```diff
@@ -0,0 +1,460 @@
...
+  return {
+    ...original,
+    confirm: mockedConfirm,
+    select: mockedSelect,
+    spinner: mockedSpinner,
+  }
+})
+
+// Import after mocks are set up
+const {buildStudio} = await import('../buildStudio.js')
```
filmaj: I've learned that this pattern of "`await import` of the module-under-test after all the vitest mocking is done" is not necessary _as long as_ all the mocking code before pulling in the module-under-test is either only `vi.mock` calls, or wrapped in `vi.hoisted`. Both of those `vi` APIs execute _before_ any `import`s. This is a function of vitest's re-writing of test code and JIT transpilation. Some of this is explained in the ["How mocking works" vitest guide](https://vitest.dev/guide/mocking/modules.html#how-it-works), the [`vi.mock` API docs](https://vitest.dev/api/vi.html#vi-mock) and the [`vi.hoisted` API docs](https://vitest.dev/api/vi.html#vi-hoisted).

We may also want to consider to start using [`vi.importModule`](https://vitest.dev/api/vi.html#vi-importmock), which auto-mocks all exports of a module, enabling easy spying. That may be easier than declaring a separate mock function for every module's export.

packages/@sanity/cli-build/src/actions/build/buildApp.ts
```diff
@@ -0,0 +1,226 @@
...
+import {styleText} from 'node:util'
+
+import {
+  type CliConfig,
+  getCliTelemetry,
+  getLocalPackageVersion,
+  getTimer,
+  isInteractive,
+  type Output,
+  UserViteConfig,
```
filmaj: This one's a `type` too

packages/@sanity/cli-build/src/actions/build/buildStudio.ts
```diff
@@ -0,0 +1,277 @@
...
+import {styleText} from 'node:util'
+
+import {
+  type CliConfig,
+  getCliTelemetry,
+  getLocalPackageVersion,
+  getTimer,
+  isInteractive,
+  type Output,
+  UserViteConfig,
```
filmaj: `type`!

filmaj (dismissed):
Left some non-blocking comments but we should tee up follow-up PRs and/or file backlog issues if not addressed.

packages/@sanity/cli-build/src/actions/build/__tests__/buildApp.test.ts
snocorp: Yup, nice catch, not sure why the linter isn't picking it up. It does in runtime's repos I think (with biome)

packages/@sanity/cli-build/src/actions/build/__tests__/buildStudio.test.ts
snocorp: I think I will leave this as a follow up exercise. Worth doing though I think

filmaj (approved):

PR #1407, opened 2026-06-30
refactor(deploy): restructure studio deploy on the shared deploy checks

gu-stav:
### Description

Second of two PRs splitting the deploy refactor; builds on the core app PR (#1406). This one restructures the **studio** deploy onto the shared scaffolding introduced there.

`deployStudio` now runs through a single `createStudioDeployment` using the same checks seam, with studio target resolution added alongside the app rules (it subsumes the inline host/URL validation). With both deploy paths refactored, the two interactive "find application" adapters collapse into one module and the two "create application" helpers into another.

No behavior change — this is a restructure.

### What to review

- `createStudioDeployment` against the old `deployStudio`: prompts, spinners, error messages and exit codes should be unchanged.
- The merged `findUserApplication` / `createUserApplication` modules and the studio target resolution in `resolveDeployTarget`.

### Testing

The integration deploy tests pass unchanged.

---

> [!NOTE]
> **Medium Risk**
> Touches the full studio deploy CLI path (target resolution, prompts, builds, schema upload, API deploy) across a large refactor; risk is mitigated by parity intent and new unit tests, but regressions in edge cases (external URL, unattended, federated + --external) are still plausible.
> 
> **Overview**
> Studio and core-app deploy now share **`runDeploy`** (fail-fast reporter + normalized errors), the same **check** helpers (`checkAutoUpdates`, `checkPackageVersion`, `checkBuild`, `verifyOutputDir`), and reporter-based config failures with optional **`solution`** hints appended on fail/warn.
> 
> **Studio** target resolution moves into **`resolveStudioDeployTarget`** (host/`--url` validation, `appId` precedence, internal vs external) with a matching dry-run **`checkStudioTarget`**. Interactive lookup is unified in **`findUserApplication`** / **`findUserApplicationForStudio`**; studio and core-app creation helpers merge into **`createUserApplication`**.
> 
> **`deployStudio`** is split into `runStudioDeployment` plus smaller helpers (`resolveStudioApplication`, `uploadStudioSchema`, `shipStudioDeployment`) instead of one monolithic flow with inline host validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9d6a500c3c6fd6ba799f970661d4dbae517cfb8e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1415, opened 2026-07-01
feat(deploy): add a --dry-run flag to the deploy command

gu-stav:
### Description

> [!NOTE]
> Third in the deploy stack — builds on #1407. This is the payoff the earlier PRs laid the foundation for.

`sanity deploy --dry-run` reports whether the studio or app can be deployed and lists the files a deploy would upload (with sizes) — without uploading, creating, or prompting for anything. When something blocks the deploy it lists every problem with a fix, and surfaces warnings.

A blocked studio:

```
Dry run — no changes made.

  ✓ Using sanity 3.99.0
  ✓ Studio built

This studio can't be deployed.

Problems to fix:
  ✗ No studio hostname configured: Set `studioHost` in sanity.cli.ts, or pass a hostname with --url

Warnings:
  ⚠ The autoUpdates config has moved to deployment.autoUpdates.

Files to deploy (0.50 MB):
  dist/index.html (0.50 MB)
```

Each check carries its fix in a `solution` field, so the same fix renders in both modes — appended after the problem in the dry-run report and in a real deploy's error (instead of baking the fix into the message).

A deployable one just lists the files:

```
Dry run — no changes made.

  ✓ Using sanity 3.99.0
  ✓ Project: my-project
  ✓ Deploys to existing studio https://my-studio.sanity.studio
  ✓ Studio built

This studio can be deployed.

Files to deploy (1.53 MB):
  dist/assets/app.js (1.53 MB)
  dist/index.html (0.00 MB)
```

The design property that matters: dry-run drives the **same** `runDeploy` sequence (`runAppDeployment` / `runStudioDeployment`, harmonized in #1407) as a real deploy, so the two can't drift. The only difference is the mode — a collecting reporter instead of the fail-fast one, and every network mutation gated off: the target is resolved read-only (`checkAppTarget` / `checkStudioTarget`, the same verdict resolver the real path uses), and the title update, schema upload, and deployment upload sit below a single dry-run stop. The build still runs locally, so the file list is accurate.

A small `deploymentPlan` module owns the plan shape and the report — a deep seam that's easy to extend later (schema diff, more per-file detail, …).

Deliberately not covered yet: schema validation doesn't run in a dry run, since it's coupled to the upload step. Straightforward follow-up on the same seam.

### What to review

- The dry-run stop + read-only target resolution in `runAppDeployment` / `runStudioDeployment` — confirm a dry run can't reach any POST or prompt.
- `deploymentPlan.ts` (plan shape + render) and that the command stays a thin caller.

### Testing

Unit tests cover `listDeploymentFiles` and `renderDeploymentPlan` (including the problems-with-fixes and warnings sections); integration tests run `deploy --dry-run` for a studio and an app and assert the report renders with no deployment POST (none is mocked, so a ship would fail). E2E tests (`deploy.test.ts`) drive the real CLI against both a studio and a core app, asserting the plan lists real built files including `index.html`, surfaces the blocking problem with its fix, and reports the deprecated `--auto-updates` flag as a warning — skipped in registry/smoke mode since `--dry-run` isn't on the published CLI yet.

---

> [!NOTE]
> **Medium Risk**
> Touches the core deploy pipeline and exit-code behavior for CI gating, but mutations stay gated and coverage is broad (unit, integration, e2e).
> 
> **Overview**
> Adds **`sanity deploy --dry-run`**, which runs the same studio/app deploy validation and local build as a real deploy, then prints a **deployment plan** (checks, problems with fixes, warnings, and file list with sizes) without uploads, creates, or prompts.
> 
> **`runDeploy`** switches on the flag: a collecting reporter runs the shared **`runAppDeployment` / `runStudioDeployment`** path; **`renderDeploymentPlan`** prints the report; blocked plans skip file listing and exit with the **same code** as fail-fast deploy. Dry-run paths use read-only **`checkAppTarget` / `checkStudioTarget`** and return before schema upload, title sync, and tarball POST.
> 
> Shared **`describeAppTarget` / `describeStudioTarget`**, **`checkAppId`** (via **`resolveAppIdIssue`**), and **`findUserApplication`** refactors keep dry-run and unattended real-deploy messages and exit codes aligned. **`--dry-run`** also skips the non-empty output-dir confirm and forces **`yes: true`** for build prompts.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1333f260d381162f40aacdb30024d298f1b3ce83. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/deploy/__tests__/deploymentPlan.test.ts
+++ b/packages/@sanity/cli/src/actions/deploy/__tests__/deploymentPlan.test.ts
@@ -1,9 +1,9 @@
-import {mkdir, mkdtemp, rm, writeFile} from 'node:fs/promises'
-import {tmpdir} from 'node:os'
+import {type Dirent, type Stats} from 'node:fs'
+import {readdir, stat} from 'node:fs/promises'
 import {join} from 'node:path'
 
 import {type Output} from '@sanity/cli-core'
-import {afterEach, beforeEach, describe, expect, test} from 'vitest'
+import {beforeEach, describe, expect, test, vi} from 'vitest'
 
 import {type DeployCheck} from '../deployChecks.js'
 import {
@@ -13,23 +13,34 @@ import {
   renderDeploymentPlan,
 } from '../deploymentPlan.js'
 
+vi.mock(import('node:fs/promises'), async (importOriginal) => ({
+  ...(await importOriginal()),
+  readdir: vi.fn(),
+  stat: vi.fn(),
+}))
+
+const mockReaddir = vi.mocked(readdir)
+const mockStat = vi.mocked(stat)
+
+// Minimal Dirent stand-in: listDeploymentFiles only reads `name` and `isDirectory()`.
+const dirent = (name: string, isDirectory: boolean): Dirent =>
+  ({isDirectory: () => isDirectory, name}) as Dirent
+
 describe('listDeploymentFiles', () => {
-  let dir: string
-
-  beforeEach(async () => {
-    dir = await mkdtemp(join(tmpdir(), 'deploy-plan-'))
-  })
-
-  afterEach(async () => {
-    await rm(dir, {force: true, recursive: true})
-  })
+  beforeEach(() => vi.clearAllMocks())
 
   test('lists nested files as sorted paths with sizes, relative to fromDir', async () => {
-    await mkdir(join(dir, 'dist', 'assets'), {recursive: true})
-    await writeFile(join(dir, 'dist', 'index.html'), 'x')
-    await writeFile(join(dir, 'dist', 'assets', 'app.js'), 'xyz')
+    mockReaddir.mockImplementation((async (dir: string) => {
+      if (dir.endsWith(join('dist', 'assets'))) return [dirent('app.js', false)]
+      if (dir.endsWith('dist')) return [dirent('index.html', false), dirent('assets', true)]
+      return []
+    }) as unknown as typeof readdir)
+    mockStat.mockImplementation(
+      (async (file: string) =>
+        ({size: file.endsWith('app.js') ? 3 : 1}) as Stats) as unknown as typeof stat,
+    )
 
-    const files = await listDeploymentFiles(join(dir, 'dist'), dir)
+    const files = await listDeploymentFiles(join('/root', 'dist'), '/root')
 
     expect(files).toEqual([
       {path: 'dist/assets/app.js', size: 3},
@@ -38,7 +49,8 @@ describe('listDeploymentFiles', () => {
   })
 
   test('returns an empty list when the directory is missing', async () => {
-    expect(await listDeploymentFiles(join(dir, 'missing'), dir)).toEqual([])
+    mockReaddir.mockRejectedValue(new Error('ENOENT'))
+    expect(await listDeploymentFiles(join('/root', 'missing'), '/root')).toEqual([])
   })
 })
 
@@ -85,7 +97,8 @@ describe('renderDeploymentPlan', () => {
     expect(text).toContain("This studio can't be deployed.")
     expect(text).toContain('Problems to fix:')
     expect(text).toContain('No project ID configured: Add `api.projectId`')
-    expect(text).toContain('Files to deploy (0.00 MB):')
+    // No files to list, so the section is omitted rather than shown as "0.00 MB"
+    expect(text).not.toContain('Files to deploy')
   })
 
   test('surfaces warnings in their own section', () => {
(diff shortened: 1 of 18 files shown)


packages/@sanity/cli/test/integration/commands/deploy.app.test.ts
snocorp, line 1: Ideally we should be relying on unit tests more than integration tests. I think one new integration test is probably sufficient here if we can cover the rest of the logic without fixtures and file i/o

packages/@sanity/cli/test/integration/commands/deploy.studio.test.ts
snocorp, line 1: Similar to [deploy.app.test.ts](https://github.com/sanity-io/cli/pull/1415/changes#diff-80a556da1498916dc48f39eaf8d44c33f069684e98379b74b1bf601aae944995) I think one new test with the rest covered in unit tests would be ideal here.

packages/@sanity/cli/src/actions/deploy/__tests__/deploymentPlan.test.ts
snocorp, line 1: Since this file does real file i/o it should be move into the integration tests folder. I think it's even better to mock out the file i/o and make this a unit test as well

packages/@sanity/cli/src/actions/deploy/__tests__/deploymentPlan.test.ts
```diff
@@ -0,0 +1,111 @@
...
+        {message: 'No project ID configured', solution: 'Add `api.projectId`', status: 'fail'},
+      ]),
+      output,
+    )
+
+    const text = lines.join('\n')
+    expect(text).toContain("This studio can't be deployed.")
+    expect(text).toContain('Problems to fix:')
+    expect(text).toContain('No project ID configured: Add `api.projectId`')
+    expect(text).toContain('Files to deploy (0.00 MB):')
```
snocorp: Seems odd to output this when there is a problem

packages/@sanity/cli/src/actions/deploy/deploymentPlan.ts
```diff
@@ -0,0 +1,117 @@
...
+      : styleText('red', `\nThis ${label} can't be deployed.`),
+  )
+
+  renderIssues(output, 'Problems to fix:', problems)
+  renderIssues(output, 'Warnings:', warnings)
+
+  output.log(`\nFiles to deploy (${formatMB(totalBytes)}):`)
+  for (const file of plan.files) {
+    output.log(`  ${file.path} (${formatMB(file.size)})`)
+  }
```
snocorp: I left a comment in the unit test for this but if there is a problem that results in no files, this could look strange.

snocorp (changes requested):
Only one potential issue with the logic around rendering the deployment plan. Also requesting a few changes to the testing approach to avoid heavy integration tests as much as possible. I think it's worth the extra effort and will give us better confidence in our changes in the long run.

packages/@sanity/cli/src/actions/deploy/__tests__/deploymentPlan.test.ts
gu-stav: Agree

packages/@sanity/cli/src/actions/deploy/deploymentPlan.ts
gu-stav: Agree

packages/@sanity/cli/src/actions/deploy/deployChecks.ts
```diff
@@ -181,144 +209,136 @@ export async function verifyOutputDir({
...
+    }
+    case 'needs-input': {
+      return {
+        exitCode: exitCodes.USAGE_ERROR,
+        message: isExternal ? 'No external studio URL configured' : 'No studio hostname configured',
+        solution: isExternal
+          ? 'Set `studioHost` in sanity.cli.ts, or pass the full URL with --url'
+          : 'Set `studioHost` in sanity.cli.ts, or pass a hostname with --url',
+        status: 'fail',
+      }
```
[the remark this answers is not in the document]
gu-stav: Intentional asymmetry: for apps, `deployment.appId` is the only non-interactive selector, so the count explains what a prompt would offer and why the ID is needed. For studios the fix is naming a host (`studioHost` or `--url`), which works the same whether that host already exists or not, so a count wouldn't change what you type. Listing the actual hostnames might help, but that's too much for a one-line diagnosis and the interactive prompt already shows them. Studio needs-input can also have 0 existing studios, so the app phrasing wouldn't carry over cleanly.

snocorp (approved):

PR #1416, opened 2026-07-01
feat(deploy): add --json flag

gu-stav:
### Description

> [!NOTE]
> Stacked on #1415 — adds `--json` on top of `--dry-run`.

`sanity deploy --json` prints machine-readable JSON instead of the human report, independent of `--dry-run`:

- `deploy --dry-run --json` → the plan (`applicationType`, `applicationVersion`, `isDeployable`, `errors`, `warnings`, `files`, `totalBytes`)
- `deploy --json` → the real deploy result (`deployed`, `applicationId`, `applicationType`, `applicationVersion`, studio `location`)

Both the human report and the JSON derive from the same checks, so they can't drift. Blocking problems come back as `errors` — a map of problem → fix — with `warnings` as a list of messages and a top-level `isDeployable` flag, so an agent gets the same "why it can't deploy" the human report shows.

In JSON mode the run's stdout is routed to stderr so a sub-step that prints progress straight to stdout (the build, the schema upload) can't corrupt the payload — only the JSON reaches stdout.

Follows the existing `--json` convention (`doctor`, `tokens list`, `schemas list`): an explicit flag, no auto-detection.

### What to review

- `deployRunner.ts` — the JSON branch and the `redirectStdoutToStderr` guard that keeps stdout clean.
- `deploymentPlanToJson` in `deploymentPlan.ts` — shares the plan with the human renderer.

### Testing

Unit covers `deploymentPlanToJson`; integration runs `deploy --dry-run --json` and a real `deploy --json`, parsing the payload and asserting no human text leaks to stdout.

---

> [!NOTE]
> **Low Risk**
> CLI output and typing changes around deploy; deploy behavior is unchanged unless `--json` is used, with broad test coverage.
> 
> **Overview**
> Adds **`sanity deploy --json`** (`-j`) so deploy output can be consumed by scripts and agents, matching other CLI commands.
> 
> With **`--dry-run --json`**, stdout is a plan: `applicationType`, `applicationVersion`, `target` (id, title, URL), `isDeployable`, `errors` (message → fix), `warnings`, `files`, and `totalBytes`. A successful real **`deploy --json`** prints `{ deployed: true, ... }` with the same `target` shape plus framework version (and optional `installationId` for config-only singleton deploys). Human and JSON views share the same deploy checks—`target` and `version` are attached on checks and aggregated in `collectPlan` / `deploymentPlanToJson` so they cannot drift.
> 
> In JSON mode, deploy progress **`output.log`** is redirected to **stderr** so only the final JSON hits stdout; failures stay on stderr with no JSON payload. The deploy command skips the “Building to …” line when `--json` is set.
> 
> **Core apps** now surface the org dashboard URL (`getCoreAppUrl`) in dry-run target messages and success output (with optional title), parallel to studio URLs. Typing adds **`UserApplicationResolved`** for org-scoped core apps used in those URLs.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 782e21595e19c63cb3784697632dde49cd493b06. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


runeb (commented):
Adding more json output from cli commands is excellent, and where we want to go. But the bot review points at something I think we need to consider, non-interactive mode. Does it make sense to support interactive prompts when the output requested is json? We could treat json output as non interactive and error early on required params the same way we do `-y`.

gu-stav:
@runeb I think that's a good idea! Opened a new PR https://github.com/sanity-io/cli/pull/1420 to change that across the CLI.

packages/@sanity/cli/src/actions/deploy/deployRunner.ts
```diff
@@ -1,56 +1,110 @@
...
-    return
-  }
 
-  try {
-    await spec.run(options, createFailFastReporter(output))
+    const result = await spec.run(options, createFailFastReporter(output))
+    restoreStdout()
+    if (json && result) output.log(JSON.stringify({deployed: true, ...result}, null, 2))
   } catch (error) {
+    restoreStdout()
```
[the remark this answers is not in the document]
gu-stav: Intentional: `--json` shapes the success output, failures are non-zero exit + stderr — same contract as every other `--json` command in this CLI (`tokens add`, `projects create`, `doctor`). The dry-run payload isn't an exception: a dry run's product is the report itself, so a blocked plan is a successful report (with `isDeployable: false`), and it still exits non-zero. A failed real deploy leaves stdout deliberately empty, so automation that checks the exit code gets an unambiguous signal.

packages/@sanity/cli/src/actions/deploy/deployRunner.ts
```diff
@@ -1,56 +1,110 @@
...
+ * in `runDeploy`.
  */
 export interface DeploySpec {
   /** Files a real deploy would upload, listed only for the dry-run plan. */
   listFiles: (options: DeployAppOptions) => Promise<DeploymentFile[]>
+  /** The framework package whose installed version the deploy reports. */
+  packageName: string
   /** The step sequence; every step reports through `reporter`. */
-  run: (options: DeployAppOptions, reporter: CheckReporter) => Promise<void>
+  run: (options: DeployAppOptions, reporter: CheckReporter) => Promise<DeployResult | void>
```
snocorp, line 46: Might be good to document when a deploy result is returned or not in the comment above.

packages/@sanity/cli/src/actions/deploy/deployRunner.ts
```diff
@@ -1,56 +1,110 @@
...
-    await spec.run(options, reporter)
-    const failed = reporter.results.find((check) => check.status === 'fail')
-    const plan: DeploymentPlan = {
-      checks: reporter.results,
-      // A blocked deploy uploads nothing, so only enumerate files for a deployable plan.
-      files: failed ? [] : await spec.listFiles(options),
-      type: spec.type,
+  // In JSON mode sub-steps (build, schema upload) print progress straight to
+  // stdout, so route the whole run's stdout to stderr and emit only the payload.
+  const restoreStdout = json ? redirectStdoutToStderr() : () => {}
```
snocorp: This is interesting and maybe unfortunate that you need to do it. I would say that all output should go to stderr unless it is intended for use as a value that could be piped to another command. Maybe something we should put into the dev/agent docs at the top level as a pattern to be followed.

snocorp (dismissed):

packages/@sanity/cli/src/actions/deploy/deployRunner.ts
gu-stav: Thank you for highlighting this. I wasn't happy with it in the first place and have reworked it in 2af8502246d5107e806ce153a41b366dd363464c

packages/@sanity/cli/src/actions/deploy/deployRunner.ts
```diff
@@ -1,59 +1,114 @@
...
-  }
+  // The JSON payload owns stdout, so the run's progress logs go to stderr; only
+  // the final JSON.stringify writes to stdout. Spinners are already on stderr.
+  const runOptions = json
+    ? {
+        ...options,
+        output: {
+          ...output,
+          log: (message = '', ...args: unknown[]) =>
+            void process.stderr.write(`${format(message, ...args)}\n`),
```
snocorp, line 67: I like this approach much better.

snocorp (approved):

PR #1418, opened 2026-07-01
test: Mock spinner

snocorp:
### Description

Mock spinner allows tests to avoid sending output to stderr.

### Testing

Fixed the bootstrapLocalTemplate tests and added some assertions as an example.

---

> [!NOTE]
> **Low Risk**
> Test-only and test-helper changes with no production runtime behavior.
> 
> **Overview**
> Adds **`createMockSpinner`** to `@sanity/cli-test`: a Vitest mock factory for `@sanity/cli-core/ux`’s `spinner` that returns a full `SpinnerInstance` stub (with optional overrides) so CLI tests avoid real terminal output.
> 
> **`bootstrapLocalTemplate` integration tests** now mock `spinner` from `@sanity/cli-core/ux` via that helper, dynamically import the module under test after mocks are registered, and drop the old `output.spinner` / `print` / `clear` stubs on the fake `Output`. Tests assert spinner usage (e.g. initial message, three `start`/`succeed` cycles per bootstrap).
> 
> `@sanity/cli-test` **tsconfig** gains a `@sanity/cli-core/*` path alias so the new helper can import `SpinnerInstance` from the ux subpath.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 064a258e89a3eddd7529ab8759ba2548b0db908c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli-test/src/index.ts
+++ b/packages/@sanity/cli-test/src/index.ts
@@ -1,5 +1,6 @@
 export * from './test/constants.js'
 export * from './test/createMockHttpServer.js'
+export {createMockSpinner} from './test/createMockSpinner.js'
 export * from './test/createMockWatcher.js'
 export * from './test/createTestClient.js'
 export * from './test/createTestToken.js'
--- a/packages/@sanity/cli-test/src/test/createMockSpinner.ts
+++ b/packages/@sanity/cli-test/src/test/createMockSpinner.ts
@@ -1,10 +1,10 @@
-import type {SpinnerInstance} from '@sanity/cli-core/ux'
-import type {PersistOptions} from 'ora'
+import {type SpinnerInstance} from '@sanity/cli-core/ux'
+import {type PersistOptions} from 'ora'
 import {Mock, vi} from 'vitest'
 
 /**
  * Creates a mock spinner function to avoid writing output to stdout.
- * @param overrides Pass any mocks or stubs for test expectations
+ * @param overrides - Pass any mocks or stubs for test expectations
  * @internal
  */
 export function createMockSpinner(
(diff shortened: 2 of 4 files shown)


runeb (commented):
@snocorp Looks good, just a question. I see this is in integration tests. We do want to catch issues with spinners and for example non-interactive users like agents. Will we still be able to ensure progress indication doesnt break any of the flows we care about?

snocorp:
> @snocorp Looks good, just a question. I see this is in integration tests. We do want to catch issues with spinners and for example non-interactive users like agents. Will we still be able to ensure progress indication doesnt break any of the flows we care about?

Good question. I think this generally makes it easier to verify the spinner is used as expected. I suppose it's possible that a new version of the spinner could break something so we shouldn't mock all the places we use it. In many existing test cases the output would get swallowed by the output capture that happens when you use testCommand and we can assert the output there so that should cover enough to be safe.

filmaj:
Re: a spinner mock in integration tests. Depends on the test. The mock should absolutely be in place for unit tests - third party deps should not interfere with those at all. The definition we have converged on for integration test is something between a unit test and the existing E2E tests, that is: an integration test mocks things (third party dependencies or otherwise) as a judgment call, depending on the test. IMO, if we want to formalize a test for spinner behaviour in non-interactive users, that should belong in E2E tests, as we are basically ensuring a third party dep's functionality. The validation itself, in an automated context like a test, would be messy: it would have to parse non-interactive TTY stdout for, basically, animations, which would look like many lines of text with minor modifications between each line (each line kind of representing a 'frame' of the animation).

My gut feeling on this is it is not worth testing that particular scenario. If there is a specific regression you can point to that involved the spinner, and we can observe that failure mode and make an assessment on if/how to catch it consistently in an automated test, I am all for parsing through it to see what can be done.

filmaj (commented):
This is great and useful - but I am hesitant to put this mock implementation in the cli-test package. If the `spinner` API export of `cli-core/ux` changes, it's not very obvious that the mock in `cli-test` also needs to change. We could put into place guidelines/docs to make that connection between the two packages explicit, but I am not sure that's a good pattern.

IMHO one avenue that in my experience scales better to provide both a module to consume as well as maintained mocks for that module, for use in testing, is to revision the mocks alongside the code it mocks out. The `test/mockSanityCommand.ts` in the `cli` package is in a similar situation: it provides a mock implementation for the `src/SanityCommand.ts` class in the `cli-core` package; it should probably exist in the `cli-core` package, too. That class leverages a lot of complexity hidden by OCLIF abstractions, and so it should not be up to the _consuming_ packages, like `cli` to mock it out: it should be provided by the source package, so that as it evolves, the mock does too, and consumers won't have to change their tests (barring a breaking change).

I think this feels awkward right now because we are mid-way through an API reorg and changing how this repo does testing, to avoid the current reliability and test run time issues we are seeing. The fact that we just stuff "test things" into the cli-test repo is a symptom of the core misunderstanding of how to unit test large projects. I think it demonstrates that testing is a secondary consideration; if we can change that, and make testing a _core_ consideration for how this repo evolves, more sustainable designs will emerge, that are easier to test, and thus should yield higher quality outcomes. Mocks are good, actually, and help to author fast tests. Mocks, therefore, need just as much love as the implementations they serve.

filmaj:
My comments above shouldn't block this PR, but at some point the tedious work of organizing APIs alongside mocks and leveraging them appropriately across the sub-packages needs to be done. Otherwise we will just continue to dig the hole and test reliability and durations will continue to suffer.

filmaj (approved):
Think it's not worth derailing this PR in the short term, but longer-term I will write up a proposal around eliminating barrel exports and how that factors into exporting mocks alongside functional modules.

PR #1423, opened 2026-07-02
feat(workbench): add `unstable_defineMediaLibrary`

gu-stav:
`unstable_defineMediaLibrary` declares the Sanity Media Library as a workbench app — a singleton whose custom `fields` become an installation config.

### Design decisions

- **The config splits in two.** A field's `defineField` value can't serialize, so only its metadata rides the wire (and deploys); the live value loads from the app's federation module. The workbench merges the two halves.
- **Discriminated by `appType`.** `fields` is the media-library payload; the shape is a union, so other config types can be added without reworking the pipeline.
- **Its own dev channel.** In dev the config travels over the dev-server registry + HMR on an `installationConfigs` stream, separate from local applications. A dev server is a local app unless it's config-only (a config with no interfaces); one with interfaces too is both.

Workbench half: sanity-io/workbench#267.

---

> [!NOTE]
> **Medium Risk**
> Touches workbench build, dev registry/HMR, and deploy gating across `@sanity/cli` and `@sanity/workbench-cli`, but behavior is behind an unstable API with heavy test coverage and deploy does not persist installation config yet.
> 
> **Overview**
> Adds **`unstable_defineMediaLibrary`** so authors declare the Sanity Media Library as a branded workbench singleton (`organizationId` + optional `fields`); field `src` modules default-export `defineField` schemas that become an **`installationConfig`** (`appType: 'media-library'`), re-exported from `@sanity/cli`.
> 
> **Installation config** is wired through build, dev, and deploy alongside views/services: federation exposes it and the build emits a generated **`./configs/installation_config`** module aggregating live field values; metadata (`name`, `title`, `public`, `src`) can travel separately from non-serializable schema values. Deploy **`assertDeployable`** treats a config as deployable content; **`buildInstallationConfigDeploymentPayload`** validates the future API payload and logs it only (not persisted yet).
> 
> **Dev** registers `installationConfigs` on the dev-server registry, broadcasts them on a dedicated **`installationConfigs`** channel (config-only servers skip `applications`), and replaces interface-only change tracking with **`exposesSetId`** / **`createExposesTracker`** so installation-config module changes trigger rebuilds/full-reload like new views.
> 
> Includes a **`fixtures/media-library`** example and broad unit/integration tests.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 50a1e8f3cbbe1cf50de5314c2a91ced797c3c9d4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/workbench-cli/src/actions/dev/exposesSetId.ts
gu-stav, line 1: Mostly renamed from `interfaceSetId.ts`, because an installation config isn't an interface but fits into the same model to calculate HMR changes too.

packages/@sanity/workbench-cli/src/actions/dev/exposesSetId.ts
```diff
@@ -0,0 +1,70 @@
...
+ */
+export function exposesSetId({installationConfigs, interfaces}: ExposeSet): string {
+  const keys = [
+    ...(interfaces ?? []).map((iface) =>
+      [iface.interface_type, iface.name, iface.entry_point].join('::'),
+    ),
+    // The config module is generated at build time, so a changed field set (add
+    // or rename) needs a rebuild — key on each field, not just "a config exists".
+    ...(installationConfigs ?? []).flatMap((config) =>
+      config.fields.map((field) => `config::${config.appType}::${field.name}`),
```
mttdnt: Just so I understand this is where we are setting definitions for when reloads occur for `sanity.cli.ts` changes, specifically the `name` field. If the `src` field is changed should that cause a rebuild?

.changeset/unstable-define-media-library.md
```diff
@@ -0,0 +1,10 @@
+---
+'@sanity/workbench-cli': minor
+'@sanity/cli': minor
+'@sanity/cli-core': patch
+---
+
+feat(workbench): add `unstable_defineMediaLibrary`
+
+Declare the Media Library as a workbench app — sugar over `unstable_defineApp`,
+which gains `isSingleton` and an optional `installationConfig`.
```
mttdnt: For changesets we should only mention public facing changes right? Should we drop the `isSingleton` mention?

mttdnt (changes requested):

packages/@sanity/workbench-cli/src/actions/dev/exposesSetId.ts
gu-stav: Fixed in https://github.com/sanity-io/cli/pull/1423/changes/3452251b8b48ca3b66cfd7d00a1d3484d2d1810e. I've also improve this further so that a particular installation config (media-library) doesn't propagate into the util directly, but only their entries in https://github.com/sanity-io/cli/pull/1423/changes/50a1e8f3cbbe1cf50de5314c2a91ced797c3c9d4

.changeset/unstable-define-media-library.md
gu-stav: I've trimmed it down 👍🏼

packages/@sanity/cli/src/actions/deploy/deployApp.ts
```diff
@@ -134,6 +135,18 @@ async function runAppDeployment(options: DeployAppOptions, reporter: CheckReport
...
+  // Endpoint isn't wired yet: validate and log the payload, don't send it.
+  if (workbench?.installationConfig) {
+    const payload = buildInstallationConfigDeploymentPayload({
+      applicationId: application.id,
+      installationConfig: workbench.installationConfig,
+    })
+    output.log('Validated the installation config (not yet persisted):')
+    output.log(JSON.stringify(payload, null, 2))
+    deployDebug('Installation config deployment payload', payload)
+  }
```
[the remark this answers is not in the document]
gu-stav: You're right that dry-run skips it, but I'm leaving it for now.

The block is a stub: the endpoint isn't wired, nothing gets sent. So all dry-run misses is catching a malformed config a bit earlier. And nothing else validates the config either — `unstable_defineApp` is a pure identity wrapper and `DefineAppInputSchema` is never parsed in the build path, so this deploy-time check is the only config validation that exists today. The dry-run gap is a symptom of that, not a standalone bug.

`assertDeployable()`, the real "can this deploy" gate, does run before the dry-run return, so dry-run isn't validation-free.

The clean placement lands when the endpoint gets wired and this becomes a real deploy step. Doing it now means splitting out a config-only validator (the payload needs an `applicationId` that dry-run doesn't resolve) to guard a payload that never leaves the process.

joshuaellis, line 148: this is unclear at the moment, you're using `application.id` to attach a singleton to – but a customer won't deploy the media-library, so how would they get an application id?

To me this is reading as I deploy an app to Brett and then I deploy the configuration, wouldn't it make more sense to have a different routing for configurations that instead look up the installation id for a media library on the organization in question and then deploy the config that way?

joshuaellis (changes requested):

packages/@sanity/cli/src/actions/deploy/deployApp.ts
gu-stav: Deployment is wired up in https://github.com/sanity-io/cli/pull/1441 that hopefully makes it clearer.

A singleton app with an installation config, deploys the config based on app-type, because we only have `type: media-library` at the moment and don't know the appId ahead of time. If we will have multiple instances in the future we will have to extend the `unstable_defineMediaLibrary` interfacewith e.g. an appId or whatever mechanism we want.

1441 works the way you described it: resolve applications, find installation id for the media library and use it.

joshuaellis (approved):

mttdnt (approved):

PR #1426, opened 2026-07-02
fix(dev): align dev server typegen defaults with `sanity typegen generate`

### LINKED ISSUE #1425: Automatic typegen (`enabled: true`) has different defaults
**Describe the bug**

We've been using sanity typegen as a standalone command to generate schema types + query types for our front-end. Recently we moved our typegen config into `sanity.cli.ts` and tried enabling the dev server typegen with `enabled: true`.

Our config in `sanity.cli.ts` is functionally identical to our previous `sanity-typegen.json`:
```ts
  typegen: {
    enabled: true,
    schema: 'schema.json',
    path: '../app/sanity/queries.ts',
    generates: '../app/sanity/schema.types.ts',
  },
```

We can still run the typegen manually, generating identical results:
```sh
pnpm sanity schema extract && pnpm sanity typegen generate
```

Now when running `sanity dev`, it detects schema changes and updates `app/sanity/schema.types.ts`, but formatting is different (creating large git diffs), and the "Query TypeMap" for `@sanity/client` is not present in the file.

We eventually discovered new options that need to be set:

```ts
  typegen: {
    enabled: true,
    schema: 'schema.json',
    path: '../app/sanity/queries.ts',
    generates: '../app/sanity/schema.types.ts',
    // if unset, these options both default to `false` in `sanity dev` and `true` in `sanity typegen generate`
    formatGeneratedCode: true,
    overloadClientMethods: true,
  },
```

The documentation on these options isn't very clear:
- https://www.sanity.io/docs/help/configuring-typegen-in-sanity-cli-config
  - Does not mention `formatGeneratedCode`
  - Description of `overloadClientMethods`:
    - "set to false to disable automatic overloading the sanity client"
    - makes it sound like something that is enabled by default, which would make sense given how `typegen generate` works
    - however, the default is actually already `false`
- https://www.sanity.io/docs/apis-and-sdks/sanity-typegen#u186wq3efxtu
  - Does not mention `formatGeneratedCode`
  - "You can opt out by setting `overloadClientMethods` to `false` in your sanity.cli.ts."

The fact that the dev server and the typegen command produce different results from the same config seems like a clear bug.

The bug seems to have been introduced in this PR sanity-io/sanity#11957 where defaults have been set in multiple places like:

```ts
  const config: TypeGenConfig = {
    formatGeneratedCode: inputConfig.formatGeneratedCode ?? false,
    generates: inputConfig.generates ?? 'sanity.types.ts',
    overloadClientMethods: inputConfig.overloadClientMethods ?? false,
    path: inputConfig.path ?? DEFAULT_QUERY_PATTERNS,
    schema: inputConfig.schema ?? 'schema.json',
  }
```

But these seem to ignore the existing defaults which appear in https://github.com/sanity-io/codegen/blob/c9f3a21ac0e41d917ea80d53139c09abbb99e80e/src/readConfig.ts#L21.

**To Reproduce**

1. Define a simple studio schema
2. Define a query file and add a simple query to it
3. Configure `typegen` in `sanity.cli.ts`, leaving `formatGeneratedCode` and `overloadClientMethods` unset
4. Run `sanity schema extract && sanity typegen generate` to generate types, noting the formatting and typemap generated for `@sanity/client`
5. Run `sanity dev` and make a small change to the studio schema
6. Note the different formatting and lack of typemap in the generated types file

**Expected behavior**

The output from generating types via the dev server should be identical to the output when running `sanity typegen generate` manually.

**Which versions of Sanity are you using?**

```
@sanity/cli (global)             7.4.0 (up to date)
@sanity/orderable-document-list  2.0.4 (up to date)
@sanity/table                    2.0.1 (latest: 3.1.3)
@sanity/ui                       3.2.0 (up to date)
@sanity/vision                   6.2.0 (up to date)
sanity                           6.2.0 (up to date)
@sanity/eslint-config-studio     6.0.0 (up to date)
```

**What operating system are you using?**

linux

**Which versions of Node.js / npm are you running?**

```
pnpm -v && node -v
8.15.4
v22.22.3
```

bjoerge:
### Description

The typegen vite plugin used by `sanity dev`/`sanity build` hand-rolled its config defaults, defaulting `formatGeneratedCode` and `overloadClientMethods` to `false` (and using a narrower set of default query patterns). Meanwhile, `sanity typegen generate` applies defaults through `configDefinition.parse()` from `@sanity/codegen`, where both default to `true`. As a result, the same `sanity.cli.ts` config produced different output depending on whether types were generated by the dev server or by the manual command: the dev server emitted unformatted code (large git diffs) without the `@sanity/client` query typemap.

This PR replaces the hand-rolled defaults with `configDefinition.parse(inputConfig)` — the exact same zod schema the `typegen generate` command uses — so the two code paths share one source of truth and can't drift again. The schema strips the extra `enabled` key from the CLI config, and the plugin's now-redundant local default query patterns were removed.

Fixes https://github.com/sanity-io/cli/issues/1425

### What to review

- `packages/@sanity/cli/src/server/vite/plugin-typegen.ts`: config resolution now goes through `configDefinition.parse()`. Note this intentionally changes the dev-server defaults: `formatGeneratedCode: true`, `overloadClientMethods: true`, `generates: './sanity.types.ts'`, and the broader default query patterns from `@sanity/codegen` (including `.astro`/`.vue`/`.svelte` and `./sanity/**`).
- Explicitly-set config values are still respected (defaults only fill in unset fields).

### Testing

- Updated the plugin unit test that had baked in the old (wrong) defaults to assert the plugin passes `configDefinition.parse({})` verbatim, so it stays in lockstep with `@sanity/codegen`.
- Verified end-to-end in `fixtures/basic-studio` with the issue reporter's config shape (`enabled: true`, format/overload unset):
  - `sanity typegen generate` output vs. dev-server-generated output is now **byte-identical** (prettier-formatted, with the `declare module "@sanity/client"` typemap).
  - Watch-mode regeneration after editing a query file remains identical to manual generation.
  - Explicit opt-out (`formatGeneratedCode: false, overloadClientMethods: false`) still produces unformatted output without the typemap.

---

> [!NOTE]
> **Medium Risk**
> Default generated `sanity.types.ts` output changes for projects that relied on implicit dev-server defaults (formatting and client typemap), though explicit config opt-outs still apply.
> 
> **Overview**
> The Vite typegen plugin used by **`sanity dev`** / **`sanity build`** no longer applies its own default config. It now resolves options with **`configDefinition.parse()`** from **`@sanity/codegen`**, the same path as **`sanity typegen generate`**, so watch/build and manual runs share one source of truth.
> 
> **Behavior change when fields are left unset:** dev-server generation now defaults **`formatGeneratedCode`** and **`overloadClientMethods`** to **`true`**, uses **`@sanity/codegen`**’s output path and query globs (broader than the removed local patterns), and strips CLI-only keys like **`enabled`** via the schema. Explicit **`sanity.cli.ts`** values are unchanged.
> 
> The plugin unit test now expects **`runTypegenGenerate`** to receive **`configDefinition.parse({})`** instead of the old hand-rolled defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 70333a05081f53a4a0c3909f085039786fb369e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1426.md
+++ b/.changeset/pr-1426.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix(dev): align dev server typegen defaults with `sanity typegen generate`
--- a/packages/@sanity/cli/src/server/vite/__tests__/plugin-typegen.test.ts
+++ b/packages/@sanity/cli/src/server/vite/__tests__/plugin-typegen.test.ts
@@ -2,6 +2,7 @@ import path from 'node:path'
 
 import {CLITelemetryStore} from '@sanity/cli-core'
 import {createMockHttpServer, createMockWatcher} from '@sanity/cli-test'
+import {configDefinition} from '@sanity/codegen'
 import {afterEach, beforeEach, describe, expect, it, vi} from 'vitest'
 
 import {sanityTypegenPlugin} from '../plugin-typegen.js'
@@ -272,7 +273,7 @@ describe('sanityTypegenPlugin', () => {
     })
   })
 
-  it('applies default config values', async () => {
+  it('applies the same default config values as `sanity typegen generate`', async () => {
     vi.useRealTimers()
 
     const plugin = sanityTypegenPlugin({
@@ -288,12 +289,7 @@ describe('sanityTypegenPlugin', () => {
     await buildEnd()
 
     expect(runTypegenGenerate).toHaveBeenCalledWith({
-      config: expect.objectContaining({
-        formatGeneratedCode: false,
-        generates: 'sanity.types.ts',
-        overloadClientMethods: false,
-        schema: 'schema.json',
-      }),
+      config: configDefinition.parse({}),
       workDir: TEST_PROJECT_DIR,
     })
   })
--- a/packages/@sanity/cli/src/server/vite/plugin-typegen.ts
+++ b/packages/@sanity/cli/src/server/vite/plugin-typegen.ts
@@ -4,6 +4,7 @@ import path from 'node:path'
 import {CLITelemetryStore} from '@sanity/cli-core'
 import {logSymbols} from '@sanity/cli-core/ux'
 import {
+  configDefinition,
   type GenerationResult,
   runTypegenGenerate,
   type TypeGenConfig,
@@ -17,12 +18,6 @@ import picomatch from 'picomatch'
 import {type Plugin} from 'vite'
 
 import {toForwardSlashes} from '../../util/toForwardSlashes.js'
-
-/**
- * Default glob patterns to watch for query file changes.
- * Covers common source directory naming conventions.
- */
-const DEFAULT_QUERY_PATTERNS = ['./src/**/*.{ts,tsx,js,jsx}', './app/**/*.{ts,tsx,js,jsx}']
 
 /** Default debounce delay in milliseconds */
 const DEFAULT_DEBOUNCE_MS = 1000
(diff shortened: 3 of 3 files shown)


runeb (approved):
Assuming this doesn't break users types, only reduces diffs and adds more core types

bjoerge:
> Assuming this doesn't break users types, only reduces diffs and adds more core types

I don't have a clear picture of the downstream impact of these changes, so not in a position to give a confident answer to that. Would love for someone from @sanity-io/typegen to weigh in/review.

kbrabrand (approved):
Nice. It was never intended to differ, so this seems fine 👍

PR #1431, opened 2026-07-03
fix(deploy): print deployment URL for SDK apps

gu-stav:
### Description

An app deploy never showed where it landed — the dry-run plan and the success line had no URL, so you couldn't see the target or confirm it afterward. Studios already print their `sanity.studio` URL; now apps do too.

A core app has no host of its own (it runs in the dashboard), so both the plan and the success message report its dashboard URL — `…/@<org>/application/<appId>` — from one shared `getCoreAppUrl` helper.

Two smaller fixes rode along: a skipped manifest ("no icon or title") shows as informational (ℹ) instead of a green check, and the dry-run "Files to deploy" header shows a file count.

Stacked on #1416.

### What to review

- `getCoreAppUrl` in `urlUtils.ts`, used by both the dry-run plan (`describeAppTarget`) and the success line (`logAppDeployed`).
- Sanity check: run `sanity deploy --dry-run` and a real deploy on a core app, and compare against a studio.

### Testing

A unit test pins the dashboard URL format; the app/studio deploy unit and integration tests cover the plan message and the success line. Type-check and lint clean.

---

> [!NOTE]
> **Low Risk**
> CLI messaging and TypeScript types only; no deploy upload or API behavior changes.
> 
> **Overview**
> SDK/core app deploys now surface **where the app lives** in the CLI, similar to studios. A shared **`getCoreAppUrl`** helper builds the org-scoped dashboard route (`…/@<org>/application/<appId>` via `getSanityUrl`).
> 
> **Dry-run** app target checks and **post-deploy success** both use that URL: `describeAppTarget` appends it to the “Deploys to existing application …” line, and `logAppDeployed` prints `Success! Application deployed to <url>` (using the app’s `organizationId`, not config).
> 
> Typing is tightened with **`UserApplicationResolved`** for core apps (`organizationId` required) on SDK fetch/list/create/update and through the deploy resolution path.
> 
> Also: skipped manifest extraction (no icon/title) uses an **info** spinner instead of success; dry-run **“Files to deploy”** includes a **file count** (`N file(s), X MB`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 594fcca4124cf4203191aa518b79aafa03ca3924. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/deploy/deployChecks.ts
+++ b/packages/@sanity/cli/src/actions/deploy/deployChecks.ts
@@ -19,6 +19,7 @@ import {
   type StudioDeployTargetResolution,
 } from './resolveDeployTarget.js'
 import {type DeployFlags} from './types.js'
+import {getCoreAppUrl} from './urlUtils.js'
 
 type DeployCheckStatus = 'fail' | 'pass' | 'skip' | 'warn'
 
@@ -220,8 +221,12 @@ export function describeAppTarget(resolution: AppDeployTargetResolution): Deploy
     }
     case 'found': {
       const {application} = resolution
+      const title = application.title ?? application.appHost
+      const dashboardUrl = application.organizationId
+        ? ` at ${getCoreAppUrl(application.organizationId, application.id)}`
+        : ''
       return {
-        message: `Deploys to existing application "${application.title ?? application.appHost}"`,
+        message: `Deploys to existing application "${title}"${dashboardUrl}`,
         status: 'pass',
       }
     }
--- a/packages/@sanity/cli/src/actions/deploy/deployChecks.ts
+++ b/packages/@sanity/cli/src/actions/deploy/deployChecks.ts
@@ -222,11 +222,8 @@ export function describeAppTarget(resolution: AppDeployTargetResolution): Deploy
     case 'found': {
       const {application} = resolution
       const title = application.title ?? application.appHost
-      const dashboardUrl = application.organizationId
-        ? ` at ${getCoreAppUrl(application.organizationId, application.id)}`
-        : ''
       return {
-        message: `Deploys to existing application "${title}"${dashboardUrl}`,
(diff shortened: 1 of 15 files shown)


packages/@sanity/cli/test/integration/commands/deploy.app.test.ts
```diff
@@ -273,6 +274,7 @@ describe('#deploy app', () => {
     if (error) throw error
     expect(stdout).toContain('Dry run — no changes made.')
     expect(stdout).toContain('Deploys to existing application "Existing App"')
+    expect(stdout).toContain(`/@org-id/application/${appId}`)
```
mttdnt: Just like FIl's comment in, https://github.com/sanity-io/cli/pull/1436. Do we want to pull out the integration tests here too?  This seems like a case that should be covered by a unit test.

mttdnt (approved):
Tested out locally and looks good! Just left a question/comment.

packages/@sanity/cli/src/actions/deploy/deployChecks.ts
```diff
@@ -220,8 +221,12 @@ export function describeAppTarget(resolution: AppDeployTargetResolution): Deploy
     }
     case 'found': {
       const {application} = resolution
+      const title = application.title ?? application.appHost
+      const dashboardUrl = application.organizationId
```
ryanbonial: We may want to be consistent with how we get the org ID so that it is the same as the `deployApp` https://github.com/sanity-io/cli/blob/deploy-report-polish/packages/@sanity/cli/src/actions/deploy/deployApp.ts#L267

gu-stav: Fixed in 594fcca4124cf4203191aa518b79aafa03ca3924 - the response types were too broad. A resolved core app always has `organizationId` set, so we are now able to use the shared helper in both places.

PR #1433, opened 2026-07-06
feat(deploy): carry the resolved deployment URL through the dry-run plan

gu-stav:
### Description

The dry-run `--json` plan was built from the checks, and it dropped the passing ones — so the deploy target (the appId and URL the human report names) never reached the JSON. A tool parsing the plan couldn't tell where the deploy would land.

Now the target is computed once and carried on the plan, so the human report and the JSON read the same value — they can't drift.

Stacked on #1431 → #1416.

### What to review

- `DeployTarget` in `deployChecks.ts` — resolved once in `describe*Target`, feeding both the check message and the structured `target`.
- `collectPlan` lifts it onto the plan; `deploymentPlanToJson` passes it straight through.

### Testing

Unit: the target check carries `{applicationId, url}` for studios and apps, and the JSON passes it through. Integration: a `--dry-run --json` plan exposes `target` end-to-end. Type-check and lint clean.

---

> [!NOTE]
> **Low Risk**
> Additive JSON field on dry-run reporting only; deploy execution paths are unchanged aside from sharing the same URL string for human messages.
> 
> **Overview**
> **`sanity deploy --dry-run --json`** now exposes **`target`** (`applicationId`, `url`) on the plan, so tools parsing JSON can see where the deploy would land without scraping pass-check messages (which were omitted from JSON before).
> 
> The URL and app id are resolved **once** in `describeStudioTarget` / `describeAppTarget`, attached on the deploy-target **`DeployCheck`**, lifted onto **`DeploymentPlan`** in `collectPlan`, and passed through **`deploymentPlanToJson`**. For **would-create** studio hostnames, `applicationId` is `null` but `url` is still set; when target resolution fails or is skipped, `target` is `null`.
> 
> Unit and integration tests cover studio/core-app checks and end-to-end `--dry-run --json` output. Minor **`@sanity/cli`** changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6dbc1e64148989d0e9a5e2b5963e89956598b027. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/test/integration/commands/deploy.studio.test.ts
+++ b/packages/@sanity/cli/test/integration/commands/deploy.studio.test.ts
@@ -319,6 +319,11 @@ describe('#deploy studio', () => {
     expect(plan.errors).toEqual({})
     expect(Array.isArray(plan.warnings)).toBe(true)
     expect(plan.files).toContainEqual(expect.objectContaining({path: 'dist/index.html'}))
+    // The target the human report names is carried structured in the JSON too
+    expect(plan.target).toEqual({
+      applicationId: 'studio-app-id',
+      url: 'https://existing-studio.sanity.studio',
+    })
   })
 
   test('keeps stdout pure JSON when --json is given a custom output directory', async () => {
--- a/.changeset/deploy-json-target.md
+++ b/.changeset/deploy-json-target.md
@@ -1,0 +1,5 @@
+---
+"@sanity/cli": minor
+---
+
+feat(deploy): surface the resolved deploy target (appId + URL) in the `--json` dry-run plan, computed once so the human and JSON reports can't drift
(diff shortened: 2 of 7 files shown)


packages/@sanity/cli/test/integration/commands/deploy.studio.test.ts
```diff
@@ -319,6 +319,11 @@ describe('#deploy studio', () => {
     expect(plan.errors).toEqual({})
     expect(Array.isArray(plan.warnings)).toBe(true)
     expect(plan.files).toContainEqual(expect.objectContaining({path: 'dist/index.html'}))
+    // The target the human report names is carried structured in the JSON too
```
mttdnt, line 322: IMO a test that should be covered and already is by a unit test

mttdnt (approved):

gu-stav:
Folded into #1416 — collapsing the deploy `--json` stack into a single PR now that the URL fix (#1431) has landed. All commits from here are in #1416. Closing.

PR #1436, opened 2026-07-06
feat(deploy): add --title flag to name a new application or studio

gu-stav:
### Description

`sanity deploy` couldn't create a coreApp without a TTY — creating one always prompted "Enter a title for your application:", which CI, agents, and piped stdin can't answer. So the first deploy was impossible unattended.

Now the create name comes from either `--title` or `app.title` in `sanity.cli.ts` — independent of `deployment.appId`. `app.title` creates a new app; `deployment.appId` targets an existing one. So `sanity deploy --yes --title "My App"` (or just `app.title` in config) creates and deploys in one shot. For studios, `--title` sets the title on a newly registered hostname (the hostname still comes from `--url`/`studioHost`).

Deploy still writes nothing to config — it stays print-only, so it won't dirty git trees or break reproducible CI.

### What to review

- The unattended guard in `findUserApplication`: `would-create` + a title now proceeds to create instead of erroring.
- Title sourcing: `--title` flag falls back to `app.title` config; studio title flows through the create path only (the coreApp-only update path is untouched — studio title semantics are still open).
- Dry-run parity: both app and studio dry-runs name the title they would set.

### Testing

Unit + integration coverage for create-via-flag, create-via-config, studio create, and the dry-run reports. Full deploy suite green.

---

> [!NOTE]
> **Low Risk**
> CLI deploy UX and API payload fields only; behavior is backward compatible when no title is supplied (still prompts interactively).
> 
> **Overview**
> Adds **`--title`** to `sanity deploy` so CI and **`--yes`** runs can create a **coreApp** or register a **studio** without the interactive “Enter a title for your application” prompt.
> 
> For **apps**, the name is taken from **`--title`** or **`app.title`** in `sanity.cli.ts` (separate from **`deployment.appId`**). **`findUserApplication`** now treats **`would-create`** as proceed-to-create when unattended if a title is present; otherwise dry-run and unattended paths fail with guidance to pass **`--title`** or set **`app.title`**. For **studios**, **`--title`** is sent on create/register API calls when a hostname is configured via **`studioHost`**/**`--url`**.
> 
> Dry-run messaging reflects the title that would be set. Unit and integration tests cover flag/config creation, studio title passthrough, and check parity.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 41b4e0e007892e9393cf37962713c5a68e409cfb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


filmaj (dismissed):
Looks like a solid fix.

We need to stop adding to the integration test suite, however. Landing unit tests for the deploy commands would go a long way to making that easy. That will come with time. I am very open to screensharing/pairing through a unit test writing session, by the way, so keep that in mind !

gu-stav:
Thanks @filmaj - appreciate the feedback. I've converted the tests to unit tests.

filmaj (approved):
YES! awesome! thank you SO much! IF we can employ this sort of attitude to our PRs moving forward (incrementally adding tiny little unit tests to the modules we change), we will be in a much better place.

Based on the changes and your experience in this PR, do you think we could _remove_ any existing integration tests from `test/integration/commands/deploy.*.test.ts`? I know these tests are really big, and have a lot of test cases, but if you skim over the test titles, and something stands out to you as already being covered by any of the existing unit tests, please don't be shy in removing them - especially if the functionality they test aren't super critical (testing e.g. edge case flag usage, or some such).

gu-stav:
I'll take a look tomorrow and see what else can be converted to unit tests and do follow-up PRs!

PR #1439, opened 2026-07-06
feat(refactor): granular cli-core exports

filmaj:
### Description

This PR changes cli-core to set up 'unbarreling' of cli-core's main `exports`. The idea is to deprecate all the import paths from `@sanity/cli-core` and instead provide more granular imports for the sub-set of functionality needed by the consumer. This is particularly impactful for import and transformation workflows relying on cli-core (such as vitest runs), as not all import graphs are needed in all cases. Combined with the fact cli-core depends on `jsdom`, a 46MB dependency (among many others), this can have significant performance implications.

This should be a backwards-compatible change.

Long term, pairing this approach with a set of mocks for each cli-core export, housed in `cli-test`, enables low maintenance and easy unit testing, allowing easier testing of this project and reducing reliance on heavy/slow integration tests. You can see some of this in action in a larger throw-away draft PR I am working on:

- [Mocks of the `cli-core/package-manager` export under `cli-test`](https://github.com/sanity-io/cli/pull/1401/changes#diff-929ba37aef6b673464c46238e78491e754a317aed1fe88c30577f564d238222b)
- [A test that uses these mocks for easy unit testing setup](https://github.com/sanity-io/cli/pull/1401/changes#diff-cca250ca8f5c10582cfc03e6aefb32021c900321f0d1199c8fb460158fd4918b)

Also combines `services/cliToken` with `services/cliUserConfig`, as both rely on each other and are functionally similar. The tests for `cliUserConfig` and `getCliToken`, as a result, are combined.

### What to review

- [x] Pointed out in [`src/_exports/config.ts`](https://github.com/sanity-io/cli/pull/1439/changes#diff-95cec0a66419a1fdced53913c75c43bf5300eb108f38a23ac06a4832495a1b50), why not combine `src/services/cliUserConfig` with `src/config/*`? They seem to do similar things...
- [x] For consistency, and this is also dependent on the answer to the previous questions, that would leave only `src/services/apiClient` in the `src/services` directory. Perhaps we should consider moving it to `src/apiClient` and exposing as a `./apiClient` export?

### Testing

You betcha

---

> [!NOTE]
> **Medium Risk**
> Touches public package export surfaces and moves auth token resolution into `cliUserConfig`, though behavior is intended to stay backward-compatible via the root barrel.
> 
> **Overview**
> **`@sanity/cli-core` now exposes many focused sub-path entry points** (`apiClient`, `config`, `browser`, `debug`, `errors`, `schemas`, `tasks`, `telemetry`, `types`, `util`, `ux`, `SanityCommand`, `ExitCodes`, etc.) so callers can avoid loading the full barrel graph—including **jsdom** via the dedicated **`@sanity/cli-core/browser`** export.
> 
> The root **`@sanity/cli-core`** barrel remains for compatibility but is frozen (explicit re-exports only, with a comment discouraging new additions). Internal modules increasingly import from **`_exports/*`** and **`apiClient`** at **`src/apiClient.ts`** instead of **`services/`**.
> 
> **`getCliToken`** is merged into **`config/cli/cliUserConfig.ts`** ( **`services/getCliToken.ts`** removed); token and user-config tests are combined. Related test and import-path fixes (e.g. **`SanityCommand`** static import, **`apiClient`** mocks) and a large **README** API reference document the new import style.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 647582221760164c6672e8747ca127f37caa198c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli-core/src/config/cli/__tests__/cliUserConfig.test.ts
filmaj, line 1: Nothing new in this test file, just combining and renaming, can be skipped for review.

packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
```diff
@@ -3,12 +3,43 @@ import {dirname, join as joinPath} from 'node:path'
...
 import {getSanityConfigDir} from '../util/getSanityConfigDir.js'
 import {readJsonFileSync} from '../util/readJsonFileSync.js'
 import {writeJsonFileSync} from '../util/writeJsonFileSync.js'
-import {clearCliTokenCache} from './cliTokenCache.js'
+import {clearCliTokenCache, getCachedToken, setCachedToken} from './cliTokenCache.js'
 
+// Re-export so existing consumers don't break
+export {clearCliTokenCache} from './cliTokenCache.js'
+
+// Only used for testing
```
filmaj, line 15: These additions are from the old, now-removed `src/services/getCliToken.ts` file.

stipsan (dismissed):
A very nice cleanup indeed 🥰

filmaj:
@stipsan do you have any opinions on the two naming / combining / exporting questions I posed in the What To Review section of the PR description?

packages/@sanity/cli-core/src/config/cli/getCliConfig.ts
```diff
@@ -1,7 +1,7 @@
 import {createRequire} from 'node:module'
 import {dirname} from 'node:path'
 
-import {debug} from '../../debug.js'
+import {debug} from '../../_exports/debug.js'
```
snocorp, line 4: Why would we want to adjust internal imports as well? Or is this only for certain imports that were re-exports themselves?

snocorp (dismissed):

packages/@sanity/cli-core/src/config/cli/getCliConfig.ts
filmaj: I basically just moved `./src/debug.ts` to `./src/_exports/debug.ts` - its a tiny wrapper around a third party debug logging dependency.

snocorp (approved):

PR #1455, opened 2026-07-08
feat(workbench): deploy core applications as singletons

gu-stav:
### Description

Internal teams need to deploy core applications as singletons (Sanity-owned products like Dashboard). The applications API already accepts an `isSingleton` flag on create, but the CLI never sent it.

An app that sets `isSingleton` via `unstable_defineApp` now deploys like a regular workbench core app and forwards `isSingleton` on the create request. The flag stays hidden — it's not in the public `unstable_defineApp` input type (set the same internal way as `applicationType`/`installationConfig`).

Eligibility (org membership) is enforced by the server, not the CLI — none of it is encoded here. A rejected deploy now surfaces the server's own message cleanly instead of a raw error dump, so the reason reaches the user without leaking internals.

Builds on #1442 (merged).

### What to review

- `deployWorkbenchApp.ts` — `createApplication`/`deployCoreApp` thread `isSingleton` onto the create POST only (Brett fixes singleton-ness at creation, so redeploys don't resend it).
- `deployApp.ts` — forwards `workbench.isSingleton` into the deploy.
- `deployRunner.ts` — `normalizeDeployError` routes through `getErrorMessage`, so an API rejection shows its message, not a dump.

### Testing

Unit: create sends `isSingleton` when set and omits it otherwise; `deployCoreApp` forwards it on create; a rejected deploy surfaces the server message.

---

> [!NOTE]
> **Medium Risk**
> Touches deploy/create paths and singleton semantics for workbench apps; server still enforces eligibility, but wrong flag handling could affect org-scoped application creation.
> 
> **Overview**
> Workbench core-app deploys can now forward an **explicit** `isSingleton` flag from `unstable_defineApp` to `POST /applications` on **create only** (redeploys omit it). The flag is **opt-in**: unset apps no longer default to `false` in resolution, and the multipart field is only sent when the app sets the property.
> 
> Deploy **dry-run**, human checks, **`--json`**, and successful deploy envelopes include `isSingleton` only when explicitly declared; installation-config singleton behavior now keys off a truthy flag rather than a defaulted `false`.
> 
> Failed deploys continue to route errors through `getErrorMessage`, so API rejections (e.g. singleton not allowed) show the server’s `message` instead of a raw HTTP dump.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 26290717f6ce84aae29f16b9f737ad2f443095ce. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/deploy-singleton-core-apps.md
+++ b/.changeset/deploy-singleton-core-apps.md
@@ -1,0 +1,6 @@
+---
+'@sanity/workbench-cli': minor
+'@sanity/cli': minor
+---
+
+feat(workbench): send `isSingleton` on core-app create (`POST /applications`) when set, surface it in the deploy report and `--json`, and relay the API's rejection message on failure.
--- a/packages/@sanity/cli/src/actions/deploy/__tests__/deployRunner.test.ts
+++ b/packages/@sanity/cli/src/actions/deploy/__tests__/deployRunner.test.ts
@@ -159,4 +159,33 @@ describe('runDeploy real deploy', () => {
     expect(output.log).not.toHaveBeenCalled()
     expect(output.error).toHaveBeenCalledWith(expect.stringContaining('boom'), {exit: 1})
   })
+
+  test('surfaces the server message on a rejected deploy, not a raw error dump', async () => {
+    const output = mockOutput()
+    const err = Object.assign(new Error('HTTP 403'), {
+      response: {
+        body: {message: 'You are not allowed to deploy this application as a singleton'},
+        headers: {},
+        method: 'POST',
+        statusCode: 403,
+        statusMessage: null,
+        url: 'https://api.sanity.io/vX/applications',
+      },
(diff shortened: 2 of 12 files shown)


.changeset/deploy-singleton-core-apps.md
```diff
@@ -0,0 +1,6 @@
+---
```
mttdnt, line 1: Because this is an internal change should we not include a changeset here?

mttdnt (approved):
Just double check we need the changeset

.changeset/deploy-singleton-core-apps.md
gu-stav: Without the workbench-cli package does not get a new release and therefore may break releasing @sanity/cli or not include the changes.

PR #1462, opened 2026-07-09
feat(deploy): report whether a deploy created or updated an application

gu-stav:
### Description

A `sanity deploy` with no `deployment.appId` succeeds without prompting — good for humans and agents — but the output never said it had **created a brand-new application**. So re-running the deploy silently created duplicate apps, and there was no machine-readable signal of what happened.

This keeps the no-input behaviour and only **adds** feedback: the success output now states created vs updated, warns (on a no-appId create) that the next deploy makes another app, and the resolved target carries `action: 'create' | 'update'` for `--dry-run`/`--json`. No existing human-facing copy changed.

#### How it worked before (where it broke)

```
sanity deploy  (workbench app, or plain coreApp)
  |
  deployment.appId set?
  |
  +-- yes --> UPDATE that app
  |             "Success! Application deployed to <url>"
  |
  +-- no  --> CREATE a new app  (no prompt)
                "Success! Application deployed to <url>"      <- same line as update
                "Next step: Add the deployment.appId ..."     <- hint, but no reason given
                |
                re-run without appId --> CREATE another new app   <- silent duplicates
```

The success line was identical for create and update, the appId hint never said *why* it mattered, and `--json` carried no create/update field — so neither a human nor an agent could tell a new app had just been made.

#### After this PR (added output, same flow)

```
sanity deploy
  |
  deployment.appId set?
  |
  +-- yes --> UPDATE
  |             "Success! Application deployed to <url>"
  |             "Updated the existing application."
  |
  +-- no  --> CREATE new app  (still no prompt)
                "Success! Application deployed to <url>"
                "Created a new application."
                Next step:
                  "Deploying again without deployment.appId
                   creates another new application."
                  + the appId snippet to pin this app

dry-run / --json (both paths):
  target.action = "create" | "update"
```

### What to review

- `logAppDeployed` in `deployApp.ts` — the added created/updated line and the redeploy warning (only when it created without a configured appId).
- `DeployTarget.action` in `deployChecks.ts` and every target construction (`deployApp.ts`, `deployStudio.ts`, the `describe*Target` verdicts) — derived from whether an existing app was resolved (`update`) or a new one made (`create`). No message copy was changed.

### Testing

`logAppDeployed` unit tests assert both the created (with redeploy warning + appId snippet) and updated output; `describeAppTarget`/`describeStudioTarget` unit tests assert `target.action` for create and update on the app, studio, and workbench paths. Integration deploy tests still pass.

---

> [!NOTE]
> **Low Risk**
> Additive messaging and a new optional JSON field on deploy targets; deploy create/update logic is unchanged aside from propagating a created flag.
> 
> **Overview**
> Adds **create vs update** visibility to `sanity deploy` without changing when apps or studios are created.
> 
> Success output now says **"Created a new application."** or **"Updated the existing application."** After a create with no `deployment.appId`, a **yellow warning** explains that another deploy without `appId` will spawn a duplicate app, alongside the existing `appId` snippet.
> 
> **`DeployTarget`** gains **`action: 'create' | 'update'`** on dry-run checks and on real **`--json`** deploy results (core apps, studios, workbench paths). Studio/app resolution tracks **`created`** via updated return shapes from `findUserApplicationForStudio`, `resolveAppApplication`, and `resolveStudioApplication`.
> 
> Deploy behavior is unchanged; this is additive CLI feedback for humans and automation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5dd9ffc1e7781b0028894ba1b2d43715b5ae9d38. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/deploy/deployApp.ts
+++ b/packages/@sanity/cli/src/actions/deploy/deployApp.ts
@@ -61,6 +61,7 @@ async function runAppDeployment(
   const {cliConfig, flags, output, sourceDir} = options
   const workDir = options.projectRoot.directory
   const organizationId = cliConfig.app?.organizationId
+  const appId = getAppId(cliConfig)
   const workbench = getWorkbench(cliConfig)
   const dryRun = !!flags['dry-run']
 
@@ -112,6 +113,7 @@ async function runAppDeployment(
   checkAppId(reporter, {cliConfig})
 
   let application: UserApplicationResolved | null = null
+  let appCreated = false
   if (flags.external) {
     reporter.report({
       message: EXTERNAL_APP_NOT_SUPPORTED,
@@ -121,13 +123,13 @@ async function runAppDeployment(
   } else if (deployApplication && workbench) {
     // Both modes, so a bad appId fails before the build rather than at the POST.
     await checkAppTarget(reporter, {
-      appId: getAppId(cliConfig),
+      appId,
       isWorkbenchApp: true,
       slug: workbench.slug,
       title: appTitle,
     })
   } else if (deployApplication) {
-    application = await resolveAppApplication(options, {dryRun, reporter})
+    ;({application, created: appCreated} = await resolveAppApplication(options, {dryRun, reporter}))
   }
 
   await checkBuild(reporter, {
@@ -243,13 +245,21 @@ async function runAppDeployment(
       title: appTitle,
       version,
     })
-    logAppDeployed({applicationId, cliConfig, organizationId, output, title: appTitle})
+    logAppDeployed({
+      applicationId,
+      cliConfig,
+      created: !appId,
+      organizationId,
+      output,
+      title: appTitle,
+    })
     return {
       applicationType: 'coreApp',
       applicationVersion: version,
       ...(exposes.length > 0 ? {exposes} : {}),
       ...(workbench.isSingleton === undefined ? {} : {isSingleton: workbench.isSingleton}),
       target: {
+        action: appId ? 'update' : 'create',
         applicationId,
         // A redeploy ignores the slug, so only a create reports the one it used.
         ...(appId ? {} : {slug}),
@@ -268,6 +278,7 @@ async function runAppDeployment(
   logAppDeployed({
     applicationId: application.id,
     cliConfig,
+    created: appCreated,
     organizationId: application.organizationId,
     output,
     title: application.title,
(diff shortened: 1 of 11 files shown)


gu-stav:

- #1462 ◀

packages/@sanity/cli/src/actions/deploy/deployStudio.ts
```diff
@@ -191,7 +201,12 @@ async function runStudioDeployment(
   return {
     applicationType: 'studio',
     applicationVersion: version,
-    target: {applicationId: application.id, title: application.title ?? null, url: location},
+    target: {
+      action: studioCreated ? 'create' : 'update',
+      applicationId: application.id,
+      title: application.title ?? null,
+      url: location,
+    },
```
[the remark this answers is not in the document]
gu-stav: Confirmed and fixed in 140b39c96.

`createFromConfiguredHost` registers a new studio for the `would-create` case but returned it non-null, so `resolveStudioApplication` couldn't tell it apart from a `found` (existing) studio — `studioCreated` stayed false and the deploy JSON reported `action: 'update'` while the dry-run and the API call created it.

`findUserApplicationForStudio` now returns `{application, created}` (`created` is true only when it registers a new host), and `resolveStudioApplication` threads that through instead of hardcoding `false` on the resolved path. Added unit tests for both the create (would-create → registered) and update (found) cases.

packages/@sanity/cli/src/actions/deploy/deployApp.ts
```diff
@@ -240,13 +242,25 @@ async function runAppDeployment(
       title: appTitle,
       version,
     })
-    logAppDeployed({applicationId, cliConfig, organizationId, output, title: appTitle})
+    logAppDeployed({
+      applicationId,
+      cliConfig,
+      created: !appId,
```
mttdnt, line 251: For both deploy app and studio I am seeing many places the `action` signal can come from. In this case either from the appId, `checkAppTarget` or `resolveAppApplication`. If possible, could these be consolidated to a single source?

mttdnt (dismissed):

packages/@sanity/cli/src/actions/deploy/deployApp.ts
gu-stav: I'd like to leave it as is for now - there is a bigger refactoring upcoming for the command, that moves it to a adapter structure, similar to the undeploy command.

joshuaellis (approved):

PR #1465, opened 2026-07-09
chore(refactor): `cli-build` unbarreled imports

filmaj:
### Description

- Have `cli-build` use the granular `cli-core` and `cli-test` exports
- Update unit tests to use the new `cli-test`-provided `cli-core` mocks
- Removed `buildStaticFiles` integration test - replaced with new unit tests
- Added some timing information for global vitest fixture setup so it is obvious how much time we spend on that in CI

### Testing

---

> [!NOTE]
> **Medium Risk**
> Import-path changes touch all build/schema entry points but are mostly mechanical; dropping the heavy `buildStaticFiles` integration test reduces regression signal for vendor/import-map behavior unless the new unit tests fully cover those paths.
> 
> **Overview**
> **`@sanity/cli-build` stops importing the `@sanity/cli-core` barrel** and pulls types, config, package-manager, telemetry, util, tasks, and debug from the package’s subpath exports. **`tsconfig` paths** are extended so tests resolve those entry points (and new `@sanity/cli-test/paths` / mock paths) consistently.
> 
> **Unit tests are realigned** with shared **`@sanity/cli-test/mocks/cli-core/*`** helpers (`createMockOutput`, package-manager/config mocks) instead of partial `importOriginal` mocks on the whole package. **`buildStaticFiles`** gains a non–workbench-app unit test (Vite build, static copy, favicons, chunk stats); **`resolveEntries`** and **`getModuleUrl`** get focused tests; dependency-version tests assert on **`output.error`** instead of simulating **`process.exit`**.
> 
> The **CLI `buildStaticFiles` integration test** (real studio build, import maps, vendor chunks) is **removed**, along with the **`es-module-lexer`** dev dependency used there. **`cli-build` Vitest** enables `tsconfigPaths`, disables console intercept for oclif helpers, and excludes integration tests from the unit project; the **root Vitest project** points at `cli-build/vitest.config.ts`. **Global fixture setup** logs **`console.time`/`timeEnd`** for initialization duration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1661f7e914ec1e96624228b8fa705e6bd075232a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli-build/src/actions/build/__tests__/buildApp.test.ts
```diff
@@ -48,21 +49,29 @@ vi.mock(import('../resolveVendorBuildConfig.js'), () => ({
   resolveVendorBuildConfig: vi.fn(),
 }))
 
+vi.mock(import('../buildDebug'), () => ({
+  buildDebug: vi.fn() as unknown as (typeof import('../buildDebug'))['buildDebug'],
```
snocorp, line 53: Odd that you had to override the type in this case.

packages/@sanity/cli-build/src/actions/build/__tests__/writeSanityRuntime.test.ts
```diff
@@ -1,32 +1,77 @@
...
+
+import {getPossibleDocumentComponentLocations} from '../getPossibleDocumentComponentLocations.js'
+import {resolveEntries, writeSanityRuntime} from '../writeSanityRuntime.js'
 
 const mockWatch = vi.hoisted(() => vi.fn())
 const mockRenderDocument = vi.hoisted(() => vi.fn())
+const mockWriteFile = vi.hoisted(() => vi.fn())
 
+vi.mock(import('node:fs/promises'), async (importOriginal) => {
+  const actual = await importOriginal()
```
snocorp, line 14: What would happen if we mocked the whole module instead? Is it a lot of other mocking?

snocorp (approved):

packages/@sanity/cli-build/src/actions/build/__tests__/buildApp.test.ts
filmaj: When you pass in an `import` return type to `vi.mock` as the first parameter, then `vi.mock` can infer what the mock's shape should look like, so then you do need to type it. Can be helpful, or can be annoying. Sometimes it is easier to just pass a string. Sometimes if you want the test/mock to be kept up to date with module changes, it can be nice to get a type error if your mock drifts out of date. In this case, probably not worth it and should drop the `import` to ignore the types. I'll do that.

packages/@sanity/cli-build/src/actions/build/__tests__/writeSanityRuntime.test.ts
filmaj: I think because the source-under-test imports the default export, we can't do that. I think if we changed that, then it would work. I'll try that!

PR #1486, opened 2026-07-13
refactor: replace tar-fs, tar, and tar-stream with modern-tar

gu-stav:
### Description

`tar-fs`, `tar` (node-tar), and `tar-stream` together carry ~30 transitive packages (~10 MB installed: `streamx`, `bare-fs`, `bare-path`, `minipass`, `minizlib`, ...) for three jobs: pack a directory into a tar stream, extract a tarball to disk, and stream tar entries. [modern-tar](https://github.com/ayuhito/modern-tar) does all three with zero dependencies and ships its own types, so `@types/tar-fs` and `@types/tar-stream` go too.

| | Before (`tar-fs` + `tar` + `tar-stream` + `@types/*`) | After (`modern-tar`) |
|---|---|---|
| Direct dependencies | 5 | 1 |
| Transitive dependencies | 22 | 0 |
| Installed size | ~6.8 MB | 0.08 MB |
| `node_modules` on disk | 10.5 MB | 0.1 MB |
| Cold install | 0.75 s | 0.38 s |
| Warm install | 0.30 s | 0.24 s |

Installed size is measured from the lockfile as the deduplicated union of the removed packages' dependency closures (sizes from disk, `@types/node` excluded since it stays regardless). The install benchmark uses an isolated fixture project containing only these dependencies: pnpm 10.34, median of 3 runs, cold = empty store and cache, warm = populated store, `node_modules` wiped between runs (Apple Silicon; deltas will be diluted inside a full `@sanity/cli` install).

Note: `tar` and `tar-stream` remain in the lockfile as transitive deps of `@sanity/export` and `@sanity/runtime-cli`; they're no longer direct dependencies.

### What to review

The behavioral-parity claims. Both extraction paths were differentially tested against the old implementations on identical fixtures:

- Remote-template extraction: byte-identical trees (content, permissions, symlinks) for both the whole-repo case and the `filePath` subtree case, including `.github` exclusion and prefix-sharing sibling dirs.
- ndjson extraction: identical output for plain ndjson, gzipped ndjson, and tar.gz; identical error for a tarball without ndjson. On plain uncompressed tar the old code crashed with an uncaught AbortError during stream cleanup — the new code extracts correctly (latent bug, path was untested).
- Deploy tarballs keep the `<basename>/...` top-level layout the service expects (verified with bsdtar).

### Testing

Existing unit and integration tests updated to mock `modern-tar/fs` instead of `tar-fs`; deploy (88), extract-ndjson, and init template tests pass.

### Notes for release

Replace `tar-fs`, `tar`, and `tar-stream` with the dependency-free `modern-tar` for deploy, backup, media-import, dataset-import, and template tarball handling


gu-stav:

- #1486 ◀

PR #1062, opened 2026-05-07
refactor: split out build logic into a new package

snocorp:
### Description

The goal of this PR is to move some logic from the main cli package into a new cli-build package. This will eventually allow the runtime-cli package to use the build logic to prepare blueprints assets to be deployed by a blueprints resource provider. I stopped here to help make these changes easier to digest. The next PR will likely involve moving a lot of files from cli to cli-build and that's a lot of noise whereas this PR mostly focuses on the new project setup.

### What to review

Ensure package setup is done correctly and follows standards and conventions for this repo.

### Testing

Testing will be done by using `pkg-pr` to create a build from this PR

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it introduces a new workspace package and rewires CLI build-time imports/packaging (favicons/static asset generation) plus CI/E2E publishing flows, which could impact build output or release packaging if misconfigured.
> 
> **Overview**
> Introduces a new workspace package, `@sanity/cli-build`, to host shared build utilities (initially `copyDir`, favicon + web manifest helpers) with its own build/test/lint configs and `_internal` export surface.
> 
> Refactors `@sanity/cli` build pipeline to consume these utilities via `@sanity/cli-build/_internal` (including replacing `read-package-up` root resolution with `getDefaultFaviconsPath`) and updates related unit tests accordingly.
> 
> Updates repo tooling/CI to recognize and ship the new package: workspace + lockfile, `check:types`/Turbo deps, Knip and Vitest projects/coverage, E2E tarball packing/install, and preview package publishing/docs; also removes `@sanity/cli`’s bundled `static` directory and notes bundle-stats TODO for the new package.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 314e9f6be415fd434fa177be37c3b7430a074585. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1490, opened 2026-07-13
ci: add cli-build to bundle stats, drop dead ignore

gu-stav:
### Description

Two cleanups to the bundle-stats workflow:

- Add `packages/@sanity/cli-build` to the measured packages (the TODO from #1062; the package is published now at 3.0.0). It exposes only `_internal/*` exports, so the `only` filter gains `_internal/**` — with a bare `only: '.'` the action hard-fails for it ("No exports or bin entries matched the active filters"). The pattern doesn't leak into the other packages: `@sanity/cli`'s `./_internal` key has no slash, so `_internal/**` doesn't match it.
- Drop `ignore: 'bin/**'`. It has been a no-op since #858: bin entries are keyed `bin:sanity` and the glob needs a slash, so it never matched anything (which is why `bin:sanity` still shows up in every comment).

### What to review

The bundle-stats comment on this PR: `@sanity/cli-build`'s three `_internal` exports should appear, everything else stays at root exports + bins as before.

### Testing

Ran the action's CLI locally: `--only '.'` against cli-build exits 1 with the "no exports" error, `--only '.' --only '_internal/**'` measures its three exports, and the combined patterns still resolve `@sanity/cli` to just `.` + `bin:sanity`.

### Notes for release

N/A

### DIFF
--- a/.github/workflows/bundle-stats.yml
+++ b/.github/workflows/bundle-stats.yml
@@ -29,7 +29,8 @@ jobs:
 
       - uses: rexxars/bundle-stats@d3671d09a4973b4b2b09c4fe8b1c07c360683397 # v1.11.3
         with:
-          packages: packages/@sanity/cli, packages/@sanity/cli-core, packages/create-sanity
+          packages: packages/@sanity/cli, packages/@sanity/cli-core, packages/@sanity/cli-build, packages/create-sanity
           build-command: pnpm build:cli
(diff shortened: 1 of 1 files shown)


gu-stav:

- #1490 ◀
    - #1491

filmaj (dismissed):

filmaj (approved):

PR #1471, opened 2026-07-10
feat(undeploy): add --json flag

gu-stav:
### Description

Agents and scripts can't consume undeploy output (SDK-1860).

This adds `--json`: the dry-run payload is a projection of the same plan the human report renders, and a real run emits an `{undeployed: true|false}` envelope. Every terminal path writes exactly one JSON document to stdout; progress goes to stderr.

Like deploy, `--json` runs unattended — confirmation prompts are for humans, and machine callers preview with `--dry-run` first.

### What to review

- `undeployPlanToJson` in `runUndeploy.ts` — derived from the same checks and application as the human render.
- The stdout/stderr split and the error envelope.

### Example output

```
$ sanity undeploy --dry-run --json
{
  "application": {
    "activeDeployment": {
      "deployedAt": "2026-07-01T09:12:00Z",
      "deployedBy": "gustav@sanity.io",
      "version": "4.10.2"
    },
    "appHost": "my-studio",
    "createdAt": "2025-11-20T08:00:00Z",
    "deletes": "application",
    "id": "ah1x2y3z",
    "organizationId": null,
    "projectId": "ppsg7ml5",
    "title": "My Studio",
    "type": "studio",
    "url": "https://my-studio.sanity.studio"
  },
  "canUndeploy": true,
  "errors": {},
  "reason": null,
  "warnings": []
}
```

With nothing to undeploy (exit 0):

```
$ sanity undeploy --dry-run --json
{
  "application": null,
  "canUndeploy": false,
  "errors": {},
  "reason": "No studio hostname configured",
  "warnings": []
}
```

A real run emits `{"undeployed": true, "application": …}` on success, `{"undeployed": false, "reason": …}` when there is nothing to undeploy, and `{"undeployed": false, "error": {"message": …}}` on failure.

### Testing

Runner unit tests trace every terminal path of a `--json` run to exactly one JSON document; command tests parse stdout for the dry-run, success, and unattended paths.

### Notes for release

`sanity undeploy` supports `--json` for machine-readable dry-run and result output

---

> [!NOTE]
> **Medium Risk**
> Extends a destructive hosting delete with an unattended JSON path that auto-confirms (mirroring deploy), though behavior is heavily tested and stdout stays parseable.
> 
> **Overview**
> Adds **`--json` / `-j`** to `sanity undeploy` so agents and scripts get machine-readable output aligned with the human dry-run plan and real-run results.
> 
[...]

PR #1494, opened 2026-07-13
feat(cli): share the --json flag definition across commands

gu-stav:
### Description

Follow-up to #1471, where review feedback asked for `--json` to live in the shared flags module. Closes [SDK-1895](https://linear.app/sanity/issue/SDK-1895/make-json-a-shared-flag-across-cli-commands).

8 commands defined their own `--json` flag inline, with 7 different description strings, and only `deploy`, `undeploy`, and `doctor` accepted `-j`. This adds `getJsonFlag()` to `sharedFlags.ts` (locked: name, `-j`, description, `default: false`) and points all 8 commands at it.

Two small user-visible side effects:

- `openapi list`, `projects create`, `schemas list`, `tokens add`, and `tokens list` now accept `-j`.
- Help text is unified to "Output the result as JSON".

I looked at doing the same for `--dry-run` but left it inline: only 3 commands define it, and the description is where the per-command meaning lives (`deploy` vs `undeploy` vs `graphql deploy`), so a shared getter would only lock the name.

Why a getter instead of oclif's built-ins: `Flags.custom()` only produces option flags (it hard-codes `type: 'option'`), so it can't declare a shared boolean. And `enableJsonFlag` is a semantics package, not just a flag — hard-coded description/help group, no `-j`, suppresses `this.log()` and JSON-formats errors when `--json` is passed. Adopting it would be a real behavior change, out of scope here.

### What to review

That locking char/description/default in `getJsonFlag()` is right for every consumer — none of the 8 commands seemed to need a per-command description.

### Testing

Added a `getJsonFlag` case to `sharedFlags.test.ts`. Existing command tests already exercise `--json` behavior.

### Notes for release

All JSON-capable commands (`deploy`, `undeploy`, `doctor`, `openapi list`, `projects create`, `schemas list`, `tokens add`, `tokens list`) now accept `-j` as shorthand for `--json`, with unified help text.

### DIFF
--- a/.changeset/pr-1494.md
+++ b/.changeset/pr-1494.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+feat(cli): share the --json flag definition across commands
--- a/packages/@sanity/cli/src/commands/deploy.ts
+++ b/packages/@sanity/cli/src/commands/deploy.ts
@@ -9,6 +9,7 @@ import {deployDebug} from '../actions/deploy/deployDebug.js'
 import {deployStudio} from '../actions/deploy/deployStudio.js'
 import {determineIsApp} from '../util/determineIsApp.js'
 import {dirIsEmptyOrNonExistent} from '../util/dirIsEmptyOrNonExistent.js'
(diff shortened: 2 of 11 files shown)


gu-stav:

- #1494 ◀

PR #1495, opened 2026-07-13
fix(project): Ensure no dataset listing/create operations during `project create` in unattended mode with no dataset specified

filmaj:
### Description

The logic of `handleDatasetCreation` is weird. Guard against entering it in the case where we don't want to create a dataset at all.

---

> [!NOTE]
> **Low Risk**
> Narrow CLI behavior change for unattended project creation; reduces unintended API calls without altering interactive flows or explicit `--dataset` usage.
> 
> **Overview**
> **`sanity project create` in unattended mode without `--dataset` no longer touches datasets.**
> 
> After a project is created, `handleDatasetCreation` runs only when a dataset name was passed on the CLI **or** the command is in attended mode (so prompts can run). Unattended runs with no `--dataset` skip that path entirely, so **`listDatasets` and `createDataset` are not invoked** and success output omits dataset details.
> 
> Tests for unattended mode were tightened to assert those mocks stay unused when no dataset flag is provided.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 37dce86a158ab0ece9995c1c582d0580a9988a30. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1495.md
+++ b/.changeset/pr-1495.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': patch
+---
+
+fix(project): Ensure no dataset listing/create operations during `project create` in unattended mode with no dataset specified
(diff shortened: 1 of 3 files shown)


snocorp (approved):
Makes sense to me

PR #1497, opened 2026-07-13
fix(cli): support unattended mode in dataset lifecycle commands

jonahsnider:
### Description

Adds unattended-mode support for dataset lifecycle commands.

### What to review

### Testing

---

> [!NOTE]
> **Medium Risk**
> Touches destructive flows (delete, copy, export/import) and backup download overwrite behavior; unattended mode makes misuse easier in automation but adds explicit guards and `--force` for delete.
> 
> **Overview**
> Adds **unattended (non-interactive) behavior** across dataset and backup CLI commands so CI/scripts can run without prompts.
> 
> When `isUnattended()` is true, commands **fail fast with `USAGE_ERROR`** (and clearer messages) instead of calling `select`/`input`/`confirm`—e.g. required `<dataset>`, `--backup-id`, `--dataset`, source/target for copy, dataset name for create, and **`--force` for delete**. **Export** and **backup download** pick **deterministic default output paths** under `cwd` when `--out` is omitted; existing output files require **`--overwrite`** rather than a confirm prompt. Interactive cancel on backup overwrite now uses **`USER_ABORT`**.
> 
> Validation and usage mistakes increasingly exit with **`exitCodes.USAGE_ERROR`** instead of generic `exit: 1`. **Delete** refreshes the type-to-confirm prompt copy; user cancel during confirm no longer forces exit code 1.
> 
> Tests are extended with `isInteractive: false` / `isUnattended` mocks for the new paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b4cdfd5c673f2bfb4eb31f57d35a79aa2cecce34. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/commands/backups/disable.ts
+++ b/packages/@sanity/cli/src/commands/backups/disable.ts
@@ -73,6 +73,10 @@ export class DisableBackupCommand extends SanityCommand<typeof DisableBackupComm
 
     if (dataset) {
       assertDatasetExists(datasets, dataset, this.output)
+    } else if (this.isUnattended()) {
+      this.error('Dataset is required in unattended mode. Pass the dataset name as an argument.', {
+        exit: 2,
+      })
     } else {
       dataset = await this.promptForDataset(datasets)
     }
--- a/packages/@sanity/cli/src/commands/backups/disable.ts
+++ b/packages/@sanity/cli/src/commands/backups/disable.ts
@@ -1,7 +1,7 @@
 import {styleText} from 'node:util'
 
 import {Args} from '@oclif/core'
-import {SanityCommand, subdebug} from '@sanity/cli-core'
+import {exitCodes, SanityCommand, subdebug} from '@sanity/cli-core'
 import {select} from '@sanity/cli-core/ux'
 import {type DatasetsResponse} from '@sanity/client'
 
@@ -74,8 +74,8 @@ export class DisableBackupCommand extends SanityCommand<typeof DisableBackupComm
     if (dataset) {
       assertDatasetExists(datasets, dataset, this.output)
     } else if (this.isUnattended()) {
-      this.error('Dataset is required in unattended mode. Pass the dataset name as an argument.', {
-        exit: 2,
+      this.error('Dataset is required in unattended mode. Pass it as the `<dataset>` argument.', {
+        exit: exitCodes.USAGE_ERROR,
       })
     } else {
       dataset = await this.promptForDataset(datasets)
--- a/packages/@sanity/cli/src/commands/backups/download.ts
+++ b/packages/@sanity/cli/src/commands/backups/download.ts
@@ -224,10 +224,7 @@ ${styleText('bold', 'backupId')}: ${styleText('cyan', opts.backupId)}`,
     docOutStream.end()
     await finished(docOutStream)
 
-    progressSpinner.set({
-      step: `Archiving files into a tarball...`,
-      update: true,
-    })
+    progressSpinner.set({step: `Archiving files into a tarball...`, update: true})
     try {
       await archiveDir(tmpOutDir, outFilePath, (processedBytes: number) => {
         progressSpinner.update({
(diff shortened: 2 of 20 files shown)


packages/@sanity/cli/src/commands/backups/disable.ts
```diff
@@ -73,6 +73,10 @@ export class DisableBackupCommand extends SanityCommand<typeof DisableBackupComm
 
     if (dataset) {
       assertDatasetExists(datasets, dataset, this.output)
+    } else if (this.isUnattended()) {
+      this.error('Dataset is required in unattended mode. Pass the dataset name as an argument.', {
```
shapirodaniel: nit, elsewhere we've used a `<dataset>` argument convention, might be worth standardizing?

packages/@sanity/cli/src/commands/backups/download.ts
```diff
@@ -293,14 +314,17 @@ ${styleText('bold', 'backupId')}: ${styleText('cyan', opts.backupId)}`,
     )
     // If the file already exists, ask for confirmation if it should be overwritten.
     if (!this.flags.overwrite && exists) {
+      if (this.isUnattended()) {
+        this.error(`File "${out}" already exists. Pass --overwrite to replace it.`, {exit: 2})
```
shapirodaniel: should we include messaging about this being a destructive action? it's implicit, but it would possibly help the agent pause and surface the action to a human or supervising agent before pressing ahead

packages/@sanity/cli/src/commands/backups/download.ts
```diff
@@ -293,14 +314,17 @@ ${styleText('bold', 'backupId')}: ${styleText('cyan', opts.backupId)}`,
...
+      }
       const shouldOverwrite = await confirm({
         default: false,
         message: `File "${out}" already exists, would you like to overwrite it?`,
       })
 
       // If the user does not want to overwrite the file, cancel the operation.
       if (!shouldOverwrite) {
-        this.error('Operation cancelled.', {exit: 1})
+        this.error('Operation cancelled.', {exit: 3})
```
shapirodaniel: curious if we're using `exit 3` as our cancellation code here and elsewhere?

packages/@sanity/cli/src/commands/datasets/delete.ts
```diff
@@ -81,20 +84,16 @@ export class DeleteDatasetCommand extends SanityCommand<typeof DeleteDatasetComm
...
-        deleteDatasetDebug(`User cancelled`, err)
-        this.error(`User cancelled`, {exit: 1})
-      }
+      await input({
+        message: `Delete dataset "${datasetName}"?\n  Type the dataset name to confirm:`,
+        validate: (input) => {
+          const trimmed = input.trim()
+          return (
+            trimmed === datasetName ||
+            `Dataset name doesn't match. Enter "${datasetName}" or press Ctrl+C to cancel.`
```
shapirodaniel, line 95: should we have an unattended way of expressing `press Ctrl+C`, ie. bailing on a command, or is this path strictly attended-mode only?

packages/@sanity/cli/src/commands/datasets/import.ts
```diff
@@ -201,10 +201,7 @@ export class ImportDatasetCommand extends SanityCommand<typeof ImportDatasetComm
     let dataset = datasetFlag ?? targetDatasetArg
     if (!dataset) {
       if (this.isUnattended()) {
-        return this.output.error(
-          'Missing dataset. Use the --dataset flag to specify a dataset: --dataset <name>',
-          {exit: 1},
-        )
+        return this.output.error('Dataset is required. Pass it with --dataset <name>.', {exit: 2})
```
shapirodaniel: would be great to standardize on this convention in the other errors above that don't currently use `--arg <name>`

shapirodaniel (dismissed):
a few non-blocking questions/suggestions, otherwise looks great

packages/@sanity/cli/src/commands/backups/download.ts
jonahsnider: yeah USER_ABORT is exit code 3, this should probs be using the constant though

packages/@sanity/cli/src/commands/datasets/delete.ts
jonahsnider: unattended mode bails with an error earlier on, so this only happens when someone is interactive

packages/@sanity/cli/src/commands/datasets/import.ts
jonahsnider: good callout, will fix

packages/@sanity/cli/src/commands/backups/download.ts
jonahsnider: IMO no, the error message already communicates that it's replacing the file. it's a local only change too which makes me less scared of misuse

shapirodaniel (approved):
🚢

PR #1499, opened 2026-07-13
feat(cli): support unattended mode in document commands

jonahsnider:
### Description

Adds unattended-mode support for document commands.

### What to review

### Testing

---

> [!NOTE]
> **Medium Risk**
> Changes affect CI/automation exit codes and when `documents create` may open an editor; validate cancellation is now USER_ABORT instead of a generic error, which scripts may need to handle.
> 
> **Overview**
> Document CLI commands now behave correctly in **unattended mode** (non-TTY / `--yes`), with clearer errors and standardized exit codes.
> 
> **`documents create`** requires a `<file>` argument when unattended instead of opening `$EDITOR`. Flag conflicts (`--replace`/`--missing`, `--id` + file) and missing dataset are rejected earlier with updated copy and `exitCodes.USAGE_ERROR`.
> 
> **`documents validate`** uses `this.isUnattended()` (not only `--yes`) to skip the warning/confirmation flow. Declining confirmation logs *Validation cancelled* and exits with `USER_ABORT` instead of a usage-style error. `--format` is validated via oclif `options`; `--file` gets existence and “is a file” checks before any prompt.
> 
> **`documents get`**, **`delete`**, **`query`**, and **`docs read`** align missing-dataset and invalid-input failures with `USAGE_ERROR` and refreshed messages. Tests and integration coverage were expanded for unattended create/validate, cancellation, and bad `--file`/`--format` paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1149ebc918cd6d47ac49a7d787b6d2d1ee2436e4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/commands/documents/create.ts
```diff
@@ -94,14 +94,34 @@ export class CreateDocumentCommand extends SanityCommand<typeof CreateDocumentCo
     const {file} = args
     const {dataset, id, json5: useJson5, missing, replace, watch} = flags
 
+    if (!file && this.isUnattended()) {
+      this.error(
+        'Document input is required in unattended mode. Pass a JSON file: sanity documents create <file>',
```
shapirodaniel: is there a specific flag that we'd need to pass the json file here, and could we pass a json string?

packages/@sanity/cli/src/commands/documents/validate.ts
```diff
@@ -113,6 +115,30 @@ export class ValidateDocumentsCommand extends SanityCommand<typeof ValidateDocum
...
+    let ndjsonFilePath
+    if (file) {
+      const filePath = path.resolve(workDir, file)
+
+      let fileStat
+      try {
+        fileStat = await stat(filePath)
+      } catch {
+        return this.output.error(
+          `File not found: ${file}. Pass an existing .ndjson file or tarball to --file.`,
```
shapirodaniel: is it worth specifying acceptable extensions here, or is any tarball fine (compressed or otherwise)?

packages/@sanity/cli/test/integration/commands/documents/validate.test.ts
```diff
@@ -47,16 +47,41 @@ describe('#documents:validate', {timeout: 60 * 1000}, () => {
...
+    test('exits if format is incorrect value', async () => {
+      const {error} = await testCommand(ValidateDocumentsCommand, [
+        '--yes',
+        '--file',
+        VALID_DOCS_PATH,
+        '--format',
+        'xml',
+      ])
+
+      expect(error?.message).toContain('Expected --format=xml to be one of: json, ndjson, pretty')
```
shapirodaniel, line 60: curious why is `pretty` valid here?

shapirodaniel (dismissed):
a few minor qs otherwise looks great

packages/@sanity/cli/src/commands/documents/create.ts
jonahsnider: `sanity documents create` takes a `<file>` argument, so you can just point it at the file. inlining it would maybe be possible but would be separate work to add that so i'm gonna skip for now

packages/@sanity/cli/src/commands/documents/validate.ts
jonahsnider: it detects the file based on the content, so technically any file extension works. which maybe means we should change `.ndjson file` to `NDJSON file` to avoid confusion?

packages/@sanity/cli/test/integration/commands/documents/validate.test.ts
jonahsnider: pretty format is the default, it's an OCLIF thing mostly

shapirodaniel (approved):
let's go! 🚀

PR #1500, opened 2026-07-13
feat(cli): support unattended mode for GraphQL commands

jonahsnider:
### Description

Adds unattended-mode support for GraphQL commands.

### What to review

### Testing

---

> [!NOTE]
> **Medium Risk**
> Changes confirmation and exit-code behavior for deploy, undeploy, and schema delete/extract—scripts relying on old exit codes or silent skips may need updates, though defaults in interactive use are largely preserved.
> 
> **Overview**
> Adds **unattended-mode** behavior across GraphQL deploy/undeploy and experimental schema commands so CI and scripted runs do not hang on prompts.
> 
> **GraphQL:** Deploy and undeploy now use `isUnattended()` instead of raw `isInteractive()`. In unattended mode, destructive or ambiguous steps require **`--force`** (generation mismatch, breaking changes, multi-API flag overrides, undeploy). User declines and cancellations exit with **`USER_ABORT`** and clearer messages instead of generic “Operation cancelled”. Missing dataset / usage mistakes use **`USAGE_ERROR`**.
> 
> **Schema:** **`schema delete`** gains **`--yes`** and blocks unattended deletes without it, with an interactive confirm otherwise. **`schema extract`** gains **`--force`** to overwrite an existing output file; unattended runs error without it. **`schema list --json`** routes fetch/permission diagnostics to **stderr** (`warn`) so **stdout** stays valid JSON.
> 
> Tests are aligned with explicit `isInteractive` mocks and the new exit codes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c2bd3f1a6b9275135a92fdc865eeb17d1b749cfc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
+++ b/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
@@ -1,5 +1,5 @@
 import {CLIError} from '@oclif/core/errors'
-import {type Output} from '@sanity/cli-core'
+import {exitCodes, type Output} from '@sanity/cli-core'
 import {confirm} from '@sanity/cli-core/ux'
 import {oneline} from 'oneline'
 
@@ -43,9 +43,9 @@ export async function resolveApiGeneration({
       throw new CLIError(
         oneline`
         Specified generation (${specifiedGeneration}) for API at index ${index} differs from the one currently deployed (${currentGeneration}).
-        Pass --force to continue.
+        Pass \`--force\` to continue.
       `,
-        {exit: 2},
+        {exit: exitCodes.USAGE_ERROR},
       )
     }
 
--- a/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
+++ b/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
@@ -1,5 +1,5 @@
 import {CLIError} from '@oclif/core/errors'
-import {exitCodes, type Output} from '@sanity/cli-core'
+import {type Output} from '@sanity/cli-core'
 import {confirm} from '@sanity/cli-core/ux'
 import {oneline} from 'oneline'
 
@@ -61,7 +61,7 @@ export async function resolveApiGeneration({
       }))
 
     if (!confirmDeploy) {
-      throw new CLIError('GraphQL deployment cancelled', {exit: exitCodes.USER_ABORT})
+      return undefined
     }
 
     return specifiedGeneration
--- a/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
+++ b/packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
@@ -1,5 +1,5 @@
 import {CLIError} from '@oclif/core/errors'
-import {isInteractive, type Output} from '@sanity/cli-core'
+import {type Output} from '@sanity/cli-core'
 import {confirm} from '@sanity/cli-core/ux'
 import {oneline} from 'oneline'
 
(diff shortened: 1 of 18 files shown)


packages/@sanity/cli/src/actions/graphql/resolveApiGeneration.ts
```diff
@@ -58,7 +60,11 @@ export async function resolveApiGeneration({
         message: 'Are you sure you want to deploy?',
       }))
 
-    return confirmDeploy ? specifiedGeneration : undefined
```
shapirodaniel, line 61: nit, there's a lot of ternary statements in this codebase, interesting that the agent decided to refactor this

packages/@sanity/cli/src/commands/graphql/undeploy.ts
```diff
@@ -121,32 +121,32 @@ export class Undeploy extends SanityCommand<typeof Undeploy> {
...
-          exit: 1,
+          exit: 2,
         },
       )
     }
 
     // Confirm deletion unless --force is used
     if (!force) {
+      if (this.isUnattended()) {
+        this.error('GraphQL API undeploy requires confirmation. Pass --force to continue.', {
```
shapirodaniel: curious if there's an acceptable subset of commands that we wouldn't want to expose in unattended mode? seeing `--force` surfaced as an escape hatch for a destructive operation makes sense in human mode, i'm wondering about whether we expect the agent harness (and human who owns it) to be responsible for deciding when to take an action like this

packages/@sanity/cli/src/commands/schemas/delete.ts
```diff
@@ -52,18 +53,44 @@ export class DeleteSchemaCommand extends SanityCommand<typeof DeleteSchemaComman
...
-    const {dataset} = flags
+    const {dataset, yes} = flags
 
     deleteSchemaDebug('Running schema delete with flags: %O', flags)
 
     const ids = parseIds(flags.ids)
 
+    if (!yes && this.isUnattended()) {
+      this.error(
+        'Schema deletion requires confirmation. Re-run with --yes to delete the schemas.',
```
shapirodaniel: in re my earlier comment on unattended destructive actions, this messaging is great, might be worth porting it to other destructive op sites

packages/@sanity/cli/src/commands/schemas/extract.ts
```diff
@@ -68,6 +74,29 @@ export class ExtractSchemaCommand extends SanityCommand<typeof ExtractSchemaComm
...
+      }
+
+      const shouldOverwrite = await confirm({
+        default: false,
+        message: `Schema file already exists at "${extractOptions.outputPath}". Overwrite it?`,
+      })
+
+      if (!shouldOverwrite) {
+        this.output.log('Schema extraction cancelled')
+        return this.exit(exitCodes.USER_ABORT)
```
shapirodaniel, line 96: super nit / fast follow candidate, if the `exitCodes` set includes semantics around `1 | 2 | 3`, might be worth a pass to refactor all the places we currently supply a bare integer

shapirodaniel (dismissed):
a few high level qs, otherwise 🚢

packages/@sanity/cli/src/commands/schemas/extract.ts
jonahsnider: yeah i don't really know why there's a mix of constants and magic numbers. i can't imagine there's a good reason. i'll standardize on the constants in my changes and come back later to update the rest

packages/@sanity/cli/src/commands/graphql/undeploy.ts
jonahsnider: IMO not really, all the destructive commands have you provide `--force`/`--yes`. i think from a liability perspective we have done our due diligence there: we shouldn't make it possible for an agent to accidentally break something in prod for you, but there are legit use cases where a nontechnical person is using AI 100% of the time when working with sanity. we don't want to force those users to become a proxy for their agent where it's just going "run this command and paste the output here" repeatedly

shapirodaniel (approved):
looks great!

PR #1508, opened 2026-07-14
fix(build): stop baking the staging flag into workbench chunks

gu-stav:
### Description

Workbench chunks deploy as federation remotes and load into host pages from any environment, but the build compile-time-defines `__SANITY_STAGING__` — baking the build machine's env into the chunks. The hosted workbench remote is built without `SANITY_INTERNAL_ENV`, so it's stuck on production and staging dev needs `SANITY_INTERNAL_WORKBENCH_REMOTE_URL` as a workaround.

Drop the define for workbench app builds so chunks read `globalThis.__SANITY_STAGING__` at runtime — the host page already injects it, and `getSanityEnv()`'s `typeof` guard treats an absent global as production. Studio and plain app builds keep the define.

Needs a workbench remote redeploy to take effect. Fixes [SDK-1901](https://linear.app/sanity/issue/SDK-1901/workbench-remote-is-env-blind-staging-dev-requires-sanity-internal).

### What to review

The `define` block in `getViteConfig.ts`.

### Testing

Unit test: workbench builds don't define `__SANITY_STAGING__` even with `SANITY_INTERNAL_ENV=staging`.

### DIFF
--- a/.changeset/pr-1508.md
+++ b/.changeset/pr-1508.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-build': patch
+---
+
+fix(build): stop baking the staging flag into workbench chunks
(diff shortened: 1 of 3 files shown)


gu-stav:

- #1508 ◀

gu-stav:
Superseded — fixing this in workbench's env.ts instead (globalThis read survives any define), so the CLI needs no special case. See SDK-1901.

PR #1517, opened 2026-07-15
feat(workbench): emit standalone SPA for workbench remotes

joshuaellis:
### Description

A workbench `sanity build` emits only the federation remote, so it can't be served standalone. With `SANITY_INTERNAL_IS_WORKBENCH_REMOTE=true`, also emit an SPA (`index.html` + bootstrap + favicons) into the same `dist`. Registers a `client` Vite env built alongside `federation`. No bridge script. Stacked on the cli-build PR. [SDK-1038](https://linear.app/sanity/issue/SDK-1038)

### What to review

`emptyOutDir:false` on both envs so they don't wipe each other (relies on the CLI's up-front `rm(dist)`). SPA bundles the app independently of the federation chunks, so `dist` grows by ~app+React.

---

> [!NOTE]
> **Medium Risk**
> Changes the Vite build pipeline and increases `dist` size for flagged remotes, but behavior is gated on an internal env var and deploy checks are unchanged.
> 
> **Overview**
> When **`SANITY_INTERNAL_IS_WORKBENCH_REMOTE`** is set and the remote exposes **`./App`**, workbench **`sanity build`** now also emits a **standalone SPA** (`index.html`, bootstrap from `.sanity/runtime/app.js`, static assets) into the same **`dist`** as the module-federation remote.
> 
> The Vite **`sanityEnvironmentPlugin`** gains optional **`clientInput`**: it registers a **`client`** build environment, runs it before **`federation`**, and sets **`emptyOutDir: false`** on both so outputs coexist (with **`dist`** still cleared once up front by the CLI). Dock-only apps and builds without the flag keep federation-only behavior. Federation input config moves from **`rollupOptions`** to **`rolldownOptions`**.
> 
> Deploy validation in **`checkBuiltOutput`** is documented only — it still requires **`mf-manifest.json`**, not **`index.html`**. Vitest covers **`clientInput`** gating and dual-environment build order.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit defadff98cd6dffbab4955e0d0f9eaa66d365383. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/workbench-cli/src/actions/build/vite/plugins/plugin-sanity-environment.ts
```diff
@@ -4,23 +4,53 @@ import {FEDERATION_DIR_NAME} from '../constants.js'
...
+                },
+              }
+            : {}),
           [FEDERATION_DIR_NAME]: {
             build: {
               copyPublicDir: false,
+              emptyOutDir: false,
               outDir: `dist`,
-              rollupOptions: {input: options.input},
+              rollupOptions: {input},
```
gu-stav: Must become `rolldownOptions` - otherwise we get deprecation warnings

joshuaellis: @claude pls fix

gu-stav (approved):

PR #1534, opened 2026-07-15
feat(cli-core): migrate HTTP layer to get-it v9 and @sanity/client v8

stipsan:
### Description

Upgrade `@sanity/cli-core` from get-it v8 to v9 (now 9.4.1), migrate the request layer to the new fetch-based API, and upgrade `@sanity/client` to v8 (which itself runs on get-it v9 + obug) — removing get-it v8 from the CLI's dependency tree entirely.

`@sanity/cli-core/request` now uses get-it v9-style options (`headers`, `httpErrors`, `debug`, `middleware`, structured `timeout`, etc.) on `createRequester`, and re-exports the v9 helpers used by CLI consumers (`HttpError`, `TimeoutError`, response types, `retry` / `debug`, `nodeReadableFromWeb()`).

Call sites in `@sanity/cli` and `@sanity/cli-build` were updated for the v9 response shape (`status` / `Headers`), `as: 'stream' | 'text'`, and `redirect: 'manual'` — including the `sanity api` and `mintProject` services that landed on `main` against the v8 API.

Debug logging now uses `obug` (a typed, ESM-native, dependency-free fork of `debug`) across cli-core, cli, and cli-build, replacing the `debug` + `@types/debug` pair. `DEBUG=sanity:*` behavior is unchanged, and `getViteConfig`'s user-config logs now go through `buildDebug` instead of calling the factory as a no-op.

### @sanity/client v8

Client v8 removes the `requester` config option, so the CLI adapts:

- The workspace-wide `resolutions` pin to `@sanity/client ^7.25.0` is removed; the catalog now points at `^8.0.0`. The `sanity` / `@sanity/import` subtree keeps its own v7 (they don't support v8 yet), so the two majors coexist; clients passed across that boundary (document validation, `uploadSchema`) only use the stable public surface and are cast at the boundary.
- `createIsolatedRequester` (the get-it v8 deep-require through the client) is deleted. Execution-context isolation is now a `resolveFetch` transport: `CliExecutionContext` gained an optional `fetch` (threaded through `invokeSanityCli`) so embedding hosts can supply their own transport; either way the CLI strips the host's `x-sanity-lineage` header before dispatch. **Note:** unlike the v7 isolated requester, the client's `sanity:client` obug logging cannot be detached per-client in v8, so a host `DEBUG` setting still reaches it — flagged as a known gap.
- The 401 `authErrors` requester middleware became `enrichAuthError()`. To preserve v7's in-flight mutation semantics (callers that wrap `err.message` keep the hint), clients returned by `getGlobalCliClient` / `getProjectCliClient` are wrapped in a proxy that enriches rejected API promises before any caller observes them — covering sub-clients and `withConfig()` derivations. `SanityCommand.catch` and `getErrorMessage` apply the same (idempotent) enrichment as a funnel for anything the proxy can't see (observables, detached `patch()`/`transaction()` builders, unwrapped clients). The members link derives the project ID from the request URL.
- All `client.request({uri})` call sites renamed to `url` (v8 deprecation warning otherwise).
- `eventsource` / `eventsource-parser` (client v8 deps published < 3 days ago) added to `minimumReleaseAgeExclude`, alongside the existing `get-it` entry.

### What to review

- `packages/@sanity/cli-core/src/apiClient.ts` — `resolveFetch`-based execution-context isolation and the auth-hint proxy replacing the custom requester
- `packages/@sanity/cli-core/src/errors/enrichAuthError.ts` — 401 hint enrichment (idempotent, URL-derived project ID)
- `packages/@sanity/cli-core/src/request/createRequester.ts` — `CreateRequesterOptions` derives from get-it's `RequesterOptions`; User-Agent overrides detected case-insensitively
- Backup downloads use structured `timeout: {headers: 15s, total: false}` (fresh per-retry headers deadline, no custom middleware) plus the 3-minute idle-read timer; destinations are deferred until a response arrives
- `services/api.ts` and `services/mintProject.ts` migrated to v9 (`httpErrors: false`, `as: 'text'` / `.text()`, `Headers` access, `status`; `URLSearchParams` preserves repeated query params)
- Shared Vitest setup forces `globalThis.fetch` for `createRequester` **and** `createClient` so nock can intercept (get-it v9 / client v8 use undici with a private dispatcher); `invokeSanityCli` tests pass the nock-patched fetch through the execution context, exercising the production lineage-stripping on the wire
- `downloadStream` tests run entirely on `get-it/mock`: `streamBody()` / `streamStall()` for stalled bodies, `toHaveBeenCancelled()` plus 9.4's signal-abort tracking (`abortCount` / `lastAbortReason`) for teardown, `respondPersist({delay})` + fake timers for per-attempt deadlines, and a `toHaveHeader` check that the lazy `@sanity/cli-core@<version>` User-Agent flows through

### Testing

- Full unit suite: 3712/3722 passing; the 3 remaining failures (`initStudio` outro, `formatDocumentValidation` snapshots) reproduce identically on the branch without these changes (environment-only)
- Full integration suite: identical pass/fail set with and without these changes (the 99 local failures are auth-requiring suites without a token in the sandbox; unchanged from baseline)
- New unit tests: `enrichAuthError` (login hint, members link, staging host, idempotency), execution-context `resolveFetch` injection and lineage stripping (including host-supplied `fetch`), auth-hint proxy behavior (direct 401, sub-client 401, `withConfig` derivation, non-401 passthrough), plus the existing `createRequester` / `nodeReadableFromWeb` / backup stream suites
- `DEBUG='sanity:*' npx sanity debug` verified end-to-end in `fixtures/basic-studio` after the obug swap (namespaces, `.extend()` output)
- `pnpm check:types`, `pnpm check:lint`, `pnpm check:deps`, and `pnpm build:cli` pass

### Notes for release

N/A

<div><a href="https://cursor.com/agents/bc-16ac5cda-6f90-4b5b-8450-d636e98628de"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-web-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-web-light.png">Open in Web</picture></a>&nbsp;<a href="https://cursor.com/background-agent?bcId=bc-16ac5cda-6f90-4b5b-8450-d636e98628de"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cursor.com/assets/images/open-in-cursor-dark.png"><source media="(prefers-color-scheme: light)" srcset="https://cursor.com/assets/images/open-in-cursor-light.png">Open in Cursor</picture></a>&nbsp;</div>


packages/@sanity/cli-core/src/request/createRequester.ts
```diff
@@ -1,79 +1,124 @@
...
+    return undefined
   }
 
-  // 4. promise (must be last)
-  if (opts?.promise !== false) {
-    const promiseDefaults = {onlyBody: true}
-    const customPromise = typeof opts?.promise === 'object' ? opts.promise : {}
-    middleware.push(promise({...promiseDefaults, ...customPromise}))
+  const customHeaders = typeof headers === 'object' ? headers : {}
+  if (customHeaders['User-Agent'] || customHeaders['user-agent']) {
```
rexxars: Theoretically I'd say

```suggestion
  if (Object.keys(customHeaders).some(header => header.toLowerCase() === 'user-agent')) {
```

packages/@sanity/cli-test/src/mocks/cli-core/request.ts
```diff
@@ -1,32 +1,28 @@
...
+}
+
 /** @internal */
 export const mockRequest: Mock = vi.fn()
 /** @internal */
 export const createRequester: Mock = vi.fn(() => mockRequest)
-// get-it/middleware mocks
 /** @internal */
-export const agent: Mock = vi.fn(() => ({finalizeOptions: vi.fn()}))
+export const HttpError = class HttpError extends Error {
```
rexxars: Are we isolating people from get-it's `HttpError` for a reason here, or could we re-export that?

.changeset/pr-1534.md
```diff
@@ -0,0 +1,8 @@
+---
+'@sanity/cli-core': minor
```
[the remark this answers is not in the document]
rexxars: (seems like all users of this API are internal, so should be safe)

packages/@sanity/cli/src/actions/backup/downloadAsset.ts
```diff
@@ -34,20 +25,9 @@ export async function downloadAsset(
...
-    timeout: {connect: CONNECTION_TIMEOUT, socket: READ_TIMEOUT},
-    url,
-  })
-
-  backupDownloadDebug(
-    'Received asset %s with status code %d',
-    normalizedFileName,
-    response.statusCode,
-  )
+  const status = await downloadStream(url, createWriteStream(assetFilePath))
```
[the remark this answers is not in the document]
rexxars: It was marked as resolved, but is it not true? Could we make the destination a stream or a deferred fn? eg `downloadStream(url, () => createWriteStream(assetFilePath))` or similar?

packages/@sanity/cli/src/actions/backup/downloadStream.test.ts
```diff
@@ -0,0 +1,152 @@
+import {Writable} from 'node:stream'
```
rexxars, line 1: I'd like to see if `get-it/mock` can make testing these kind of things easier - it seems to me like we're kind of re-testing core request behavior here, but I can do a follow-up PR for it.

(If it can't fill this role, it might also reveal gaps that we should patch :) )

packages/@sanity/cli-core/src/request/createRequester.ts
```diff
@@ -1,79 +1,124 @@
...
  *
- * Each key corresponds to a middleware. Omitting a key uses the default,
- * `false` disables it, and an object merges with the defaults.
+ * Extends get-it v9's requester options with Sanity CLI defaults for
+ * User-Agent headers and debug logging.
  *
  * @public
  */
-export type MiddlewareOptions = {
+export type CreateRequesterOptions = {
```
rexxars: This is duplicating a lot... is it to avoid dual-package hazards or something? Could we not reuse get-it's `RequesterOptions`?

packages/@sanity/cli/src/actions/backup/downloadStream.ts
```diff
@@ -0,0 +1,93 @@
...
+  nodeReadableFromWeb,
+  retry,
+  type StreamResponse,
+  type WrappingMiddleware,
+} from '@sanity/cli-core/request'
+
+const CONNECTION_TIMEOUT = 15 * 1000 // 15 seconds
+const READ_TIMEOUT = 3 * 60 * 1000 // 3 minutes
+
+const connectionTimeout: WrappingMiddleware = async (options, next) => {
```
rexxars: This is interesting. I feel like there's a missing gap in get-it here that I want to plug, but in the meantime this works

rexxars (dismissed):
Overall this looks good, just left some comments and potential room for improvemnts

rexxars (dismissed):
Sweet. Thanks for also help prove the mocks/api gaps in get-it v9!

rexxars (approved):
I didn't like how some of the request interception was done, so I provided some better hooks in get-it + client and made some changes - this looks good from my end now.

PR #1540, opened 2026-07-16
feat(workbench): centralize app id generation

gu-stav:
### Description

App id's were minted three different ways across `dev`, `build`, and `deploy`, and the three could collide. A running `sanity dev` app and its deployed twin shared `deployment.appId`, so the workbench treated them as one — duplicate keys, ambiguous panel routing, and an app view that failed to load or re-render when switching between them ([SDK-1906](https://linear.app/sanity/issue/SDK-1906)).

This mints the id in one place (`packages/@sanity/workbench-cli/src/appId.ts`), three ways, so the modes can't collide:

| Mode                            | Id                  | Derived from                                                                                                 |
| ------------------------------- | ------------------- | ------------------------------------------------------------------------------------------------------------ |
| `sanity dev`                    | `<host>-<port>`     | the address the dev server actually bound (port stays visible in the URL)                                    |
| `sanity build` / `sanity start` | shape hash          | sha1 of the app's declared shape, inlined into the bundle as `__SANITY_APP_ID__`; stable wherever it's built |
| `sanity deploy`                 | applications-API id | app created in Brett first (a redeploy reuses `deployment.appId`), then the build is baked with that id      |

We are planning to move to use `slug` more in the future, so this is a temporary solution.

For `build`/`start`, editing component code keeps the id; changing the declared shape (name, organization, entry, an interface, the config) mints a new one.

The deploy path depends on Brett decoupling create from deploy ([brett#538](https://github.com/sanity-io/brett/pull/538)).

### Testing

Unit tests cover the id derivation (`appId.test.ts`), the create/deploy split, and the dev registry id. The deploy, build, and dev suites were updated for the new id plumbing.

---

> [!NOTE]
> **Medium Risk**
> Changes core deploy ordering, API create vs deploy contracts, and runtime bus identity across dev/build/deploy—high user impact but covered by extensive test updates.
> 
> **Overview**
> Fixes workbench treating a local `sanity dev` instance and its deployed twin as the same app by **centralizing id minting** in `@sanity/workbench-cli` (`resolveAppId` / `buildAppId`): **dev** uses `host-port` (registry and Vite `define` align on the configured port), **build/start** use a SHA-1 of the declared app shape, and **deploy** inlines the Brett applications API id after the record exists.
> 
> **Deploy flow** splits application **create** from **deployment** (`createCoreApp` / `createStudio` + `deployWorkbenchApp`): first deploy creates the app, builds with `applicationId`, then ships the tarball; failures roll back the new record, **`--no-build` is blocked** on first deploy, and workbench studio targeting uses the app **`slug`** instead of `studioHost`.
> 
> Builds also write **`sanity-app-id.txt`** so `sanity start` advertises the same id as the bundle (including post-deploy API ids); dev no longer passes `deployment.appId` into workbench dev.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5243f742839cca4aedf5224d4ead57b6b83374fb. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


gu-stav:

- #1564
    - #1540 ◀

packages/@sanity/workbench-cli/src/services/applications.ts
```diff
@@ -73,35 +73,20 @@ export async function getApplication(applicationId: string): Promise<Application
...
-    slug,
-    tarball,
-    title,
-    type,
-    version,
-    visibility,
-    workspaces,
-  } = options
+  const {isSingleton, organizationId, projectId, slug, title, type, visibility} = options
   const formData = new FormData()
```
mttdnt: We switched from multipart to json for the create so this is erroring out when Im testing. Should be something like:

```
 const client = await getClient()
  return client.request({
    method: 'POST',
    uri: `/applications`,
    body: {
      type,
      title,
      organizationId,
      slug,
      ...(isSingleton === undefined ? {} : {isSingleton}),
      ...(visibility ? {visibility} : {}),
      // Studio config is set once, at create — it's immutable on redeploy.
      ...(projectId ? {config: {studio: {projectId}}} : {}),
    },
  })
```

packages/@sanity/cli/src/actions/deploy/deployStudio.ts
mttdnt, line 423: this should have the `schemaDescriptorId` right?

mttdnt (changes requested):
Tried deploying a studio, ran into a few issues

packages/@sanity/workbench-cli/src/actions/deploy/deployWorkbenchApp.ts
```diff
@@ -10,152 +10,116 @@ import {
...
   } = options
   const tarball = pack(dirname(sourceDir), {entries: [basename(sourceDir)]}).pipe(createGzip())
 
-  const spin = spinner('Deploying to sanity.studio').start()
+  const spin = spinner(label).start()
   try {
-    if (appId) {
-      await createDeployment({
-        applicationId: appId,
-        interfaces,
```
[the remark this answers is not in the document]
gu-stav: Good catch, this was real. The create-time rollback stayed armed through `deployWorkbenchApp`, so a metadata PATCH failure after the deployment already shipped would `deleteApplication` a live app (and the studio path never disarmed it at all).

Fixed in 5243f7428: `deployWorkbenchApp` now fires an `onDeployed` hook the instant the deployment ships, before the metadata sync, and both callers disarm rollback there. Pre-ship failures (build, verify, `createDeployment`) still roll back; nothing past a live deployment does. Added unit tests for both directions.

mttdnt (approved):

PR #1556, opened 2026-07-17
test(workbench): consolidate unit-test mocking into cli-test

gu-stav:
### Description

Follow-up from the undeploy stack (SDK-1886), scoped to `@sanity/workbench-cli`. The package hand-rolled test scaffolding that `@sanity/cli-test` already ships: a mock `Output` factory, a partial `ux` mock, and the `getGlobalCliClient` mock triple duplicated across four files.

This adopts the shared exports so there's one source of truth: `createMockOutput` (promoted to `@sanity/cli-test/test/util`), `mocks/cli-core/ux`, and `mocks/cli-core/apiClient`. It also wires `cli-test` into the package — devDependency, tsconfig paths, and vitest `tsconfigPaths`.

### What to review

- `createMockOutput` lands in cli-test (`test/util`) and replaces the `as unknown as Output` casts.
- The four deploy/undeploy/services tests now mock `@sanity/cli-core` by spreading the shared `apiClient` mock instead of hoisting their own `getGlobalCliClient`; `deployWorkbenchApp` also swaps its partial `ux` mock for the shared one.
- `eslint-config-cli` now lets `__tests__/**` import devDependencies — needed by `devTestHelpers.ts`, which isn't a `.test.ts` file.

### Testing

Covered by the existing workbench-cli unit suite; all tests pass. Typecheck, lint, format, and knip are clean (the pre-existing `BrettInterface` type noise on `main` is untouched and out of scope).

### DIFF
--- a/packages/@sanity/workbench-cli/src/actions/deploy/__tests__/deployConfig.test.ts
+++ b/packages/@sanity/workbench-cli/src/actions/deploy/__tests__/deployConfig.test.ts
@@ -1,23 +1,23 @@
 import {Readable} from 'node:stream'
 
-import {type Output} from '@sanity/cli-core'
+import * as apiClient from '@sanity/cli-test/mocks/cli-core/apiClient'
+import {createMockOutput} from '@sanity/cli-test/test/util'
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest'
 
 import {deployConfig, resolveInstallationId, summarizeConfig} from '../deployConfig.js'
 
-const mockGetGlobalCliClient = vi.hoisted(() => vi.fn())
-const mockRequest = vi.hoisted(() => vi.fn())
+const mockRequest = vi.fn()
 
 vi.mock('@sanity/cli-core', async (importOriginal) => ({
   ...(await importOriginal<typeof import('@sanity/cli-core')>()),
-  getGlobalCliClient: mockGetGlobalCliClient,
+  ...(await import('@sanity/cli-test/mocks/cli-core/apiClient')),
 }))
 
 // The tarball content is irrelevant here — stub a readable so no build output
 // has to exist on disk.
 vi.mock('tar-fs', () => ({pack: () => Readable.from(['remote'])}))
 
-const output = {error: vi.fn(), log: vi.fn(), warn: vi.fn()} as unknown as Output
+const output = createMockOutput()
 
 /** Answer the installations list with `data`, and the config POST with a record. */
 function stubBrett(data: unknown[]) {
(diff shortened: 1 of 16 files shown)


gu-stav:

- #1556 ◀

packages/@sanity/workbench-cli/src/actions/deploy/__tests__/deployConfig.test.ts
```diff
@@ -1,23 +1,23 @@
...
 import {afterEach, beforeEach, describe, expect, test, vi} from 'vitest'
 
 import {deployConfig, resolveInstallationId, summarizeConfig} from '../deployConfig.js'
 
-const mockGetGlobalCliClient = vi.hoisted(() => vi.fn())
-const mockRequest = vi.hoisted(() => vi.fn())
+const mockRequest = vi.fn()
 
 vi.mock('@sanity/cli-core', async (importOriginal) => ({
   ...(await importOriginal<typeof import('@sanity/cli-core')>()),
```
filmaj, line 12: Two things:
1. Don't import the entirety of `@sanity/cli-core` in the source-under-test (`deployConfig.ts`). If you do, then you are transitively importing JSDOM (a 42MB dependency) and that significantly slows down transpilation, which means tests will be impacted too. Instead, use the granular exports from `@sanity/cli-core` to pull in only what the module needs. Looks like `deployConfig.ts` only needs `import {type Output, subdebug} from '@sanity/cli-core'`, so I would split that up into two import calls: `import {type Output} from '@sanity/cli-core/types'` and `import {subdebug} from '@sanity/cli-core/debug'`. That a) significantly speeds up building `deployConfig.ts` and b) also means that when testing the module, neither of those core imports need to be mocked since those particular exports and their import graphs are extremely lightweight.
2. When you do `await importOriginal` in `vi.mock` calls, you are essentially pulling in the entire module you are trying to mock - which, while convenient, also significantly impacts transpilation time. vitest transpiles tests on demand on every test invocation, meaning all of `sanity/cli-core` is pulled into the tranpilation graph (including 42MB of JSDOM) to run this test. Not good! If you look at the end of the CI runs for the `unit tests`, there will be an import time breakdown. In CI, a vast majority of time for running tests is spent just processing the import graph and transpiling - on average, only about 10% of time is spent running tests (see e.g. [this PR's recent run](https://github.com/sanity-io/cli/actions/runs/29580359359/job/87884365045#step:6:1603): 290 seconds importing things, 30 seconds to transpile, and only 25 seconds to run the tests!). Wherever we can, we should use a) the 'thin' `cli-core` exports and b) avoid these sorts of `importOriginal` calls for mocking.

The mocks provided by `cli-tests` for the `cli-core` exports are there and easy to use. They are extremely lightweight and provide a mock for every single export in `cli-core`, so in theory you should never need to `importOriginal` for any `cli-core` mocks. See e.g. [this test](https://github.com/sanity-io/cli/blob/a66212cb4330dc040ea9f856f1976aca21f85c1e/packages/%40sanity/cli/src/prompts/__tests__/promptForDatasetName.test.ts#L8): it `vi.mock`s `cli-core/ux`, doesn't need to call `importOriginal`, and just imports the `cli-test` mock for all the `cli-core/ux` APIs. Easy and super fast.

PR #1561, opened 2026-07-17
feat: add `sanity api` command for direct HTTP API access

runeb:
### Description

Adds a new top-level command, `sanity api <endpoint>`, modeled on the `gh api` pattern: direct, bearer-authenticated HTTP access to every Sanity API documented in the published OpenAPI specifications. It is the fallback for humans and agents whenever a tailored command group doesn't exist yet — the CLI takes care of auth, host selection and API versioning, the user supplies the path.

Closes [AIGRO-5218](https://linear.app/sanity/issue/AIGRO-5218/add-sanity-api-cli-command-which-directly-drives-http-apis-from).

**How it works with the OpenAPI specs.** The set of reachable APIs is driven by the published specs, not hand-maintained code: a routing manifest (`src/generated/apiRoutes.ts`) is distilled from `https://www.sanity.io/docs/api/openapi` (the same source as `sanity openapi list|get`). Per API family it records the host (`api.sanity.io` vs `<projectId>.api.sanity.io`), the default API version, and the path patterns. Endpoint paths copied verbatim from `sanity openapi get <slug>` just work: `{projectId}`/`{dataset}` placeholders are filled from flags or CLI config, and the request is routed to the right host with a sensible version. Route matching is deterministic: a pattern only matches when the request path consumes it completely, literal segment matches outrank placeholder matches, the global host wins host-ambiguous ties (it needs no project ID), and same-host ties resolve to the first manifest entry.

**Regeneration runs on every build** (`prebuild` hook → `pnpm generate:api-routes`), so every build — including release builds — compiles a manifest freshly distilled from whatever is published at that moment. The generator retries transient failures, fetches specs concurrently, and **fails the build** when the spec index is unreachable, any listed spec fails to fetch, or the distilled manifest would be empty — a release can never ship with a missing or empty endpoint list. `--check` mode additionally lets CI detect drift of the committed snapshot. Proxy-related environment variables (`HTTPS_PROXY`, `NODE_EXTRA_CA_CERTS`, …) pass through turbo to the build so the spec fetch works behind TLS-intercepting proxies. Build-time generation was chosen over runtime fetching deliberately: routing info is tiny and degrades gracefully (flags and full URLs always override), runtime fetching would couple every cold CLI call to the docs site's availability plus ~24 spec fetches per cache miss, and agents/CI want deterministic per-version behavior. This matches how stripe/aws/twilio vendor specs into releases. Possible follow-up: an on-miss runtime lookup with a disk cache for APIs published between releases.

**Usage examples**

```sh
sanity api users/me
sanity api projects/{projectId}
sanity api 'data/query/{dataset}' -f query='*[_type == "movie"][0..2]'
sanity api projects/{projectId} -X PATCH -F displayName="My project"
echo '{"mutations": []}' | sanity api 'data/mutate/{dataset}' --input - -H 'Content-Type: application/json'
sanity api jobs/123 --include --api-version v2025-02-19
```

**Behavior (gh-api parity where it makes sense)**

- `-f/--raw-field` (string) and `-F/--field` (typed: `true`/`false`/`null`/numbers, `@file`, `@-` for stdin) with gh's nested-key semantics: `a[b]=1` builds objects, `list[]=x` appends to arrays, consecutive `key[][sub]=…` fields fill the current array element until a sub-key repeats (so arrays of multi-key objects are expressible), and a bare `key[]` declares an empty array
- `--input <file|->` for raw bodies, read as raw bytes so binary uploads (eg assets) arrive byte-for-byte
- Method defaults to GET, or POST when fields/`--input` are present; for GET/HEAD, fields become query parameters
- Bodies built from fields are sent with `Content-Type: application/json`; raw `--input` bodies are sent without a default Content-Type — provide one with `-H` when the API requires it (gh api parity)
- Response body passes through to stdout (pretty-printed JSON; `--pretty` colorizes; non-JSON verbatim); `--include` prints status + headers
- HTTP errors print the error body and exit 1 (with a `sanity login` hint on 401); usage errors — bad fields, unresolved placeholders, empty or version-only paths — exit 2, per the repo exit-code convention
- `--global` / `--project-hosted` force a host; full URLs are accepted only when they are `https://` and on a Sanity API host, so the CLI token is never sent to other hosts or over unencrypted connections
- Auth, API host selection and URL construction reuse the existing CLI client factories; `--token` (or `SANITY_AUTH_TOKEN`) substitutes an explicit token and `--anonymous` sends no token; the request itself uses a bare `get-it` requester so status, headers and non-JSON bodies pass through unfiltered

### What to review

**The diff is ~2,700 additions, but only ~800 lines need close review.** Where the lines are:

| Bucket | Lines | Review posture |
|---|---|---|
| Hand-written core logic | ~800 | **Read closely** (see priorities below) |
| Unit tests (4 files) | ~950 | Skim for coverage gaps |
| Generated routing manifest (`src/generated/apiRoutes.ts`) | ~400 | Don't review line-by-line — it's build output, rewritten on every build; review its *generator* instead |
| `commands/api.ts` declarative surface (flag definitions, help text, examples) | ~220 | Skim for naming/UX |
| Generated README docs (`pnpm readme`) | ~100 | Skip |
| Generation script plumbing (fetch/retry/rendering) | ~150 | Skim |
| Changeset, package.json, turbo.json, oxfmt ignore | ~20 | Glance |

**Review priorities, highest first:**

1. `src/actions/api/resolveEndpoint.ts` — the security- and correctness-critical piece: full-URL constraints (https-only, Sanity API host allowlist — the CLI token must never leave that boundary), placeholder substitution, deterministic host matching via segment scoring and tie-breaks, API version precedence (`--api-version` > version embedded in path > matched spec default > fallback constant)
2. `src/services/api.ts` — token handling (`--token` wins over the stored login token; user-supplied `Authorization` header wins over both, case-insensitively; `--anonymous` drops the token), header assembly, body serialization (field objects → JSON with JSON content-type, strings/Buffers verbatim without one)
3. `src/actions/api/parseFields.ts` — gh-style field parsing/typing with gh's array-element grouping semantics
4. `src/commands/api.ts` `run()`/`performRequest()` — method inference, fields→query vs body, exit-code mapping
5. `src/actions/api/distillApiRoutes.ts` + `scripts/generate-api-routes.ts` — what the manifest is built from; runs on every build via `prebuild` and fails the build on fetch errors or an empty manifest

Note for local development: `pnpm build:cli` requires network access to `www.sanity.io` (by design — see above).

Known limitations (intentional for v1, called out for follow-up): streaming endpoints (`data/listen`, large exports) are buffered; no `--jq`/`--paginate`.

### Testing

**Unit tests.** 90 tests cover field parsing (typing, gh array-element grouping and empty arrays, `@file`/stdin, error cases), endpoint resolution (routing, tie-breaks, full-pattern-consumption matching, placeholders including `datasets/{name}`, version precedence, https/host allowlisting, empty-path rejection), spec distillation (hosts, base paths, embedded versions, query-marker stripping, determinism), and command behavior via `testCommand` + `nock` against both hosts (auth header presence/absence including `--token` and case-insensitive `-H` overrides, POST bodies and Content-Type defaults, byte-for-byte binary bodies, query fields, stdin input, JSON-string response printing, error passthrough and exit codes). Both prebuild outcomes are verified: a successful build regenerates the manifest, and a build with the spec endpoint unreachable fails with exit 1 after retries.

**Live smoke tests (local builds against a non-production environment): ✅ works as designed.** Smoke-test passes drove the command live across **all 24 `sanity openapi list` families** on test projects:

- **Host routing correct for all three spec shapes**: global-hosted, project-hosted (verified via debug traces of the resolved URLs), and version-in-path families (auto-injects the matched spec's default version)
- **gh parity confirmed**: method defaults, fields→query on GET, typed `-F` conversion, `@file`/`@-` expansion (with `-f` staying literal and `@`-content not re-typed — both matching gh), `--input <file>` and `--input -`
- **Binary fidelity**: a 272-byte fixture containing invalid-UTF-8 sequences and all 256 byte values uploads with a matching sha256 via both `--input <file>` and stdin
- **Write lifecycle end-to-end**: documents (mutations + Actions API), dataset create, CORS, webhook, schedule, custom role, asset upload, dataset copy → async job polled to `completed`, agent-action prompt returning a real AI response — all created, read back, and deleted
- **Error semantics**: 4xx bodies pass through verbatim with exit 1 (401 login hint shown, suppressed under `--anonymous`); usage errors and URL-constraint rejections exit 2; feature-gated endpoints surface the API's own error body — the command does no client-side plan/permission checks, by design

A follow-up smoke pass on the current head verified the review-round changes live: gh array-element grouping (one element per repeated key, bare `key[]` → `[]`, scalar appends intact), `projects/{projectId}/datasets/{name}` resolving from the dataset, prefix paths (`data`) falling back to the global host while full paths still route project-hosted, no default Content-Type on `--input` with field-built bodies still marked `application/json`, case-insensitive `-H` overrides replacing defaults without duplicates, `--token` behavior (used verbatim, exclusive with `--anonymous`), repeated query keys parsed into arrays with `__proto__` handled safely, and an `--input`-driven document write lifecycle end-to-end. Notably, the mutate endpoint accepted a piped JSON body without an explicit Content-Type — so the plain `--input -` flow keeps working even without `-H`, which stays recommended only for endpoints that are strict about media types.

### Notes for release

Added a new `sanity api <endpoint>` command: make authenticated HTTP requests to any Sanity API documented in the published OpenAPI specs, with `{projectId}`/`{dataset}` placeholder substitution, gh-style field flags, `--token` support, and automatic host and API-version routing. Discover available APIs with `sanity openapi list`.

https://claude.ai/code/session_01SvSx8KFcwJBduAUszLYH8c


runeb:
bugbots comment

> Minor doc/help tweaks for cors add (-y) and tokens add role text.

seems related to the merge commit bringing the PR up to date with `main` , not this PRs changes

Edit: rebased instead

packages/@sanity/cli/src/actions/api/distillApiRoutes.ts
```diff
@@ -0,0 +1,126 @@
...
+
+    for (const base of byHost.values()) {
+      const patterns = new Set<string>()
+      let defaultApiVersion = base.defaultApiVersion
+
+      for (const path of paths) {
+        const {segments, version} = normalizePathSegments(path)
+        defaultApiVersion = defaultApiVersion ?? version
+        const pattern = [...base.segments, ...segments].join('/')
+        if (pattern !== '') patterns.add(pattern)
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in ab40c0f: path keys are now split on `?` before normalization, with a regression test, and the manifest has been regenerated (the `{userApplicationId}?` artifact is gone).

One nuance on impact: in the current manifest this couldn't misroute — the affected endpoint still matched the Applications entry via its other patterns, and that entry is global-hosted with no default version, identical to the unmatched fallback. It was a latent bug for future specs rather than a live one. Fixed regardless.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/commands/api.ts
```diff
@@ -0,0 +1,323 @@
...
+    flags: {input?: string},
+    fields: Record<string, FieldValue>,
+    hasFields: boolean,
+    useQueryFields: boolean,
+    stdin: string | undefined,
+  ): Promise<string | unknown | undefined> {
+    if (flags.input !== undefined) {
+      if (flags.input === '-') return stdin
+      try {
+        return readFileSync(flags.input, 'utf8')
```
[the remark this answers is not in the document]
runeb: Fixed for `--input` in ab40c0f: file bodies are read as raw bytes (`readFileSync` without encoding), stdin is read via `buffer()` instead of `text()`, and the request layer passes `Buffer` bodies through untouched instead of stringifying. Added a test asserting a non-UTF-8 body arrives byte-for-byte.

The second flagged location (`-F key=@file`) is intentionally unchanged: typed fields become string values inside a JSON body, where UTF-8 text is the only representable form — raw binary can't be carried in a JSON string field regardless of how the file is read. This also matches `gh api -F @file` behavior.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,236 @@
...
+ */
+function matchRoutes(segments: string[], routes: ApiRouteEntry[]): RouteMatch | undefined {
+  let best: RouteMatch | undefined
+
+  for (const entry of routes) {
+    for (const pattern of entry.pathPatterns) {
+      const score = scorePattern(segments, pattern.split('/'))
+      if (score === 0) continue
+      const isBetter =
+        !best || score > best.score || (score === best.score && entry.host === 'global')
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in b2f1889: the global-host preference now only fires when it actually flips a project-host tie (`best.host === 'project' && entry.host === 'global'`). Same-host ties keep the first entry, so the matched spec — and the default API version it pins — no longer depends on later entries in the manifest. Added tests for both behaviors: the `copy`/`projects-api` overlap deterministically resolves to the first (most specific) entry, and the cross-host global preference holds regardless of entry order.

Worth noting the blast radius was version-only: both entries in the overlapping case are global-hosted, so the host could never be wrong — only which spec's default `apiVersion` got applied.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/services/api.ts
```diff
@@ -0,0 +1,147 @@
...
+
+  let requestBody: Buffer | string | undefined
+  if (typeof body === 'string' || Buffer.isBuffer(body)) {
+    requestBody = body
+  } else if (body !== undefined) {
+    requestBody = JSON.stringify(body)
+  }
+  if (requestBody !== undefined) {
+    requestHeaders['Content-Type'] = 'application/json'
+  }
```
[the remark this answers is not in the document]
runeb: This one is a deliberate default rather than a bug, so I've documented it instead of changing behavior (b2f1889 updates the command description and the `--input` flag help).

Rationale: the primary `--input` workflow is JSON — piping mutations/documents from stdin — and the Sanity APIs those bodies target require `application/json`. Defaulting raw bodies to `application/octet-stream` (or omitting the header) would break that main flow with confusing 400s, while binary uploads are the exception and already need an explicit `-H 'Content-Type: …'` to convey the actual media type (which the live smoke test did, successfully). This also matches `gh api`, which sends JSON content type for its request bodies.

Help text now states: request bodies are sent with `Content-Type: application/json` unless a Content-Type header is provided with `-H` (recommended for binary `--input` bodies).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,239 @@
...
+  if (!ALLOWED_URL_HOST_RE.test(url.hostname)) {
+    throw new ApiUsageError(
+      `Refusing to send authenticated requests to "${url.hostname}" - only *.api.sanity.io hosts are supported`,
+    )
+  }
+
+  const query = parseQueryString(url.search.replace(/^\?/, ''))
+  url.search = ''
+
+  return {kind: 'url', query, url: url.toString()}
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in ea7e9ff: full-URL endpoints must be `https://` — any other protocol is refused with a usage error (exit 2) before a request is made, so the token can never go out unencrypted. There's no legitimate use for `http://` here since the host allowlist only permits Sanity API hosts, which are HTTPS-only services. Test added.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,239 @@
...
+  }
+
+  // An API version embedded in the path (eg `v2021-06-07/projects`) is peeled
+  // off and re-applied as the version segment of the final URL.
+  let embeddedVersion: string | undefined
+  if (segments[0] === '{apiVersion}') {
+    segments.shift()
+  } else if (API_VERSION_SEGMENT_RE.test(segments[0])) {
+    embeddedVersion = segments.shift()
+  }
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in ea7e9ff: the empty-path check runs after version peeling, so an endpoint consisting of only an API version (`v2025-02-19`, `{apiVersion}`) is rejected as a usage error (exit 2) instead of producing a request to the API root. Tests added for both forms.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/parseFields.ts
```diff
@@ -0,0 +1,183 @@
...
+      return
+    }
+
+    const existing = container[segment]
+    if (existing === undefined) {
+      container[segment] = nextIsArray ? [] : {}
+    } else if (typeof existing !== 'object' || existing === null) {
+      throw new ApiUsageError(`Field "${key}" conflicts with an earlier field`)
+    }
+    container = container[segment] as FieldValue[] | Record<string, FieldValue>
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in eadbd93, following the pattern used in vercel/vercel's CLI (their `parseQueryString` builds user-keyed dictionaries with `Object.create(null)`): field containers and query-string dictionaries are null-prototype, and the conflict check uses `Object.hasOwn` instead of `in`. That neutralizes the class structurally rather than blacklisting key names — `__proto__` can't reach a prototype (there isn't one), and inherited names like `toString`/`constructor` work as ordinary field keys instead of false-conflicting. The same treatment was applied to the endpoint query-string parser, which had the sibling flaw. Tests cover pollution absence, `__proto__` serializing into the body as a normal key, and inherited-name keys.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,251 @@
...
+    } else if (patternSegment === segments[index]) {
+      score += 2
+    } else {
+      return 0
+    }
+  }
+
+  // Require at least one literal segment match so a pattern like
+  // `{libraryId}/...` can't capture arbitrary paths.
+  if (score < 2) return 0
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in 04efd91: `scorePattern` counts literal segment matches explicitly and returns 0 when there are none, so placeholder-only patterns can never match — which is what the guard's comment claimed all along. Test added with a `{resourceType}/{resourceId}`-style pattern asserting it can't capture unrelated paths.

Context on impact: no pattern in the current manifest is placeholder-only (the distiller always prefixes the spec's literal server base path), so this was unreachable today — but with new OpenAPI specs rolling out continuously, closing latent routing gaps is worth it.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,253 @@
...
+  // provided in place of placeholders still match the spec patterns.
+  const match = forceHost ? undefined : matchRoutes(substituted, routes)
+  const host = forceHost ?? match?.host ?? 'global'
+
+  if (host === 'project' && !projectId) {
+    throw new ProjectIdRequiredError()
+  }
+
+  return {
+    apiVersion: apiVersion ?? embeddedVersion ?? match?.defaultApiVersion ?? API_DEFAULT_VERSION,
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in 0b885fd: forcing a host now restricts route matching to specs served on that host instead of skipping matching entirely. The matched spec's default API version applies under `--global`/`--project-hosted` exactly as the documented precedence says (e.g. `sanity api projects --global` uses the Projects API's default version, and dual-host APIs match the entry for the forced host). When no spec on the forced host matches, the fallback constant applies as before. Tests added for both cases.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/README.md
```diff
@@ -113,7 +114,7 @@ Code for sanity cli
 - [`sanity telemetry disable`](#sanity-telemetry-disable)
 - [`sanity telemetry enable`](#sanity-telemetry-enable)
 - [`sanity telemetry status`](#sanity-telemetry-status)
-- [`sanity tokens create [LABEL]`](#sanity-tokens-create-label)
+- [`sanity tokens add [LABEL]`](#sanity-tokens-add-label)
```
jonahsnider: this looks like a bad rebase, if you rerun `pnpm readme` from this project it'll regenerate

packages/@sanity/cli/README.md
```diff
@@ -122,6 +123,89 @@ Code for sanity cli
...
+
+    echo '{"mutations": []}' | sanity api 'data/mutate/{dataset}' --input -
+
+  Include the response status and headers, pinning the API version
+
+    $ sanity api jobs/123 --include --api-version v2025-02-19
+
+  Authenticate with a specific token instead of the logged-in session
+
+    SANITY_AUTH_TOKEN=<token> sanity api users/me
```
jonahsnider, line 210: should this command also allow `--token`?

packages/@sanity/cli/scripts/generate-api-routes.ts
```diff
@@ -0,0 +1,153 @@
...
+import {fileURLToPath} from 'node:url'
+
+import pMap from 'p-map'
+
+import {OPENAPI_SPEC_INDEX_URL} from '../src/actions/api/constants.ts'
+import {distillApiRoutes, type SpecSource} from '../src/actions/api/distillApiRoutes.ts'
+import {type ApiRouteEntry, type OpenApiDocument} from '../src/actions/api/types.ts'
+
+const OUTPUT_PATH = join(
+  dirname(fileURLToPath(import.meta.url)),
```
jonahsnider: nit: `import.meta.dirname`<br>can be used here

packages/@sanity/cli/scripts/generate-api-routes.ts
```diff
@@ -0,0 +1,153 @@
...
+ * Regenerate with: pnpm generate:api-routes
+ */
+import {type ApiRouteEntry} from '../actions/api/types.js'
+
+export const apiRoutes: ApiRouteEntry[] = ${serialized}
+`
+}
+
+async function main(): Promise<void> {
+  const checkOnly = process.argv.includes('--check')
```
jonahsnider: nit: you could use `node:util` `parseArgs` for this

packages/@sanity/cli/scripts/generate-api-routes.ts
```diff
@@ -0,0 +1,153 @@
...
+  writeFileSync(OUTPUT_PATH, rendered)
+  console.log(`Wrote ${routes.length} route entries to ${OUTPUT_PATH}`)
+}
+
+try {
+  await main()
+} catch (error) {
+  console.error(error)
+  process.exitCode = 1
+}
```
jonahsnider: this script can use top-level await instead of wrapping everything in `main()`, for simplicity

packages/@sanity/cli/src/actions/api/__tests__/parseFields.test.ts
jonahsnider, line 1: might be interesting to go through [https://github.com/cli/cli/blob/trunk/pkg/cmd/api/fields_test.go](<https://github.com/cli/cli/blob/trunk/pkg/cmd/api/fields_test.go>) and see if they're testing anything we aren't, or if there's any interesting edge cases we should handle

packages/@sanity/cli/src/services/api.ts
jonahsnider: is it intentional that all requests with a body are marked as JSON?

seems like current logic is:

1. call fn with `body` as a buffer
2. `Buffer.isBuffer(body)` returns true
3. `requestBody = body`
4. `requestBody !== undefined` is now true, so content-type is set to JSON

which is no good

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,254 @@
...
+  // Require at least one literal segment match so a placeholder-only pattern
+  // like `{resourceType}/{resourceId}` can't capture arbitrary paths.
+  if (literalMatches === 0) return 0
+
+  // Prefer patterns the request path consumes completely over patterns that
+  // merely share a prefix with it.
+  return patternSegments.length <= segments.length ? score + 1 : score
+}
+
+function parseQueryString(queryString: string): Record<string, string | string[]> {
```
jonahsnider, line 242: i think this can be replaced with `querystring.parse`from node:querystring. it uses a null prototype object and uses arrays for repeated values

packages/@sanity/cli/src/commands/api.ts
```diff
@@ -0,0 +1,328 @@
...
+
+    if (response.statusCode >= 400) {
+      const statusLine = `HTTP ${response.statusCode}${
+        response.statusMessage ? ` ${response.statusMessage}` : ''
+      }`
+      const loginHint =
+        response.statusCode === 401 && !flags.anonymous
+          ? `. You may need to login again with ${styleText('cyan', 'sanity login')}`
+          : ''
+      this.error(`${statusLine}${loginHint}`, {exit: 1})
```
jonahsnider: nit: use the `exitCodes` enum instead of hardcoding `1`

packages/@sanity/cli/src/commands/api.ts
```diff
@@ -0,0 +1,328 @@
...
+    fields: Record<string, FieldValue>,
+    hasFields: boolean,
+    useQueryFields: boolean,
+    stdin: Buffer | undefined,
+  ): Promise<Buffer | unknown | undefined> {
+    if (flags.input !== undefined) {
+      if (flags.input === '-') return stdin
+      try {
+        // Read raw bytes - `--input` bodies may be binary (eg asset uploads)
+        return readFileSync(flags.input)
```
jonahsnider: this function is async, but we aren't using `fs/promises` to read the file here?

packages/@sanity/cli/src/actions/api/parseFields.ts
```diff
@@ -0,0 +1,185 @@
...
+ * Supports nested keys with bracket syntax: `a[b]=1` produces `{a: {b: 1}}`
+ * and `a[]=1` appends to an array. Raw fields (`-f`) keep values as strings;
+ * typed fields (`-F`) convert `true`/`false`/`null` and numbers, and expand
+ * `@<file>` / `@-` to file or stdin contents.
+ */
+export function parseFields(options: ParseFieldsOptions): Record<string, FieldValue> {
+  const {fields = [], rawFields = [], readFile, stdin} = options
+  // Null-prototype containers keep user-supplied keys like `__proto__` or
+  // `toString` ordinary own properties: nothing to pollute, nothing inherited.
+  const result: Record<string, FieldValue> = Object.create(null)
```
jonahsnider: nit: null prototype is safe but honestly for things like this where it's just raw user input, i like using `Map` and not worrying at all about any weirdness

jonahsnider (changes requested):
generally looks good, i have a few questions + change requests, as well as a handful of nitpicky things

packages/@sanity/cli/scripts/generate-api-routes.ts
runeb: Done in 43551fc.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/scripts/generate-api-routes.ts
runeb: Done in 43551fc.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/scripts/generate-api-routes.ts
runeb: Done in 43551fc — straight top-level flow now. A failed fetch rejects at top level, which still exits non-zero and fails the build.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/README.md
runeb: Regenerated in cba4ac8 (via `manifest:generate` + `pnpm readme` — the committed README had drifted because main's `tokens add` → `tokens create` rename (#1566) landed after this branch's last regen). The fresh regen syncs that along with the updated `sanity api` help.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/README.md
runeb: Added in 43551fc: `--token`/`-t` (exclusive with `--anonymous`), threaded through the CLI client factories so it takes precedence over the stored login token — same pattern as `datasets import --token`. Covered by a test.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/__tests__/parseFields.test.ts
runeb: Good call — this surfaced a real semantic gap, fixed in 43551fc. gh doesn't append a new array element per `[]`: consecutive `key[][sub]=…` fields fill the *current* (last) element until a sub-key repeats, so arrays of multi-key objects are expressible (`labels[][name]=bug labels[][color]=red` → one object), and a bare `key[]` declares an empty array. Our parser previously appended a new element per `[]`, which made multi-key elements impossible to build. I've ported gh's semantics (including the nested-array reuse rule) and brought over the interesting cases from fields_test.go: element grouping, empty arrays, deep nesting (`nested[][k1][k2][k3]=v`), mixed scalar/object arrays, `-f` keeping `@`-values literal, and the type-conflict errors.

Two deliberate divergences: `-F` converts floats as well as integers (gh converts only integers; Sanity APIs commonly take decimals — `-f` keeps strings), and gh's `{owner}`/`:owner` placeholder expansion inside field values doesn't apply here since our placeholders live in the endpoint path.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/services/api.ts
runeb: You're right — and my earlier claim on the Bugbot thread that gh does the same was wrong: gh only sets `application/json` when it builds the body from fields (`bodyIsJSON` in cli/cli's http.go); `--input` bodies get no default Content-Type. Fixed in 43551fc to match: field-built bodies are marked `application/json`, raw `--input` bodies send no default Content-Type (pass `-H` when the API requires one). Help text, README, the stdin example and tests updated accordingly.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
runeb: Done in 43551fc — `parseQueryString` is now a thin wrapper over `node:querystring`'s `parse`, kept as a function only for the narrowing cast from `ParsedUrlQuery`.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/commands/api.ts
runeb: Done in 43551fc — both here (`RUNTIME_ERROR`) and the usage-error path (`USAGE_ERROR`).

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/commands/api.ts
runeb: Done in 43551fc — `--input` file reads now use `node:fs/promises`. The remaining `readFileSync` is only for `-F key=@file`, which runs inside the synchronous field parser.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/parseFields.ts
runeb: Kept null-prototype objects here deliberately: these containers are the JSON request body itself (nested, and passed straight to `JSON.stringify`), so `Map`s would need a recursive conversion back to plain objects before serialization. Null-prototype gives the same isolation from prototype weirdness without the extra pass. For the flat user-keyed maps (`fieldsToQuery`, `parseHeaders`) I've switched those to null-prototype too in 43551fc so the treatment is consistent everywhere user keys land.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  // Require at least one literal segment match so a placeholder-only pattern
+  // like `{resourceType}/{resourceId}` can't capture arbitrary paths.
+  if (literalMatches === 0) return 0
+
+  // Prefer patterns the request path consumes completely over patterns that
+  // merely share a prefix with it.
+  return patternSegments.length <= segments.length ? score + 1 : score
+}
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in 6f86b7b: a pattern now only matches when the request path consumes it completely, so a request that is a mere prefix of a longer pattern (`data` vs `data/query/{dataset}`) falls back to the global host instead of inheriting the longer route's host and version — no more project-ID prompt for a path that can't be valid on the project host. This also made the old "fully-consumed beats prefix" score bonus redundant, so it's gone. Test added for both the one- and two-segment prefix cases.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/resolveEndpoint.ts
```diff
@@ -0,0 +1,244 @@
...
+      return projectId
+    }
+    if (placeholder === 'dataset' || placeholder === 'datasetName') {
+      if (!dataset) {
+        throw new ApiUsageError(
+          `Unable to resolve {${placeholder}} - provide a dataset with --dataset or configure one in sanity.cli.ts`,
+        )
+      }
+      return dataset
+    }
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in 6efc7ad: `{name}` now fills from the dataset when it directly follows a `datasets` segment, so the projects spec's `projects/{projectId}/datasets/{name}` paths can be copied verbatim. The context check matters because `{name}` is not always a dataset — the media-libraries spec uses it for video renditions (`.../renditions/{name}`), where it correctly stays unresolved. Tests added for both contexts.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/services/api.ts
```diff
@@ -0,0 +1,169 @@
...
+    // caller knows the media type, we don't (gh api behaves the same way)
+    requestBody = body
+  } else if (body !== undefined) {
+    requestBody = JSON.stringify(body)
+    requestHeaders['Content-Type'] = 'application/json'
+  }
+
+  for (const [key, value] of Object.entries(headers)) {
+    requestHeaders[key] = value
+  }
```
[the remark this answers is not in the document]
runeb: Confirmed and fixed in 6efc7ad: request headers are now keyed lowercase — the `authorization`/`content-type` defaults are set lowercase and user `-H` names are lowercased at merge time — so an override replaces the default regardless of how either side is cased, and duplicate differently-cased entries are structurally impossible. Test added asserting a user-cased `Authorization` header wins over the CLI token.

---
_Generated by [Claude Code](https://claude.ai/code)_

packages/@sanity/cli/src/actions/api/parseFields.ts
jonahsnider: love it, consistently doing null prototype everywhere it's user-input seems good

jonahsnider (approved):

PR #1562, opened 2026-07-17
fix(cli): add actionable login guidance

jonahsnider:
Make it clearer in unattended mode when you need to login and how to do it

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---

> [!NOTE]
> **Low Risk**
> User-facing messaging and help text only; auth behavior is unchanged aside from clearer errors and warnings.
> 
> **Overview**
> Centralizes **login guidance** in `loginInstructions.ts` and surfaces it whenever the CLI needs credentials.
> 
> When auth is missing or invalid, **`ensureAuthenticated`** and **`init`** now **`warn`** with copy that lists token, `--provider`, and `--sso` login options (including `--no-open`). **`sanity init --yes`** without a session **throws** that same message instead of a one-line “run `sanity login`” hint, using **`exitCodes.RUNTIME_ERROR`**.
> 
> **`sanity login --help`** documents accepted **`--provider`** IDs via **`LOGIN_PROVIDER_IDS`** (`google`, `github`, `sanity`, `vercel`). Tests assert the shared message on stderr/warn paths and on the provider flag description.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21368c0bf39e84b0e1a7bcc6ee48995936d3a56a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/actions/auth/login/loginInstructions.ts
+++ b/packages/@sanity/cli/src/actions/auth/login/loginInstructions.ts
@@ -1,0 +1,13 @@
+export const LOGIN_PROVIDER_IDS = ['google', 'github', 'sanity', 'vercel'] as const
+
+export const LOGIN_REQUIRED_MESSAGE = [
+  'No valid authentication credentials found.',
+  '',
+  'Authenticate with one of these commands:',
+  '  echo "$TOKEN" | sanity login --with-token',
+  '  sanity login --provider <providerId> --no-open',
+  `    Provider IDs: ${LOGIN_PROVIDER_IDS.join(', ')}`,
+  '  sanity login --sso <organizationSlug> --no-open',
+  '',
+  '`--no-open` prints a login URL instead of opening a browser.',
+].join('\n')
--- a/packages/@sanity/cli/src/commands/login.ts
+++ b/packages/@sanity/cli/src/commands/login.ts
@@ -5,6 +5,7 @@ import {type FlagInput} from '@oclif/core/interfaces'
 import {SanityCommand} from '@sanity/cli-core'
 
 import {login} from '../actions/auth/login/login.js'
+import {LOGIN_PROVIDER_IDS} from '../actions/auth/login/loginInstructions.js'
 
 export class LoginCommand extends SanityCommand<typeof LoginCommand> {
   static override description = 'Log in to your Sanity account'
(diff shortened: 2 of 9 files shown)


packages/@sanity/cli/src/actions/auth/login/loginInstructions.ts
```diff
@@ -0,0 +1,13 @@
+export const LOGIN_PROVIDER_IDS = ['google', 'github', 'sanity', 'vercel'] as const
+
+export const LOGIN_REQUIRED_MESSAGE = [
+  'No valid authentication credentials found.',
+  '',
+  'Authenticate with one of these commands:',
+  '  echo "$TOKEN" | sanity login --with-token',
+  '  sanity login --provider <providerId> --no-open',
+  `    Provider IDs: ${LOGIN_PROVIDER_IDS.join(', ')}`,
```
shapirodaniel, line 9: totally orthogonal, but curious -- in a future `use_sanity_cli` / MCP context, would we be able to signal the agent to leverage `give_feedback` there's a provider an agent wants to use that isn't listed under `LOGIN_PROVIDER_IDS`?

packages/@sanity/cli/src/commands/login.ts
```diff
@@ -42,7 +43,7 @@ export class LoginCommand extends SanityCommand<typeof LoginCommand> {
       description: 'Open a browser window to log in (`--no-open` only prints URL)',
     }),
     provider: Flags.string({
-      description: 'Log in using the given provider',
+      description: `Log in using a provider ID (${LOGIN_PROVIDER_IDS.join(', ')})`,
```
shapirodaniel, line 46: super nit / non-blocking, might be worth extracting `getLoginProviders` since we use it here + `loginInstructions.ts`?

shapirodaniel (approved):
looks great! minor non-blocking qs + suggestions

packages/@sanity/cli/src/actions/auth/login/loginInstructions.ts
jonahsnider: for MCP, you probably wouldn't even be able to submit feedback unless you're authenticated? but it's an interesting idea. arguably feedback from clients struggling to auth is one of the more important things to capture

probably would be a good idea to have a `sanity feedback` command that doesn't need auth

packages/@sanity/cli/src/commands/login.ts
jonahsnider: would it just be doing `LOGIN_PROVIDER_IDS.join(', ')`? i think if it's literally just that probs not super useful to add?

PR #1575, opened 2026-07-22
fix(workbench): derive config id with Web Crypto

gu-stav:
### Description

`sanity dev` on a workbench app with a config (media library) throws in the browser:

> Uncaught Error: Module "node:crypto" has been externalized for browser compatibility. Cannot access "node:crypto.hash" in client code.

`deriveConfigs` computed the config id with `node:crypto`'s `hash`. That module is dev-server orchestration, but the workbench Vite server's dependency scan pulls it into the browser module graph — and every other node-ish import there is type-only, so `node:crypto` was the one value import that leaked.

The fix derives the id with the Web Crypto API (`crypto.subtle.digest`), which exists in both Node and secure browser contexts. The id is only a local change-detection token — the deployed installation id comes from the applications API — so swapping the algorithm is safe. `deriveConfigs` is now async; its three call sites already run in async contexts.

Introduced by #1540.

### What to review

- `deriveInterfaces.ts`: `deriveConfigs` is async and hashes via the `contentHash` helper.
- The `eslint-disable` on the crypto call — the global is available at runtime on our Node target (>=22.12) and in the browser; `eslint-plugin-n` only flags it against Node's "stable at 23" docs label.

### Testing

Unit tests updated for the now-async `deriveConfigs` (stable id, changes on edit, rejects a non-singleton config).

---

> [!NOTE]
> **Medium Risk**
> Changes the hashing algorithm for local build/config ids (new hex values for the same app shape) and threads async through build and dev registration; deployed API ids are unaffected.
> 
> **Overview**
> Fixes **`sanity dev`** crashing in the browser when a workbench app declares config (e.g. media library): Vite’s dep scan was pulling **`deriveConfigs`** / **`appId`** into the client graph, and **`node:crypto`** cannot run there.
> 
> **Hashing** for local workbench identity tokens now uses **`globalThis.crypto.subtle.digest` (SHA-256, hex)** instead of **`node:crypto`**, in both **`deriveConfigs`** (config change-detection id) and **`buildAppId`** (build/start shape id). **`resolveAppId`** is narrowed to the sync **host-port** dev id only so the browser-facing export stays free of crypto imports.
> 
> **`deriveConfigs`** and **`buildAppId`** are **async**; dev registration, preview **`start`**, and **`sanity build`** paths **`await`** them. Deployed application ids from the API are unchanged—only locally derived hashes change algorithm (SHA-1 → SHA-256), which is acceptable for these tokens.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f2bcbfb638b001ba8599b0d8a5e40b2b9793deea. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-workbench-crypto-browser.md
+++ b/.changeset/fix-workbench-crypto-browser.md
@@ -1,0 +1,6 @@
+---
+"@sanity/workbench-cli": patch
+"@sanity/cli": patch
+---
+
+Derive workbench app and config ids with the Web Crypto API so `node:crypto` no longer crashes the dev server's browser bundle.
--- a/packages/@sanity/cli/src/actions/build/buildApp.ts
+++ b/packages/@sanity/cli/src/actions/build/buildApp.ts
@@ -55,7 +55,9 @@ export async function buildApp(options: BuildOptions): Promise<void> {
     vite: cliConfig.vite,
     // Shared with deploy: it passes the resolved application id (minted by the
     // API on a first deploy) to inline; a plain build has none, so it hashes the shape.
-    workbenchAppId: workbench ? (options.applicationId ?? buildAppId(workbench)) : undefined,
+    workbenchAppId: workbench
+      ? (options.applicationId ?? (await buildAppId(workbench)))
+      : undefined,
     workDir,
   })
 }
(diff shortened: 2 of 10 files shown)


gu-stav:

- #1575 ◀

joshuaellis (dismissed):

joshuaellis (approved):

PR #1592, opened 2026-07-23
feat(cli): add `sanity new` to mint unclaimed projects without logging in

shapirodaniel:
### Description

this PR adds `sanity new` as an alias of `sanity projects mint`. high level, we're allowing humans and agents to create projects without an account, through public apis that produce a claim URL that allows transfer of a newly-created project to the user's desired organization within 72hrs of creation

### What's included

- `sanity new` mints a new project and issues a robot token that can be used to drive programmatic content lake operations
- reminders are periodically surfaced on cli commands until the project is claimed or expires
- a `nextjs` project is scaffolded, and claims metadata is woven into scaffold-accessible environment
- `sanity init` hints at `sanity new` availability so humans and agents can discover and leverage the new behavior
- minor changes to existing cli behavior to play nicely with `sanity new`

### For reviewers

we initially gave a stacked diff a try, but the result was a bit unwieldy: `sanity new` functionality is deep intertwined, and easier to review with agentic support if it's broken into commits on a single diff as per early feedback. we've reshipped as a series of commits instead, and recommend reviewing commit-by-commit or pointing a review agent at the pr to do a holistic pass

### Testing

we added unit coverage and conducted manual uat to exercise all branching logic related to credential management

Screenshot 2026-07-28 at 4 19 52 AM
Screenshot 2026-07-28 at 4 20 14 AM

### Notes for release

Add `sanity new` to mint an unclaimed Sanity project without logging in, scaffold a Studio and a Next.js frontend for it, and claim it with a Sanity account within 72 hours to keep it. Pass `--no-scaffold` to mint without scaffolding, or `--json` for a machine-readable payload that writes no files.

---

> [!NOTE]
> **High Risk**
> Changes CLI authentication precedence, token caching, and login/logout revocation paths while introducing long-lived robot tokens in `.env` and user config—security-sensitive and easy to misconfigure across nested directories.
> 
> **Overview**
> Introduces **`sanity new`** (a top-level alias over project minting) so users and agents can create an unclaimed Sanity project without logging in, receive a robot token and claim URL, and optionally scaffold **`./sanity`** plus **`./web`** (Next.js via `create-next-app`, `next-sanity`). **`--json`** and **`--no-scaffold`** support agent-style and project-only flows; mint guards refuse remints when ancestor `.env` credential boundaries already exist (including blank guarded placeholders).
> 
> **Credential handling** is centralized in **`resolveCliCredential`**: environment `SANITY_AUTH_TOKEN`, then minted-project identity from the nearest ancestor `.env` boundary plus an **`unclaimedProjects`** ledger (with `.env` recovery), then the stored login session. **`getCliToken`** delegates to that stack; cache invalidation covers ledger writes. **Login** revokes the stored session via **`getCliUserConfig`**, not the active robot/env token, and warns when env or mint tokens still outrank the new session. **Logout** only revokes the stored session and explains when env/ledger robots remain active.
> 
> **`sanity init`** gains a “two ways to start” banner (suppressed in minted directories), mint-aware unattended errors, and clearer handling when a robot token authenticates without a user email. A **`createFlow`** helper adds rail-style CLI output for mint narration; a **claim reminders** prerun hook is registered in oclif config.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9dde7d8d468d4cd8c79c008fc0fb35559564237. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli-core/src/_exports/index.ts
+++ b/packages/@sanity/cli-core/src/_exports/index.ts
@@ -24,6 +24,10 @@ export {
   type ConfigStore,
 } from '../config/cli/types/cliConfig.js'
 export {type UserViteConfig} from '../config/cli/types/userViteConfig.js'
+export {
+  resolveMintedProjectToken,
+  UNCLAIMED_PROJECTS_CONFIG_KEY,
+} from '../config/cli/unclaimedProjects.js'
 export {type ApplicationType, isWorkbenchApp} from '../config/cli/workbenchApp.js'
 export {findProjectRoot} from '../config/findProjectRoot.js'
 export {findProjectRootSync} from '../config/findProjectRootSync.js'
--- a/packages/@sanity/cli-core/src/_exports/index.ts
+++ b/packages/@sanity/cli-core/src/_exports/index.ts
@@ -24,10 +24,6 @@ export {
   type ConfigStore,
 } from '../config/cli/types/cliConfig.js'
 export {type UserViteConfig} from '../config/cli/types/userViteConfig.js'
-export {
-  resolveMintedProjectToken,
-  UNCLAIMED_PROJECTS_CONFIG_KEY,
-} from '../config/cli/unclaimedProjects.js'
 export {type ApplicationType, isWorkbenchApp} from '../config/cli/workbenchApp.js'
 export {findProjectRoot} from '../config/findProjectRoot.js'
 export {findProjectRootSync} from '../config/findProjectRootSync.js'
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -22,8 +22,9 @@ export const _internals = {
 
 /**
  * The active CLI credential together with where it came from, in precedence order:
- * a nonblank `SANITY_AUTH_TOKEN` in the environment, then the minted-project ledger record
- * selected by the current directory's `.env` project id, then the stored login session.
+ * a nonblank `SANITY_AUTH_TOKEN` in the environment, then the minted-project credential selected
+ * by the current directory's `.env` project id (ledger first, root `.env` as recovery), then the
+ * stored login session.
  *
  * @internal
  */
@@ -48,10 +49,9 @@ export async function resolveCliCredential(cwd?: string): Promise<ResolvedCliCre
     return {source: 'environment', token: envToken}
   }
 
-  // A minted directory carries the robot token in both `.env` (for the app runtime) and the
-  // ledger. This resolves it from the ledger, keyed by the `.env` project id, because env
-  // injection never runs before a config file exists — the gap that stops `sanity init`
-  // authenticating in a freshly minted directory.
+  // A minted directory carries the robot token in both `.env` and the ledger. Resolve the ledger
+  // copy first, with the root `.env` copy as recovery when the ledger write failed. Env injection
+  // never runs before a config file exists, which is the gap this directory-aware tier closes.
   let minted
   try {
     minted = resolveMintedProjectCredential(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY), cwd)
@@ -132,9 +132,16 @@ export function setCliUserConfig(prop: 'authToken', value: string | undefined): 
     writeJsonFileSync(configPath, {...config, [prop]: value}, {pretty: true})
   }
 
-  // Invalidate the in-process token cache so subsequent getCliToken() calls
-  // pick up the new value from disk.
-  clearCliTokenCache()
+  // A successful explicit login must take effect for the rest of this process. Otherwise an
+  // invalid project-local mint token can keep winning normal resolution after the fresh session
+  // is stored, and the command that initiated login immediately fails its next authenticated
+  // request. Preserve an explicitly exported token as the highest-precedence process credential.
+  const envToken = process.env.SANITY_AUTH_TOKEN?.trim()
+  if (value && !envToken) {
+    setCachedToken(value)
+  } else {
+    clearCliTokenCache()
+  }
 }
 
 /**
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -98,6 +98,21 @@ export async function getCliToken(): Promise<string | undefined> {
   setCachedToken(token)
   return token
 }
+
+/**
+ * Make a freshly stored login session the active credential for the rest of this process.
+ *
+ * Embedded authentication flows use this after an invalid active credential triggered login.
+ * Standalone login deliberately does not, so environment and minted-project credentials retain
+ * their normal precedence and can be reported to the user.
+ *
+ * @param token - The newly stored login session token
+ * @internal
+ */
+export function activateCliSessionForProcess(token: string): void {
+  setCachedToken(token)
+}
+
 const cliUserConfigSchema = {
   authToken: z.optional(z.string()),
 }
@@ -133,16 +148,7 @@ export function setCliUserConfig(prop: 'authToken', value: string | undefined): 
     writeJsonFileSync(configPath, {...config, [prop]: value}, {pretty: true})
   }
 
-  // A successful explicit login must take effect for the rest of this process. Otherwise an
-  // invalid project-local mint token can keep winning normal resolution after the fresh session
-  // is stored, and the command that initiated login immediately fails its next authenticated
-  // request. Preserve an explicitly exported token as the highest-precedence process credential.
-  const envToken = process.env.SANITY_AUTH_TOKEN?.trim()
-  if (value && !envToken) {
-    setCachedToken(value)
-  } else {
-    clearCliTokenCache()
-  }
+  clearCliTokenCache()
 }
 
 /**
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -10,7 +10,7 @@ import {readJsonFileSync} from '../../util/readJsonFileSync.js'
 import {writeJsonFileSync} from '../../util/writeJsonFileSync.js'
 import {clearCliTokenCache, getCachedToken, setCachedToken} from './cliTokenCache.js'
 import {type ConfigStore} from './types/cliConfig.js'
-import {resolveMintedProjectToken, UNCLAIMED_PROJECTS_CONFIG_KEY} from './unclaimedProjects.js'
+import {resolveMintedProjectCredential, UNCLAIMED_PROJECTS_CONFIG_KEY} from './unclaimedProjects.js'
 
 // Re-export so existing consumers don't break
 export {clearCliTokenCache} from './cliTokenCache.js'
@@ -18,6 +18,56 @@ export {clearCliTokenCache} from './cliTokenCache.js'
 // Only used for testing
 export const _internals = {
   getCliUserConfig,
+}
+
+/**
+ * The active CLI credential together with where it came from, in precedence order:
+ * a nonblank `SANITY_AUTH_TOKEN` in the environment, then the minted-project ledger record
+ * selected by the current directory's `.env` project id, then the stored login session.
+ *
+ * @internal
+ */
+export type ResolvedCliCredential =
+  | {projectId: string; source: 'minted-project'; token: string}
+  | {source: 'environment'; token: string}
+  | {source: 'none'}
+  | {source: 'session'; token: string}
+
+/**
+ * Resolve the active CLI credential and its source. Uncached: callers that explain which
+ * identity is in effect (login/logout warnings) need a fresh answer, not a source inferred
+ * from a cached string. Token-oriented callers should keep using {@link getCliToken}.
+ *
+ * @param cwd - Directory whose `.env` selects the minted-project ledger record; defaults to
+ * the process working directory
+ * @internal
+ */
+export async function resolveCliCredential(cwd?: string): Promise<ResolvedCliCredential> {
+  const envToken = process.env.SANITY_AUTH_TOKEN?.trim()
+  if (envToken) {
+    return {source: 'environment', token: envToken}
+  }
+
+  // A minted directory carries the robot token in both `.env` (for the app runtime) and the
+  // ledger. This resolves it from the ledger, keyed by the `.env` project id, because env
+  // injection never runs before a config file exists — the gap that stops `sanity init`
+  // authenticating in a freshly minted directory.
+  let minted
+  try {
+    minted = resolveMintedProjectCredential(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY), cwd)
+  } catch {
+    // Fall through to the login session.
+  }
+  if (minted) {
+    return {projectId: minted.projectId, source: 'minted-project', token: minted.token}
+  }
+
+  const sessionToken = _internals.getCliUserConfig('authToken')
+  if (sessionToken) {
+    return {source: 'session', token: sessionToken}
+  }
+
+  return {source: 'none'}
 }
 
 /**
@@ -41,31 +91,11 @@ export async function getCliToken(): Promise<string | undefined> {
     return cached
   }
 
-  const token = process.env.SANITY_AUTH_TOKEN
-  if (token) {
-    const trimmed = token.trim()
-    setCachedToken(trimmed)
-    return trimmed
-  }
+  const credential = await resolveCliCredential()
+  const token = credential.source === 'none' ? undefined : credential.token
 
-  // A minted directory carries the robot token in both `.env` (for the app runtime) and the
-  // ledger. This resolves it from the ledger, keyed by the `.env` project id, because env
-  // injection never runs before a config file exists — the gap that stops `sanity init`
-  // authenticating in a freshly minted directory.
-  let mintedToken: string | undefined
-  try {
-    mintedToken = resolveMintedProjectToken(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY))
-  } catch {
-    // Fall through to the login session.
-  }
-  if (mintedToken) {
-    setCachedToken(mintedToken)
-    return mintedToken
-  }
-
-  const configToken = _internals.getCliUserConfig('authToken')
-  setCachedToken(configToken)
-  return configToken
+  setCachedToken(token)
+  return token
 }
 const cliUserConfigSchema = {
   authToken: z.optional(z.string()),
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -23,8 +23,8 @@ export const _internals = {
 /**
  * The active CLI credential together with where it came from, in precedence order:
  * a nonblank `SANITY_AUTH_TOKEN` in the environment, then the minted-project credential selected
- * by the current directory's `.env` project id (ledger first, root `.env` as recovery), then the
- * stored login session.
+ * by the nearest ancestor `.env` credential boundary (ledger first, that `.env` as recovery), then
+ * the stored login session.
  *
  * @internal
  */
@@ -39,8 +39,8 @@ export type ResolvedCliCredential =
  * identity is in effect (login/logout warnings) need a fresh answer, not a source inferred
  * from a cached string. Token-oriented callers should keep using {@link getCliToken}.
  *
- * @param cwd - Directory whose `.env` selects the minted-project ledger record; defaults to
- * the process working directory
+ * @param cwd - Directory where the minted-project `.env` ancestor search starts; defaults to the
+ * process working directory
  * @internal
  */
 export async function resolveCliCredential(cwd?: string): Promise<ResolvedCliCredential> {
@@ -50,8 +50,9 @@ export async function resolveCliCredential(cwd?: string): Promise<ResolvedCliCre
   }
 
   // A minted directory carries the robot token in both `.env` and the ledger. Resolve the ledger
-  // copy first, with the root `.env` copy as recovery when the ledger write failed. Env injection
-  // never runs before a config file exists, which is the gap this directory-aware tier closes.
+  // copy first, with the selected ancestor `.env` copy as recovery when the ledger write failed.
+  // Env injection remains scoped to a discovered project root, which is the gap this independent
+  // directory-aware tier closes.
   let minted
   try {
     minted = resolveMintedProjectCredential(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY), cwd)
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -48,9 +48,10 @@ export async function getCliToken(): Promise<string | undefined> {
     return trimmed
   }
 
-  // A minted directory carries only its (non-secret) project id in `.env`; the robot token lives
-  // in the ledger. This resolves it without env injection, which never runs before a config file
-  // exists — the gap that stops `sanity init` authenticating in a freshly minted directory.
+  // A minted directory carries the robot token in both `.env` (for the app runtime) and the
+  // ledger. This resolves it from the ledger, keyed by the `.env` project id, because env
+  // injection never runs before a config file exists — the gap that stops `sanity init`
+  // authenticating in a freshly minted directory.
   let mintedToken: string | undefined
   try {
     mintedToken = resolveMintedProjectToken(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY))
--- a/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
@@ -10,6 +10,7 @@ import {readJsonFileSync} from '../../util/readJsonFileSync.js'
 import {writeJsonFileSync} from '../../util/writeJsonFileSync.js'
 import {clearCliTokenCache, getCachedToken, setCachedToken} from './cliTokenCache.js'
 import {type ConfigStore} from './types/cliConfig.js'
+import {resolveMintedProjectToken, UNCLAIMED_PROJECTS_CONFIG_KEY} from './unclaimedProjects.js'
 
 // Re-export so existing consumers don't break
 export {clearCliTokenCache} from './cliTokenCache.js'
@@ -45,6 +46,20 @@ export async function getCliToken(): Promise<string | undefined> {
     const trimmed = token.trim()
     setCachedToken(trimmed)
     return trimmed
+  }
+
+  // A minted directory carries only its (non-secret) project id in `.env`; the robot token lives
+  // in the ledger. This resolves it without env injection, which never runs before a config file
+  // exists — the gap that stops `sanity init` authenticating in a freshly minted directory.
+  let mintedToken: string | undefined
+  try {
+    mintedToken = resolveMintedProjectToken(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY))
+  } catch {
+    // Fall through to the login session.
+  }
+  if (mintedToken) {
+    setCachedToken(mintedToken)
+    return mintedToken
   }
 
   const configToken = _internals.getCliUserConfig('authToken')
@@ -109,6 +124,16 @@ export function getCliUserConfig(prop: 'authToken'): string | undefined {
 }
 
 /**
+ * Keys whose writes must invalidate the in-process token cache, because `getCliToken` resolves
+ * from them: the stored login session (`authToken`) and the minted-project ledger. Without this,
+ * dropping a claimed project from the ledger leaves the robot token cached for the rest of the
+ * process, breaking the post-claim handoff to the logged-in user.
+ */
+function affectsToken(key: string): boolean {
+  return key === 'authToken' || key === UNCLAIMED_PROJECTS_CONFIG_KEY
+}
+
+/**
  * Get a key-value store backed by the CLI user configuration file
  * (`~/.config/sanity/config.json`).
  *
@@ -136,7 +161,7 @@ export function getUserConfig(): ConfigStore {
       const configPath = getCliUserConfigPath()
       mkdirSync(dirname(configPath), {recursive: true})
       writeJsonFileSync(configPath, {...config, [key]: value}, {pretty: true})
-      if (key === 'authToken') clearCliTokenCache()
+      if (affectsToken(key)) clearCliTokenCache()
     },
 
     // No mkdirSync needed: if readConfig() succeeded the directory already exists.
@@ -145,7 +170,7 @@ export function getUserConfig(): ConfigStore {
       if (!(key in config)) return
       const {[key]: _, ...rest} = config
       writeJsonFileSync(getCliUserConfigPath(), rest, {pretty: true})
-      if (key === 'authToken') clearCliTokenCache()
+      if (affectsToken(key)) clearCliTokenCache()
     },
   }
 }
--- a/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
+++ b/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
@@ -1,0 +1,44 @@
+import fs from 'node:fs'
+import path from 'node:path'
+
+import {parse as parseDotenv} from 'dotenv'
+
+import {debug} from '../../_exports/debug.js'
+
+/** User config key holding minted-but-unclaimed projects, keyed by project id. */
+export const UNCLAIMED_PROJECTS_CONFIG_KEY = 'unclaimedProjects'
+
+/**
+ * Read a single key straight from the directory's `.env`, using dotenv's grammar so it matches
+ * `readEnvValues` (used by mint/init/logout/nudges) exactly. Unlike Vite's `loadEnv`, it ignores
+ * `process.env`, so a shell `export` from another project can't shadow this directory's `.env`.
+ */
+function readEnvFileValue(cwd: string, key: string): string | undefined {
+  const envPath = path.join(cwd, '.env')
+  if (!fs.existsSync(envPath)) return undefined
+  return parseDotenv(fs.readFileSync(envPath, 'utf8'))[key]?.trim() || undefined
+}
+
+/**
+ * Resolve the robot token for the minted project the current directory points at, given the
+ * ledger `records` (the `unclaimedProjects` config value). Lets the CLI authenticate in a freshly
+ * minted directory that has no config file — where env injection never runs. The token is owned
+ * by the ledger, not `.env`: `.env` only carries the non-secret project id. Never throws.
+ */
+export function resolveMintedProjectToken(
+  records: unknown,
+  cwd: string = process.cwd(),
+): string | undefined {
+  try {
+    if (!records || typeof records !== 'object') return undefined
+
+    const projectId = readEnvFileValue(cwd, 'SANITY_PROJECT_ID')
+    if (!projectId) return undefined
+
+    const record = (records as Record<string, {token?: unknown}>)[projectId]
+    return typeof record?.token === 'string' ? record.token : undefined
+  } catch (err) {
+    debug('failed to resolve minted project token: %s', err)
+    return undefined
+  }
+}
--- a/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
+++ b/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
@@ -27,27 +27,34 @@ export interface MintedProjectCredential {
 /**
  * Resolve the robot credential for the minted project the current directory points at, given the
  * ledger `records` (the `unclaimedProjects` config value). Lets the CLI authenticate in a freshly
- * minted directory that has no config file — where env injection never runs. Mint writes the same
- * token into `.env` for the app runtime (which is why `.env` must stay gitignored); the CLI
- * resolves it from the ledger instead because env injection never runs before a config file
- * exists, and `.env` is user-mutable. Never throws.
+ * minted directory that has no config file — where env injection never runs. The ledger remains
+ * authoritative when it has a usable token; the root `.env` token is the recovery path when the
+ * ledger write failed. Mint writes both values before scaffolding, and `.env` must stay gitignored.
+ * Never throws.
  */
 export function resolveMintedProjectCredential(
   records: unknown,
   cwd: string = process.cwd(),
 ): MintedProjectCredential | undefined {
   try {
-    if (!records || typeof records !== 'object') return undefined
-
     const projectId = readEnvFileValue(cwd, 'SANITY_PROJECT_ID')
     if (!projectId) return undefined
 
-    const record = (records as Record<string, {token?: unknown}>)[projectId]
-    const token = record?.token
-    // A blank token cannot authenticate anything; treat the record as invalid so resolution
-    // falls through to the stored login session instead of surfacing an unusable credential.
-    if (typeof token !== 'string' || !token.trim()) return undefined
-    return {projectId, token}
+    const record =
+      records && typeof records === 'object'
+        ? (records as Record<string, {token?: unknown}>)[projectId]
+        : undefined
+    const recordToken = record?.token
+    if (typeof recordToken === 'string' && recordToken.trim()) {
+      return {projectId, token: recordToken}
+    }
+
+    // Mint can still persist the project-local credential if its separate user-config ledger
+    // write fails. Read only this directory's `.env` (never process.env or nested Studio env)
+    // so commands at the mint root can authenticate without silently taking another project's
+    // shell export.
+    const envToken = readEnvFileValue(cwd, 'SANITY_AUTH_TOKEN')
+    return envToken ? {projectId, token: envToken} : undefined
   } catch (err) {
     debug('failed to resolve minted project credential: %s', err)
     return undefined
--- a/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
+++ b/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
@@ -19,18 +19,23 @@ function readEnvFileValue(cwd: string, key: string): string | undefined {
   return parseDotenv(fs.readFileSync(envPath, 'utf8'))[key]?.trim() || undefined
 }
 
+export interface MintedProjectCredential {
+  projectId: string
+  token: string
+}
+
 /**
- * Resolve the robot token for the minted project the current directory points at, given the
+ * Resolve the robot credential for the minted project the current directory points at, given the
  * ledger `records` (the `unclaimedProjects` config value). Lets the CLI authenticate in a freshly
  * minted directory that has no config file — where env injection never runs. Mint writes the same
  * token into `.env` for the app runtime (which is why `.env` must stay gitignored); the CLI
  * resolves it from the ledger instead because env injection never runs before a config file
  * exists, and `.env` is user-mutable. Never throws.
  */
-export function resolveMintedProjectToken(
+export function resolveMintedProjectCredential(
   records: unknown,
   cwd: string = process.cwd(),
-): string | undefined {
+): MintedProjectCredential | undefined {
   try {
     if (!records || typeof records !== 'object') return undefined
 
@@ -38,9 +43,23 @@ export function resolveMintedProjectToken(
     if (!projectId) return undefined
 
     const record = (records as Record<string, {token?: unknown}>)[projectId]
-    return typeof record?.token === 'string' ? record.token : undefined
+    const token = record?.token
+    // A blank token cannot authenticate anything; treat the record as invalid so resolution
+    // falls through to the stored login session instead of surfacing an unusable credential.
+    if (typeof token !== 'string' || !token.trim()) return undefined
+    return {projectId, token}
   } catch (err) {
-    debug('failed to resolve minted project token: %s', err)
+    debug('failed to resolve minted project credential: %s', err)
     return undefined
   }
+}
+
+/**
+ * Token-only view of {@link resolveMintedProjectCredential}, kept for existing consumers.
+ */
+export function resolveMintedProjectToken(
+  records: unknown,
+  cwd: string = process.cwd(),
+): string | undefined {
+  return resolveMintedProjectCredential(records, cwd)?.token
 }
--- a/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
+++ b/packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
@@ -8,15 +8,53 @@ import {debug} from '../../_exports/debug.js'
 /** User config key holding minted-but-unclaimed projects, keyed by project id. */
 export const UNCLAIMED_PROJECTS_CONFIG_KEY = 'unclaimedProjects'
 
+const MINTED_CREDENTIAL_BOUNDARY_KEYS = [
+  'SANITY_AUTH_TOKEN',
+  'SANITY_CLAIM_URL',
+  'SANITY_PROJECT_ID',
+] as const
+
+function mentionsMintedCredentialBoundary(contents: string): boolean {
+  return contents.split(/\r?\n/u).some((line) => {
+    let statement = line.trimStart()
+    if (!statement || statement.startsWith('#')) return false
+
+    statement = statement.replace(/^export\s+/u, '')
+    return MINTED_CREDENTIAL_BOUNDARY_KEYS.some((key) => {
+      if (!statement.startsWith(key)) return false
+      const nextCharacter = statement[key.length]
+      return nextCharacter === undefined || !/[A-Z0-9_]/iu.test(nextCharacter)
+    })
+  })
+}
+
+function isMissingFileError(error: unknown): boolean {
+  return error instanceof Error && 'code' in error && error.code === 'ENOENT'
+}
+
 /**
- * Read a single key straight from the directory's `.env`, using dotenv's grammar so it matches
- * `readEnvValues` (used by mint/init/logout/nudges) exactly. Unlike Vite's `loadEnv`, it ignores
- * `process.env`, so a shell `export` from another project can't shadow this directory's `.env`.
+ * Find the nearest ancestor `.env` that establishes a Sanity credential boundary. The walk stays
+ * lexical, parses only the selected file, and never consults or mutates `process.env`.
  */
-function readEnvFileValue(cwd: string, key: string): string | undefined {
-  const envPath = path.join(cwd, '.env')
-  if (!fs.existsSync(envPath)) return undefined
-  return parseDotenv(fs.readFileSync(envPath, 'utf8'))[key]?.trim() || undefined
+function readMintedProjectEnv(cwd: string): Record<string, string> | undefined {
+  let directory = path.resolve(cwd)
+
+  while (true) {
+    let contents: string | undefined
+    try {
+      contents = fs.readFileSync(path.join(directory, '.env'), 'utf8')
+    } catch (error) {
+      if (!isMissingFileError(error)) return undefined
+    }
+
+    if (contents !== undefined && mentionsMintedCredentialBoundary(contents)) {
+      return parseDotenv(contents)
+    }
+
+    const parent = path.dirname(directory)
+    if (parent === directory) return undefined
+    directory = parent
+  }
 }
 
 export interface MintedProjectCredential {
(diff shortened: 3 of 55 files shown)


shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/logout.ts
```diff
@@ -15,14 +15,23 @@ export class LogoutCommand extends SanityCommand<typeof LogoutCommand> {
...
+    if (envToken) {
+      this.warn(
+        'SANITY_AUTH_TOKEN is set in the environment (often via ./.env) — logging out cannot end it. Remove that variable to stop acting as its identity.',
+      )
+    }
+
+    // Target the stored login session directly to avoid sending env token to session endpoint.
+    const sessionToken = getCliUserConfig('authToken')
+    if (!sessionToken) {
+      if (!envToken) this.log('No login credentials found')
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 7ca35306: logout now reads the ledger for the current directory and warns that the robot identity survives logout, and no longer reports "no credentials" when a ledger identity exists.

packages/@sanity/cli/src/actions/init/initAction.ts
```diff
@@ -390,19 +403,32 @@ async function ensureAuthenticated(
...
+    ])
+    throw new InitError(
+      SANITY_PROJECT_ID
+        ? `This directory has an unclaimed Sanity project (${SANITY_PROJECT_ID}) but its token isn't ` +
+            'available here. Set SANITY_AUTH_TOKEN from its .env, or claim the project, then re-run.'
+        : 'Not logged in. Run `sanity login` to authenticate, then re-run this command. ' +
+            'Alternatively, run `sanity new` to create a project without logging in — ' +
+            'you can claim it with a Sanity account later. See `sanity new --help`.',
+      exitCodes.RUNTIME_ERROR,
+    )
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 7ca35306: init only calls it an unclaimed mint when the local ledger has a record; a bare SANITY_PROJECT_ID (also written by `sanity init --env`, and surviving a claim) falls back to generic guidance.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,347 @@
...
+          if (ensureEnvGitignored(process.cwd()).added) {
+            flow.line(styleText('dim', 'Added .env to .gitignore so the token stays out of git.'))
+          }
+        }
+        if (written.skippedKeys.length > 0) {
+          flow.note(`./.env already has ${written.skippedKeys.join(', ')} — make sure they read:`)
+          for (const key of written.skippedKeys) {
+            flow.line(`${key}="${envValues[key as keyof typeof envValues]}"`)
+          }
+        }
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 7ca35306 (runs whether keys were written or already present) and hardened in 569b6617 (warns when the gitignore write fails so the token is never silently committable).

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,350 @@
...
+        if (gitignored) {
+          flow.line(styleText('dim', 'Added .env to .gitignore so the token stays out of git.'))
+        }
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 569b6617: ensureEnvGitignored now reports {added, ignored}; mint warns when the gitignore write fails, so a token in .env is never silently committable.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,350 @@
...
+      } catch (err) {
+        this.warn(`Couldn't write ./.env (${err instanceof Error ? err.message : err})`)
+        flow.highlight('Save these credentials yourself — they are not stored anywhere else:')
+        printEnvValues()
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 569b6617: the token and claim details remain recoverable from the local ledger, so the .env-write-failure copy no longer says they are stored nowhere else.

packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
```diff
@@ -0,0 +1,30 @@
...
+  cwd: string = process.cwd(),
+): string | undefined {
+  try {
+    if (!records || typeof records !== 'object') return undefined
+
+    const projectId = loadEnv('development', cwd, 'SANITY_PROJECT_ID').SANITY_PROJECT_ID?.trim()
+    if (!projectId) return undefined
+
+    const record = (records as Record<string, {token?: unknown}>)[projectId]
+    return typeof record?.token === 'string' ? record.token : undefined
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 8bcc2ef1: resolveMintedProjectToken reads the directory .env file directly instead of Vite loadEnv, so a shell-exported SANITY_PROJECT_ID cannot select another project ledger token.

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,252 @@
...
+      claimUrl: minted.claimUrl,
+      expiresAt: minted.expiresAt,
+      mintedAt: new Date().toISOString(),
+      projectId: minted.resourceId,
+      token: minted.token,
+    }
+    writeRecords(records)
+  } catch (err) {
+    debug('failed to record minted project: %s', err)
+  }
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 65efdb7e: recordMintedProject now returns success and mint warns when the ledger write fails — important because a bare minted directory authenticates via the ledger (no auto-injected .env), so a silent failure must not hide behind mint success.

packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
```diff
@@ -0,0 +1,58 @@
...
+
+    let value = unprefixed.slice(eq + 1).trim()
+    // Strip a matching pair of surrounding quotes.
+    if (value.length >= 2 && (value[0] === '"' || value[0] === "'") && value.at(-1) === value[0]) {
+      value = value.slice(1, -1)
+    }
+    return value.trim() || undefined
+  }
+  return undefined
+}
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 65efdb7e: resolveMintedProjectToken now parses .env with dotenv (the same grammar readEnvValues uses), so commented/duplicate SANITY_PROJECT_ID lines resolve identically everywhere — no more divergent first-match parser.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,259 @@
...
+    if (record) {
+      const msLeft = new Date(record.expiresAt).getTime() - now
+      if (msLeft > 0) {
+        const state = await confirm(record)
+        if (state === 'claimed') announceClaimed(record)
+        else if (state === 'expired') announceExpired(record)
+        else announce(renderAmbientLine(record, msLeft))
+      }
+    }
+  }
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 3fe22785: the cwd project is now claim-verified and dropped regardless of the announce slot (only the ambient line waits for the slot), so a sibling project announcing first can no longer leave a claimed cwd project stuck in the ledger and outranking a login session. confirm() is memoized to avoid a second lookup per run.

shapirodaniel:
bugbot run

packages/@sanity/cli/oclif.config.js
```diff
@@ -10,6 +10,7 @@ export default {
       './dist/hooks/prerun/injectEnvVariables.js',
       './dist/hooks/prerun/setupTelemetry.js',
       './dist/hooks/prerun/warnings.js',
+      './dist/hooks/prerun/claimReminders.js',
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in d5cb78ac: getUserConfig now invalidates the in-process token cache on minted-ledger writes (not just authToken), so once claimReminders drops a claimed project the cached robot token is cleared and the rest of the process falls through to the login session — restoring the post-claim handoff.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,367 @@
...
+      minted = await mintUnclaimedProject({displayName})
+    } catch (err) {
+      spin?.fail('Minting your project failed.')
+      throw err
+    }
+    spin?.succeed('Project minted')
+
+    // A `--force`-superseded live project may hold real content.
+    // Keep showing unclaimed nudges until claim or expiry.
+    const recorded = recordMintedProject(minted)
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in fb269bc8: recordMintedProject is now skipped in --json mode, so the "no files written" contract holds (the --json path writes no .env either) and it can no longer seed phantom global claim nudges. The caller owns the token returned in the payload.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,368 @@
...
+          }
+        }
+        if (gitignore.added) {
+          flow.line(styleText('dim', 'Added .env to .gitignore so the token stays out of git.'))
+        } else if (!gitignore.ignored) {
+          // The write failed — .env now holds a token but may not be ignored. Never silent.
+          this.output.warn(
+            "Couldn't add .env to .gitignore — add it yourself so the robot token in .env is never committed.",
+          )
+        }
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 37d5e426: mint now detects an already git-tracked .env (isEnvTracked) and warns to `git rm --cached .env` instead of printing the "added to .gitignore, token is protected" line — since gitignore has no effect on tracked files. (The separate template-side gap, where scaffolded .gitignore ignores only *.local, is tracked for the templates work.)

shapirodaniel:
bugbot run

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,273 @@
...
+  try {
+    const records = readRecords()
+    if (!(projectId in records)) return true
+    delete records[projectId]
+    writeRecords(records)
+    return true
+  } catch (err) {
+    debug('failed to forget minted project: %s', err)
+    return false
+  }
```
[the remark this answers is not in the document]
shapirodaniel: i'm comfortable accepting the tradeoff of possible contention here given that 1/ agents will have read unclaimed project metadata into context before attempting a ledger write, and 2/ modifying global state to guarantee atomic changes is too large a blast radius to warrant for something recoverable as per 1/

deferring until we see legit bug reports

packages/@sanity/cli/src/actions/init/initAction.ts
```diff
@@ -390,19 +404,35 @@ async function ensureAuthenticated(
...
   }
 
   trace.log({step: 'login'})
   output.warn(LOGIN_REQUIRED_MESSAGE)
 
+  renderNewCommandBanner(output)
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 9bf72a62: the interactive path now hoists the same ledger check the unattended path uses and suppresses the "two ways to start" banner when the cwd has a minted project — so it no longer points users at `sanity new`, which the remint guard would refuse there.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/actions/init/initAction.ts
```diff
@@ -390,19 +404,36 @@ async function ensureAuthenticated(
...
+            'you can claim it with a Sanity account later. See `sanity new --help`.',
+      exitCodes.RUNTIME_ERROR,
+    )
   }
 
   trace.log({step: 'login'})
   output.warn(LOGIN_REQUIRED_MESSAGE)
 
+  // Suppressed in a minted directory: `sanity new` would just be refused by the remint guard.
+  if (!mintedRecord) renderNewCommandBanner(output)
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in 1ae656cd: the suppression now keys off the same GUARDED_ENV_KEYS the remint guard uses (moved to util/envFile and shared), so init no longer steers users at `sanity new` whenever any guarded .env key is present — covering ledger-write failure, copied minted dirs, and bare `sanity init --env` leftovers, not just the ledger-record case. The ledger record still selects the more specific "your token is in .env" message.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/__tests__/init/init.authentication.test.ts
```diff
@@ -56,6 +55,12 @@ vi.mock('../../../util/detectFramework.js', () => ({
   }),
 }))
 
+// Isolate from any ambient .env in the working directory: init only reads it to tailor the
+// not-logged-in hint, and a developer's local .env must not steer these auth assertions.
+vi.mock('../../../util/envFile.js', () => ({
+  readEnvValues: vi.fn(() => ({})),
+}))
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in a4269abf: the init auth test file now exports GUARDED_ENV_KEYS from its envFile mock, so the not-logged-in paths produce guidance instead of a TypeError. (Confirmed the other envFile-mocking test, logout, is unaffected — logout.ts does not reference GUARDED_ENV_KEYS.)

shapirodaniel:
bugbot run

packages/@sanity/cli/src/services/mintProject.ts
```diff
@@ -0,0 +1,164 @@
...
+  debug('checking claim state via project host at %s', url)
+
+  try {
+    const response = await fetch(url, {
+      headers: {Authorization: `Bearer ${robotToken}`},
+      signal: AbortSignal.timeout(options?.timeoutMs ?? 1500),
+    })
+    if (response.status === 404) return 'expired'
+    if (response.status === 401) return undefined
+    if (!response.ok) return undefined
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in aae7124d: lookupClaimStateViaProject now returns a distinct `revoked` state on 401 (vs undefined for genuine fail-open network errors), and the nudge pass drops a revoked project from the ledger — so getCliToken falls through to the login session instead of serving the dead token. Surfaced with a "token is no longer valid — run `sanity login`" message.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,323 @@
...
+      'dim',
+      `⏳ Unclaimed Sanity project ${record.projectId} expires in ${humanizeMsLeft(msLeft)} — claim it to keep it:`,
+    ) + `\n${styleText(['cyan', 'underline'], record.claimUrl)}`
+  )
+}
+
+const claimedMessage = (record: UnclaimedProjectRecord): string =>
+  `${logSymbols.success} Sanity project ${record.projectId} has been claimed — it's yours to keep.`
+const expiredMessage = (record: UnclaimedProjectRecord): string =>
+  `⌛ Unclaimed Sanity project ${record.projectId} expired on ${record.expiresAt}. Run \`sanity new\` to mint a new one.`
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in b9f1fbbd: the expiry farewell now points at `sanity new --force`. Since the expired credentials remain in .env, a plain `sanity new` is refused by the remint guard, so --force is the accurate recovery (it mints a replacement and leaves .env for you to update).

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,323 @@
...
+const revokedMessage = (record: UnclaimedProjectRecord): string =>
+  `⚠ Sanity project ${record.projectId}'s token is no longer valid — run \`sanity login\` to continue.`
```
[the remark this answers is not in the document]
shapirodaniel: Addressed in b9f1fbbd: the revoke message now tells the user to remove SANITY_AUTH_TOKEN from .env after `sanity login`. Dropping the ledger record alone is insufficient in a config directory, where .env is auto-injected and the dead token outranks the new session in getCliToken.

shapirodaniel:
bugbot run

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,376 @@
...
+    // Under `--json` oclif suppresses `this.log` and prints the returned result instead, so the
+    // narrated flow only appears in human runs and JSON mode never writes to the filesystem.
+    const {output} = this
+    const json = this.jsonEnabled()
+    const flow = createFlow(output.log)
+    // How this command was invoked (`sanity new` vs `sanity projects mint`).
+    const invoked = [this.config.bin, ...(this.id?.split(':') ?? [])].join(' ')
+
+    const envPath = path.join(process.cwd(), '.env')
+    const guard: GuardResult = await this.guardExistingProject(envPath, invoked)
```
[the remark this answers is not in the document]
shapirodaniel: there's no `.env.local` mint-and-claim credentials path, that said we can def add docs guidance on this to prevent agents from inadvertently writing two sets of creds to `.env` and `.env.local`

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,328 @@
...
+  )
+
+  const urgent = tier >= 3
+  const headline =
+    tier === 1
+      ? `⏳ Claim your Sanity project — ${record.projectId} expires in ${timeLeft}`
+      : tier >= 4
+        ? `🚨 Final reminder — project ${record.projectId} expires in ${timeLeft}`
+        : `${urgent ? '🚨' : '⏰'} Project ${record.projectId} expires in ${timeLeft}`
+  const bell = tier >= 4 ? '' : ''
```
jwoods02: Both branches are empty strings so the final reminder never actually rings the bell. Either `\u0007` in the first branch or delete the variable.

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,328 @@
...
+    return state
+  }
+
+  for (const record of Object.values(records)) {
+    if (new Date(record.expiresAt).getTime() > now) continue
+    const state = await confirm(record)
+    switch (state) {
+      case 'claimable': {
+        continue
+        break
```
jwoods02: Unreachable, the `continue` above already exits the iteration.

packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
```diff
@@ -0,0 +1,44 @@
...
+  const envPath = path.join(cwd, '.env')
+  if (!fs.existsSync(envPath)) return undefined
+  return parseDotenv(fs.readFileSync(envPath, 'utf8'))[key]?.trim() || undefined
+}
+
+/**
+ * Resolve the robot token for the minted project the current directory points at, given the
+ * ledger `records` (the `unclaimedProjects` config value). Lets the CLI authenticate in a freshly
+ * minted directory that has no config file — where env injection never runs. The token is owned
+ * by the ledger, not `.env`: `.env` only carries the non-secret project id. Never throws.
```
jwoods02: This isn't true, mint writes the full robot token into `.env` as `SANITY_AUTH_TOKEN` (mint.ts:200), which is why we gitignore it and warn when it's tracked. However this one is true: `getCliToken` resolves from the ledger rather than `.env` because `.env` is mutable. 

Worth rewording before someone reads this and concludes `.env` needs no secret handling. Same wording appears in the `getCliToken` comment in cliUserConfig.ts.

packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
```diff
@@ -37,6 +38,20 @@ export async function getCliToken(): Promise<string | undefined> {
     return trimmed
   }
 
+  // A minted directory carries only its (non-secret) project id in `.env`; the robot token lives
+  // in the ledger. This resolves it without env injection, which never runs before a config file
+  // exists — the gap that stops `sanity init` authenticating in a freshly minted directory.
+  let mintedToken: string | undefined
+  try {
+    mintedToken = resolveMintedProjectToken(getUserConfig().get(UNCLAIMED_PROJECTS_CONFIG_KEY))
```
jwoods02: Unsure this is the right approach.

What happens today: inside any sanity project, `.env` gets loaded into the environment on every command, and an env `SANITY_AUTH_TOKEN` outranks your login session. That's old behaviour and this PR doesn't touch it. But today you only end up there by putting a token in `.env` yourself, so it's not a surprise.

What this PR adds: `sanity new` writes the robot token into `.env` for you, and this ledger tier makes it win even in a bare directory where env injection doesn't run. So in a minted directory the CLI always prefers the robot token over whoever is actually logged in. Pre-claim that's fair enough, the robot is the only identity that can access the project.

Why I'm unsure it's correct: the preference doesn't really end at claim. The ledger entry gets cleaned up once the claim check succeeds, but the same token is still sitting in `.env`, and once a studio is scaffolded that file loads into the environment on every command. So in the normal mint → init → claim flow you carry on acting as "Editor Token (Robot)" in that directory forever, even after running sanity login, until you hand-delete the line from .env. And nothing tells you. 

Maybe we just need to encode this in docs or the claim message, there could also be issues if we wipe .env for you as well.

packages/@sanity/cli/src/services/mintProject.ts
```diff
@@ -0,0 +1,166 @@
...
+  const base = override
+    ? override.replace(/\/$/, '')
+    : `https://${projectId}.api.sanity.${isStaging() ? 'work' : 'io'}`
+  const url = `${base}/v2026-05-04/projects/${projectId}`
+  debug('checking claim state via project host at %s', url)
+
+  try {
+    const response = await fetch(url, {
+      headers: {Authorization: `Bearer ${robotToken}`},
+      signal: AbortSignal.timeout(options?.timeoutMs ?? 1500),
```
jwoods02: Default is still 1500 here (and on `lookupClaimState` at line 61) but every caller now passes 500 after Wednesday's decision. Worth flipping the defaults to 500 so the next caller doesn't quietly get the old timeout.

packages/@sanity/cli/src/services/mintProject.ts
```diff
@@ -0,0 +1,166 @@
...
+
+  const response = await fetch(url, {
+    body: JSON.stringify({displayName, resourceType: 'project'}),
+    headers: {'Content-Type': 'application/json'},
+    method: 'POST',
+  })
+
+  if (!response.ok) {
+    const body = await response.text().catch(() => '')
+    throw new Error(`Mint failed (HTTP ${response.status}): ${body || response.statusText}`)
```
jwoods02: A 404 here means the kill switch is off but it surfaces as `Mint failed (HTTP 404): ...`. The friendly 'provisioning disabled, or rate limited' message below only fires on a 200 with no claimToken, which isn't how the kill switch responds. Suggest special-casing 404 and 429 with the human message.

packages/@sanity/cli/src/services/mintProject.ts
```diff
@@ -0,0 +1,166 @@
...
+export async function mintUnclaimedProject(options: {displayName: string}): Promise<MintedProject> {
+  const displayName = options.displayName.trim()
+  if (!displayName || displayName.length > 80) {
+    throw new Error('Display name must be 1-80 characters.')
+  }
+
+  const url = `${getProvisionApiBase()}/${PROVISION_API_VERSION}/provision`
+  debug('minting unclaimed project at %s', url)
+
+  const response = await fetch(url, {
```
jwoods02: I think there's a HTTP setup in CLI to use. Bare `fetch` means no proxy support and no CLI user agent, unlike everything else that goes through the client stack. 

Someone behind a corporate proxy will have every sanity command work except `sanity new`, and the server side can't tell CLI mints apart for triage. Fine to keep if it's deliberate for the unauthenticated endpoints, but deserves a comment saying so.

packages/@sanity/cli/src/commands/__tests__/logout.test.ts
```diff
@@ -76,19 +92,66 @@ describe('#logout', () => {
...
+    const {error, stderr, stdout} = await testCommand(LogoutCommand)
+
+    expect(error).toBeUndefined()
+    expect(stderr).toContain('acts as unclaimed Sanity project abc123')
+    // The ledger identity counts as credentials — don't claim there are none.
+    expect(stdout).not.toContain('No login credentials found')
+    expect(mockGetMintedProjectRecord).toHaveBeenCalledWith('abc123')
+  })
+
+  test('env token plus stored session: warns about the env token and ends the session', async () => {
```
jwoods02, line 149: This proves logout succeeds when both tokens are present, but the nock mock matches on URL only, so it doesn't prove it was the session token rather than `SANITY_AUTH_TOKEN` that hit `/auth/logout`. That's the exact leak the change prevents, worth asserting the Authorization header.

jwoods02 (commented):
Some code comments, doing some manual testing also

In terms of functionally sanity new still doesn't scaffold a project and requires you to run sanity init afterwards. I think we should add that here

packages/@sanity/cli/src/util/claimNudges.ts
shapirodaniel: nice, imo let's nuke the bell entirely, it's only a human-accessible thing and it's not super differentiated, terminals ring or click for a number of reasons including user inputs

packages/@sanity/cli/src/util/claimNudges.ts
shapirodaniel: nice catch! removed in 4d3af36b

packages/@sanity/cli/src/util/claimNudges.ts
shapirodaniel: removed in 4d3af36b

packages/@sanity/cli-core/src/config/cli/unclaimedProjects.ts
shapirodaniel: nice catch! reworded in 733d043d

packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
shapirodaniel: >Maybe we just need to encode this in docs or the claim message

totally agree, in 733d043d we're now warning if an `.env` token "outranks" the user's session, the same way we warn at `sanity logout`

in 4d3af36b, the final claim "farewell" message gives similar guidance about the directory being robot auth'd and how to switch to user auth

packages/@sanity/cli/src/services/mintProject.ts
shapirodaniel: nice catch! fixed in d1739c24

packages/@sanity/cli/src/services/mintProject.ts
shapirodaniel: yeah def agree on special-casing, i've added human-friendly copy to reference minting as disabled / rate limit reached in d1739c24

packages/@sanity/cli/src/services/mintProject.ts
shapirodaniel: great catch, and thanks for pointing me to `createRequester`! updated in d1739c24

packages/@sanity/cli/src/commands/__tests__/logout.test.ts
shapirodaniel: great call out, yeah let's def do this -- updated to assert `Authorization` header content in 733d043d

shapirodaniel:
> Some code comments, doing some manual testing also
> 
> In terms of functionally sanity new still doesn't scaffold a project and requires you to run sanity init afterwards. I think we should add that here

in offline sync, we agreed this is def something we want in v1 launch, and it's a good fit for an immediate fast follow, given the stealth launch of `sanity new` and the likely additional scope, which would complicate an already pretty large diff

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,247 @@
...
+    announce(renderReminder(lead.record, lead.msLeft, {solo: false}))
+    announce(renderAggregateReminder(rest))
+  }
+
+  if (touched.size > 0) {
+    const fresh = readRecords()
+    for (const id of touched) {
+      if (id in records) fresh[id] = records[id]
+      else delete fresh[id]
+    }
```
jwoods02: Tiny leftover from the tier rework: `touched` is only populated by `drop()`, which deletes the id from `records` in the same breath, so `id in records` can never be true here. The copy branch was live in the old code when the tier path bumped `lastNudgeTier` on a record it was keeping, but that path is gone now.

```suggestion
    for (const id of touched) delete fresh[id]
```

(`touched` could honestly be renamed `dropped` at that point, since that's all it tracks now.)

shapirodaniel: nice call out, this was def vestigial, fixed in 8103d4af 🙏

packages/@sanity/cli-core/src/_exports/index.ts
```diff
@@ -24,6 +24,10 @@ export {
   type ConfigStore,
 } from '../config/cli/types/cliConfig.js'
 export {type UserViteConfig} from '../config/cli/types/userViteConfig.js'
+export {
+  resolveMintedProjectToken,
+  UNCLAIMED_PROJECTS_CONFIG_KEY,
+} from '../config/cli/unclaimedProjects.js'
```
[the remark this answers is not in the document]
shapirodaniel: nice one bugbot, fixed in 8103d4af

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

shapirodaniel:
bugbot run

packages/@sanity/cli/src/util/claimNudges.ts
```diff
@@ -0,0 +1,247 @@
...
+      }
+      case 'revoked': {
+        announce(revokedMessage(record))
+        drop(record.projectId)
+        break
+      }
+      default: {
+        if (msLeft > 0) {
+          live.push({msLeft, record})
+        } else if (state !== 'claimable') {
```
jwoods02: Coming back to this one because it's now a decision rather than an oversight, and a test pins it. `state !== 'claimable'` means a failed lookup (500ms timeout, flaky wifi, any non-2xx) on a record past its local expiry gets dropped from the ledger and announced as deleted, so the claim URL and token go with it.

Worth noting the mint guard takes the opposite stance on the same uncertainty: `mint.test.ts` has "lookup failure never authorizes the expired auto-proceed". Same unknown, two philosophies. I'd rather we fail closed here too, only dropping on a confirmed `expired` or `revoked`, and soften the copy when we genuinely don't know. If you'd rather keep it, that's fine, but let's say so out loud, because the test currently reads as intent.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,571 @@
...
+          flow.line(
+            `Those names follow the Next.js convention. On ${scaffold.detectedFramework}, use its ` +
+              'own public prefix for the project id and dataset.',
+          )
+        }
+      }
+
+      if (scaffold.frontendPath) {
+        flow.highlight(`Created ./${STUDIO_DIR} (Studio) and ./${FRONTEND_DIR} (frontend).`)
+        flow.line(`cd ${STUDIO_DIR} && npx sanity dev    to start the Studio on port 3333`)
```
jwoods02: This prints the dev command but not the way in. A pre-claim Studio has no browser session, so `localhost:3333` is just a login screen unless you open `http://localhost:3333/#token=<token>`, which the sanity.new agent instructions call our as the designed path. Anyone following only the CLI's output gets stuck there with no hint.

One more line here would close it (same for the no-frontend branch at 437)

packages/@sanity/cli/src/actions/scaffold/createFrontend.ts
```diff
@@ -0,0 +1,139 @@
...
+  debug('Scaffolding frontend: npx %s', args.join(' '))
+
+  const progress = spinner({
+    discardStdin: !cancelSignal,
+    text: 'Creating your Next.js app\n',
+  }).start()
+  let result
+  try {
+    result = await execa('npx', args, {
+      cancelSignal,
```
jwoods02, line 83: The cancellation plumbing is good but there's no timeout, here or in `installPackages`, and a cancel signal doesn't help with a stalled tarball fetch that never returns: `sanity new` just sits there with the project already minted and the 72 hour clock running.

The claim URL also only reaches the terminal after the scaffold (mint.ts:462). It does go into the `.env` first, which is a decent safety net, but I think we should print it pre scaffold

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,571 @@
...
+            'directory authenticate, so scaffolding without it would leave a project that cannot ' +
+            'read its own content. Write ./.env, then scaffold with:',
+        )
+        printScaffoldRecipe()
+      }
+      flow.gap()
+    }
+
+    const shouldScaffold =
+      !json && this.flags.scaffold && !guard.hasExistingKeys && credentialsOnDisk
```
jwoods02, line 479: Nits, grouped, all optional:

- `!guard.hasExistingKeys` here means `--force` in a directory with no scaffold mints and then never scaffolds, and the note only mentions `.env` being left alone. Right for the expired lane, a bit silent for `--force`.
- `mint.test.ts:51` mocks `TOKEN_ENV_FILES` as `'./.env, or sanity/.env.local in a scaffolded project'` but the real value is `'./.env or sanity/.env.local'`, so the copy in the command description and the `CLAIMED_PROJECT_IN_ENV` suggestion isn't actually pinned by anything.
- `initStudio` still prints its own "Success! / Get started by running ..." block with an absolute path, mid-narration, right before we print different commands for the same thing. `preclaim` is already threaded in to suppress the `sanity manage` line, so the same seam could suppress the whole block when a caller owns the outro.
- We pin `next-sanity@13` when should probably default to latest here? We control this package

jwoods02 (commented):
Only minor things now!

packages/@sanity/cli/src/util/claimNudges.ts
shapirodaniel: >I'd rather we fail closed here too, only dropping on a confirmed expired or revoked, and soften the copy when we genuinely don't know

agreed, we're now retaining creds / claim url in all but confirmed terminal states, fixed in 52f0135b

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: nice catch, i've added the `#token=<token>` hash fragment to all studio url outputs (scaffold and no-scaffold), fixed in d4a775be

packages/@sanity/cli/src/actions/scaffold/createFrontend.ts
shapirodaniel: nice call out / catch on the possibility of `sanity new` hanging indefinitely on `npx` package download blips, and yeah agree printing the claim url ahead of scaffolding makes sense: fixed in 07783276, d4a775be

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: these were all super helpful call outs! addressed in 07783276, d4a775be

packages/@sanity/cli/src/commands/new.ts
```diff
@@ -0,0 +1,10 @@
+import {MintProjectCommand} from './projects/mint.js'
+
+/**
+ * Top-level alias for `sanity projects mint`.
+ * Subclassed to share implementation and give `sanity new` its own entry in help output.
+ */
+export class NewCommand extends MintProjectCommand {
+  // Don't inherit the parent's `project:mint` alias — it must resolve to a single command.
+  static override hiddenAliases: string[] = []
+}
```
jwoods02, line 80: **Blocking.** `sanity new --help` currently prints `projects mint`'s help text, because this class overrides only `hiddenAliases`. So the first thing a newcomer reads is "Mint an unclaimed Sanity project" — our word, not theirs — followed by a line about "the scaffolded Studio folder" before anything has said a scaffold happens, and five examples that all start "Mint …".

Low traffic, since an agent usually runs the command. But it is the one surface a curious human goes looking for, and the UAT guide treats its content as a deliverable.

Giving `sanity new` its own description rather than sharing a body: these are different documents now, one for a first-timer and one for the plumbing. Costs a duplicated "72 hours" — see the review body about a shared const.

Cut, on the basis that the docs page covers it: the `--project-id` resolution mechanics, "robot token", the grant list, `NEXT_PUBLIC_*`/`SANITY_STUDIO_*` as literals, and the `--json` credentials paragraph. The security point stays, in plainer form — being vague there could cost someone real money.

```suggestion
import {Flags} from '@oclif/core'

import {FRONTEND_DIR, STUDIO_DIR} from '../actions/scaffold/scaffoldProject.js'
import {MintProjectCommand} from './projects/mint.js'

/**
 * Top-level alias for `sanity projects mint`.
 * Subclassed to share implementation and give `sanity new` its own entry in help output.
 *
 * Help text is deliberately not shared with the parent: `projects mint` is the plumbing command
 * and can stay technical, while `sanity new` is the door most people arrive at.
 */
export class NewCommand extends MintProjectCommand {
  static override description =
    'Create a Sanity project without an account, and claim it within 72 hours to keep it.\n' +
    '\n' +
    `Sets up two folders here: ./${STUDIO_DIR}, a Studio where you write and edit your content, ` +
    `and ./${FRONTEND_DIR}, a Next.js website that reads it. Both are already connected to your ` +
    'new project, so you can start them straight away. Use --no-scaffold if you just want the ' +
    'project and nothing else.\n' +
    '\n' +
    'The project is real and works immediately, but it is only yours for 72 hours. Claim it with ' +
    'a Sanity account before the deadline and everything you have built stays exactly as it is. ' +
    'Claiming is free and takes about a minute. Miss the deadline and the project and its ' +
    'content are deleted.\n' +
    '\n' +
    'Two things to keep private. The claim link, because anyone who opens it becomes the owner. ' +
    'And the access token saved in ./.env, because it can read and change everything in the ' +
    'project - .env is added to .gitignore for you, and the token must never go into code that ' +
    'runs in the browser.\n' +
    '\n' +
    'Fetch https://sanity.new for full instructions, or point your AI agent at it.'

  static override examples = [
    {
      command: '<%= config.bin %> <%= command.id %>',
      description: 'Create a project with a Studio and a website',
    },
    {
      command: '<%= config.bin %> <%= command.id %> "My New Project"',
      description: 'Create a project called "My New Project"',
    },
    {
      command: '<%= config.bin %> <%= command.id %> --yes',
      description: 'Create a project without being asked anything',
    },
    {
      command: '<%= config.bin %> <%= command.id %> --no-scaffold',
      description: 'Create the project only, with no Studio or website',
    },
    {
      command: '<%= config.bin %> <%= command.id %> --json',
      description: 'Create a project and print its details as JSON',
    },
  ]

  static override flags = {
    ...MintProjectCommand.flags,
    scaffold: Flags.boolean({
      allowNo: true,
      default: true,
      description: `Set up a Studio in ./${STUDIO_DIR} and a Next.js website in ./${FRONTEND_DIR} (on by default)`,
    }),
  }

  // Don't inherit the parent's `project:mint` alias — it must resolve to a single command.
  static override hiddenAliases: string[] = []
}
```

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,607 @@
...
+    'The robot token has full content access to this project: create, edit, publish, and deploy ' +
+    'schemas. It cannot deploy a hosted Studio, create datasets, or manage settings. The dataset ' +
+    'is private pre-claim: frontend reads must run server-side, and the token must never sit ' +
+    'under a client-exposed prefix like NEXT_PUBLIC_* or SANITY_STUDIO_*.\n' +
+    '\n' +
+    `After the claim, run \`sanity login\` and remove SANITY_AUTH_TOKEN from ${TOKEN_ENV_FILES} ` +
+    'to act as your own account; until then, CLI commands in this directory keep ' +
+    'authenticating as the robot.\n' +
+    '\n' +
+    'Minting is rate-limited per machine. Fetch https://sanity.new for agent instructions.'
```
jwoods02: Two problems. A reader is told to "run them inside the scaffolded Studio folder" before anything mentions a scaffold, and `--no-scaffold` — now the only route to a non-Next.js frontend — goes unmentioned. Separately the dataset sentence stops one clause short: claiming makes the dataset public, so someone reading only this would assume privacy is the steady state and be surprised when their content becomes world-readable.

Keeping this one technical; it is the plumbing command and that register is right. I have shortened the prefix guidance rather than expanding it, since the docs page carries the per-framework detail.

```suggestion
  static override description =
    'Mint an unclaimed Sanity project without logging in.\n' +
    '\n' +
    `By default this also scaffolds a Studio into ./${STUDIO_DIR} and a Next.js frontend into ` +
    `./${FRONTEND_DIR}, with credentials wired into both; pass --no-scaffold to skip it. ` +
    'Credentials are written to ./.env (SANITY_PROJECT_ID, SANITY_DATASET, and ' +
    'SANITY_AUTH_TOKEN, a robot token) and .env is gitignored, so `sanity` commands run from ' +
    'here authenticate as the project with no account. Commands that need a project id read it ' +
    `from a Sanity config file or --project-id, so run those from ./${STUDIO_DIR}. Use --json ` +
    'for a machine-readable payload: it writes no files, scaffolds nothing, and the caller owns ' +
    'the credentials.\n' +
    '\n' +
    'Claiming: the project must be claimed with a Sanity account within 72 hours (expiresAt) or ' +
    'it is permanently deleted, content included. The claim URL is single-use and whoever opens ' +
    "it takes ownership: keep it out of git and shared channels. If you're an agent, surface the link to the end user " +
    "immediately. Everything keeps working after you've claimed, including the robot token.\n" +
    '\n' +
    'The robot token has full content access to this project: create, edit, publish, and deploy ' +
    'schemas. It cannot deploy a hosted Studio, create datasets, or manage settings. The dataset ' +
    'is private until you claim, so frontend reads must run server-side; claiming makes it ' +
    'public. Wherever it lives, the token must never sit under a browser-exposed prefix.\n' +
    '\n' +
    `After the claim, run \`sanity login\` and remove SANITY_AUTH_TOKEN from ${TOKEN_ENV_FILES} ` +
    'to act as your own account; until then, CLI commands in this directory keep ' +
    'authenticating as the robot.\n' +
    '\n' +
    'Minting is rate-limited per machine. Fetch https://sanity.new for agent instructions.'
```

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,607 @@
...
+    flow.line("Project credentials keep working after you've claimed.")
+    flow.line(hyperlink(styleText('cyan', minted.claimUrl), minted.claimUrl))
+    flow.gap()
+    flow.note(
+      styleText('dim', 'If an agent is running this, surface the full claim URL to the user.'),
+    )
+    flow.line(
+      styleText('dim', 'They need to click this link to claim the project before it expires.'),
+    )
+    flow.gap()
```
jwoods02: Not suggesting we drop the agent instruction — an agent can run this without having fetched sanity.new, so it needs telling. Two smaller problems:

1. A human running this reads about themselves in the third person: "surface the full claim URL **to the user**", "**They** need to click this link". You have to work out that "they" means you.
2. It takes two lines to say one thing.

Putting the condition first means a human knows in three words that it isn't aimed at them, and an agent still gets the instruction:

```suggestion
    flow.note(
      styleText(
        'dim',
        'If you are an agent: give this claim URL to the person you are working for. They have ' +
          'to open it themselves before the deadline.',
      ),
    )

File: packages/@sanity/cli/src/commands/projects/mint.ts [hunk 38]
```diff
@@ -0,0 +1,607 @@
...
+      const printFrontendEnv = () => {
+        for (const [key, value] of Object.entries(scaffold.frontendEnv)) {
+          flow.line(`${key}="${value}"`)
+        }
+        if (scaffold.detectedFramework && scaffold.detectedFramework !== 'Next.js') {
+          flow.line(
+            `Those names follow the Next.js convention. On ${scaffold.detectedFramework}, use its ` +
+              'own public prefix for the project id and dataset.',
+          )
+        }
```
jwoods02: **Blocking.** An Astro user gets Next.js variable names and is then asked to translate them:

```
Found Astro here, so only ./sanity was created.
Add these to your app:
NEXT_PUBLIC_SANITY_DATASET="…"
NEXT_PUBLIC_SANITY_PROJECT_ID="…"
Those names follow the Next.js convention. On Astro, use its own public prefix …
```

We know the framework at this point, so we are printing values we know are wrong. See the `scaffoldProject.ts` comments for deriving the prefix — with a caveat: `envPrefix` from `@vercel/frameworks` is set on only ~15 of 72 frameworks, Nuxt's value predates `runtimeConfig`, and current SvelteKit has none. So it cannot be trusted blindly.

Once the prefix is available, this line becomes the honest fallback for frameworks that genuinely have no convention (Remix, Angular) rather than the default:

```suggestion
        if (scaffold.detectedFramework && !scaffold.frontendEnvPrefix) {
          flow.line(
            `Those names follow the Next.js convention. On ${scaffold.detectedFramework}, use its ` +
              'own public prefix for the project id and dataset.',
          )
        }
```

Worth a test: `detectedFramework: 'Astro'` asserts `PUBLIC_SANITY_PROJECT_ID`.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,607 @@
...
+        flow.line('Scaffold a frontend yourself, then add these:')
+        printFrontendEnv()
+      }
+      flow.gap()
+      flow.note(
+        'Your dataset is private until you claim it, so reading content from your frontend needs a ' +
+          'token. Copy SANITY_AUTH_TOKEN from ./.env when you add those reads, keep them ' +
+          'server-side, and never put it under a browser-exposed prefix like NEXT_PUBLIC_ or ' +
+          'SANITY_STUDIO_: that publishes a credential with full write access.',
+      )
```
jwoods02: This is 60 words of prefix guidance arriving before the user has written a line of code, and it is the single longest line in the output — at 80 columns it wraps into four rows with no rail, which is most of why the end of the run looks broken.

Since docs now covers the prefix rules per framework, I would cut rather than extend. Keep the two facts that are actionable here (the token is needed for reads, and it belongs on the server), add the part people are surprised by (claiming makes it public), and let docs carry the rest:

```suggestion
      flow.note(
        'Your content is private until you claim, so anything reading it needs the token from ' +
          './.env. Keep those reads server-side and never expose the token to the browser: it ' +
          'can change everything in this project. Claiming makes your content readable without ' +
          'it.',
      )
```

Then a docs link in the outro. That is ~40 words instead of ~60, drops the framework literals, and never says "prefix".

packages/@sanity/cli/src/actions/scaffold/scaffoldProject.ts
```diff
@@ -0,0 +1,243 @@
...
+export interface ScaffoldResult {
+  frontendEnv: Record<string, string>
+  frontendEnvWritten: boolean
+  studioPath: string
+  warnings: string[]
+
+  detectedFramework?: string
+  frontendPackageManager?: PackageManager
+  frontendPath?: string
+}
```
jwoods02, line 67: Part 1 of 3 for the framework-prefix fix (see `mint.ts:453`).

`detectFrameworkRecord` returns the whole `Framework` record, so `envPrefix` is available — but checked against `@vercel/frameworks@3.29.0` as resolved here, coverage is patchy: only 15 of 72 frameworks define it, `nuxtjs` gives `NUXT_ENV_` (pre-`runtimeConfig`; modern Nuxt uses `NUXT_PUBLIC_`), and `sveltekit-1` — current SvelteKit — has none. Astro (`PUBLIC_`) and Vite (`VITE_`) are correct.

So: a small override table for the wrong-or-missing ones, `envPrefix` for the rest, `undefined` meaning "no known convention" so the caller can fall back to today's message.

```suggestion
// `envPrefix` from @vercel/frameworks is set on only ~15 of 72 frameworks, and Nuxt's value
// predates runtimeConfig. Override the ones users are most likely to reach for; `undefined`
// means "no public-prefix convention", and the caller explains the situation instead.
const PUBLIC_ENV_PREFIX_OVERRIDES: Record<string, string> = {
  nuxtjs: 'NUXT_PUBLIC_',
  'sveltekit-1': 'PUBLIC_',
}

export interface ScaffoldResult {
  frontendEnv: Record<string, string>
  frontendEnvWritten: boolean
  studioPath: string
  warnings: string[]

  detectedFramework?: string
  frontendEnvPrefix?: string
  frontendPackageManager?: PackageManager
  frontendPath?: string
}
```

This pins us to `@vercel/frameworks`' slugs — if that dep floats, the override table is what protects us.

packages/@sanity/cli/src/actions/scaffold/scaffoldProject.ts
```diff
@@ -0,0 +1,243 @@
...
+  if (detected) {
+    debug('Detected %s in %s, leaving the frontend alone', detected.name, workDir)
+    return {
+      detectedFramework: detected.name,
+      frontendEnv,
+      frontendEnvWritten: false,
+      studioPath,
+      warnings,
+    }
+  }
```
jwoods02: Part 2 of 3. Derive the keys from the resolved prefix instead of hardcoding `NEXT_PUBLIC_`, and pass the prefix out so `mint.ts` knows whether it needs to apologise.

Net effect: Astro gets `PUBLIC_*`, Nuxt `NUXT_PUBLIC_*`, Vite and SvelteKit the right thing, greenfield stays `NEXT_PUBLIC_*` (we scaffold Next, so that is correct), and Remix or Angular still get the honest "translate this yourself" line.

```suggestion
  const frontendEnvPrefix = detected
    ? (PUBLIC_ENV_PREFIX_OVERRIDES[detected.slug] ?? detected.envPrefix)
    : 'NEXT_PUBLIC_'

  const frontendEnv = {
    [`${frontendEnvPrefix ?? 'NEXT_PUBLIC_'}SANITY_DATASET`]: dataset,
    [`${frontendEnvPrefix ?? 'NEXT_PUBLIC_'}SANITY_PROJECT_ID`]: projectId,
  }

  if (detected) {
    debug('Detected %s in %s, leaving the frontend alone', detected.name, workDir)
    return {
      detectedFramework: detected.name,
      frontendEnv,
      frontendEnvPrefix,
      frontendEnvWritten: false,
      studioPath,
      warnings,
    }
  }
```

packages/@sanity/cli/src/actions/scaffold/scaffoldProject.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  return {
+    frontendEnv,
+    frontendEnvWritten,
+    frontendPackageManager: resolvedPackageManager,
+    frontendPath,
+    studioPath,
+    warnings,
+  }
```
jwoods02, line 256: Part 3 of 3 — carry `frontendEnvPrefix` out of the scaffolded branch too, so the field is populated on both paths.

```suggestion
  return {
    frontendEnv,
    frontendEnvPrefix,
    frontendEnvWritten,
    frontendPackageManager: resolvedPackageManager,
    frontendPath,
    studioPath,
    warnings,
  }
```

packages/@sanity/cli/src/services/mintProject.ts
```diff
@@ -0,0 +1,188 @@
...
+    url,
+  })
+
+  if (response.statusCode === 404) {
+    throw new Error(
+      'Minting new projects is currently disabled. Try again later, or run `sanity login` and `sanity init` to create a project.',
+    )
+  }
+  if (response.statusCode === 429) {
+    throw new Error('Mint rate limit reached for this machine. Try again in an hour.')
```
jwoods02: Naming the window invites someone to time around it, and it dates badly if the number ever changes. Saying a limit exists is the useful half.

```suggestion
    throw new Error('Mint rate limit reached for this machine. Try again later.')
```

`mint.ts:127` is fine as it stands.

packages/@sanity/cli/src/actions/init/__tests__/initAction.test.ts
```diff
@@ -70,6 +79,16 @@ vi.mock('../../auth/login/login.js', () => ({
   login: mockLogin,
 }))
 
+vi.mock('../../../util/envFile.js', () => ({
+  GUARDED_ENV_KEYS: ['SANITY_AUTH_TOKEN', 'SANITY_PROJECT_ID', 'SANITY_CLAIM_URL'],
+  inspectEnvKeys: mockInspectEnvKeys,
+  TOKEN_ENV_FILES: './.env, or sanity/.env.local in a scaffolded project',
```
jwoods02: This stubs `TOKEN_ENV_FILES` as `'./.env, or sanity/.env.local in a scaffolded project'`, but the real constant (`envFile.ts:19`) is `'./.env or sanity/.env.local'`.

That string is interpolated into the command description, the `CLAIMED_PROJECT_IN_ENV` suggestion, two claim reminders, and both the login and logout warnings — six user-facing surfaces, none of them asserted anywhere. A stale stub here is how it drifted without a test noticing.

Dropping the stub so the tests exercise the real value:

```suggestion
```

packages/@sanity/cli/src/commands/__tests__/init/init.authentication.test.ts
```diff
@@ -56,6 +55,15 @@ vi.mock('../../../util/detectFramework.js', () => ({
   }),
 }))
 
+// Isolate from any ambient .env in the working directory: init only reads it to tailor the
+// not-logged-in hint, and a developer's local .env must not steer these auth assertions.
+vi.mock('../../../util/envFile.js', () => ({
+  GUARDED_ENV_KEYS: ['SANITY_AUTH_TOKEN', 'SANITY_PROJECT_ID', 'SANITY_CLAIM_URL'],
+  inspectEnvKeys: vi.fn(() => ({blankKeys: [], presentKeys: [], values: {}})),
+  readEnvValues: vi.fn(() => ({})),
+  TOKEN_ENV_FILES: './.env, or sanity/.env.local in a scaffolded project',
```
jwoods02: This stubs `TOKEN_ENV_FILES` as `'./.env, or sanity/.env.local in a scaffolded project'`, but the real constant (`envFile.ts:19`) is `'./.env or sanity/.env.local'`.

That string is interpolated into the command description, the `CLAIMED_PROJECT_IN_ENV` suggestion, two claim reminders, and both the login and logout warnings — six user-facing surfaces, none of them asserted anywhere. A stale stub here is how it drifted without a test noticing.

Dropping the stub so the tests exercise the real value:

```suggestion
```

(Same as `initAction.test.ts:85`.)

jwoods02 (commented):
Copy pass with Opus. Mostly string changes, plus one small logic change (using the framework we already detect to pick the right env var prefix).

**In this review:** 13 suggestions you can apply directly, and 1 comment that needs a decision from you (the claim URL wrapping — I don't know the right fix).

**I'd fix four before merge:** the `sanity new` help text, the Next.js variable names we show to non-Next frameworks, the expired-project message, and the claim URL wrapping. Everything else can land whenever.

**Not in this review:** a redesign of what `sanity new` prints when it runs. That's a reordering of `flow.*` calls rather than a string change. Proposed output at the bottom for reference.

### How I decided what to cut

For each string: can someone act on it without already knowing what the jargon means?

- "Never copy this token into code that runs in a browser" — yes, actionable without knowing what an environment variable is.
- "Never put it under a client-exposed prefix like `NEXT_PUBLIC_*`" — no, only if you already know the concept.

So: keep inline anything where not knowing costs you something you can't undo (losing the project, leaking the token). Everything that explains a concept goes to docs.

There's a docs page landing alongside this (I'll send you the preview), so a few comments suggest deleting text rather than rewording it. What I'd move out of CLI output:

- the `NEXT_PUBLIC_` / `PUBLIC_` / `VITE_` prefix rules per framework
- why the Studio needs a `#token=` link, and why port 3333
- what a dataset is, and what private vs public means for reads
- post-claim cleanup: which files to remove the token from
- `--no-scaffold` and the non-Next.js path

One more thing: `sanity new` and `projects mint` want different help text rather than a shared body. `projects mint` is the plumbing command and can stay technical. `sanity new` is where most people start.

┌ Setting up your Sanity project.
│
◇ Created "My Project"
◇ Project ID: abc12345
◇ Dataset: production (where your content lives)
│
◆ Claim your project by 1 August 2026, 09:22 UTC
│
│ https://www.sanity.io/manage/claim/<secret>
│
│ Until then it is temporary — the project and everything in it is deleted at that
│ deadline. Claiming is free, takes about a minute, and nothing you have built
│ changes. Treat the link like a password: whoever opens it becomes the owner.
│
◆ Created two folders
│ ./sanity your Studio, where you write and edit content
│ ./web your website, a Next.js app that reads it
│
│ cd sanity && npx sanity dev
│ then open http://localhost:3333/#token=<token>
│ the token in that link signs you in — there is no account yet
│ cd web && npm run dev
│ then open http://localhost:3000/
│
◆ Your access token is in ./.env and ./sanity/.env.local
│ It can read and change everything in this project, so both files are kept out of
│ version control for you. Never copy it into code that runs in a browser.
│
│ Your content is private until you claim the project, so anything reading it needs
│ that token. Claiming makes it readable without one.
│
│ Framework setup, and what to do after claiming: <docs link>
│
└ Claim your project: https://www.sanity.io/manage/claim/<secret>

shapirodaniel:
synced offline, we're planning to split this PR up a bit to land core `sanity new / sanity projects mint` functionality first, then fast-follow tighter scoped ambient credential management and claims reminder behaviors as these touch core cli functionality

please see https://github.com/sanity-io/cli/pull/1622

shapirodaniel:
canceling in favor of https://github.com/sanity-io/cli/pull/1622, which modifies scope, public api (`sanity projects unclaimed`, no `sanity projects mint`, `sanity new` handles provisioning + scaffolding directly)

PR #1598, opened 2026-07-24
feat: AIGRO-4919 - Invoking allowed commands from outside the cli

mwritter:
Screenshots - Locally tested linking `cli` changes to `mellon` and running `pnpm inspect` from `mellon`
image

Screenshot 2026-07-24 at 3 43 28 PM

image

---

> [!NOTE]
> **Medium Risk**
> Expands the attack surface for remote callers (many destructive API commands are allowed under MCP policy), though mitigated by explicit policies, token isolation, and disabled local project/config discovery; regressions in policy exhaustiveness or context routing could leak tokens or host filesystem behavior.
> 
> **Overview**
> Adds **in-process CLI invocation** for embedding hosts (e.g. MCP): `@sanity/cli/invokeSanityCli` runs allowed oclif commands, returns `{exitCode, output}`, and never shells out.
> 
> **`@sanity/cli-core/executionContext`** uses `AsyncLocalStorage` so each call can supply a **per-invocation token** (bypasses env/config/cache), **stdout/stderr sinks**, forced **non-interactive** behavior, and **no filesystem project-root discovery** (callers must pass flags like `--project-id`). `SanityCommand` logging/warnings and `getCliToken` honor this context; the normal `sanity` binary is unchanged.
> 
> **MCP command policy** is an exhaustive allow / deny / conditional table (`mcpPolicy`) synced to the oclif manifest; denied commands look like unknown commands. Help is filtered to the invokable surface (plain text, no probing denied topics). Includes shell-style **`tokenizeCliArgs`** (no shell) and a small guard on **`cors add`** so programmatic runs do not stat the host cwd.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab297aa7bf82a7d047d99e2774bbac9a01452a2e. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/util/tokenizeCliArgs.ts
```diff
@@ -0,0 +1,55 @@
+/**
+ * Tokenize a command-line argument string into an argv array without ever
+ * invoking a shell. Supports single quotes, double quotes, and backslash
+ * escapes — enough to round-trip anything a user would type after `sanity `.
+ *
+ * @throws When the input contains an unterminated quote.
+ * @internal
+ */
+export function tokenizeCliArgs(input: string): string[] {
```
jonahsnider: IMO, rather than keeping this in the CLI project, the MCP server should handle parsing the args. this keeps the bundle size smaller for the CLI

packages/@sanity/cli/src/exports/invokeSanityCli/InvokableCommands.ts
```diff
@@ -0,0 +1,20 @@
...
+}
+
+export const invokableCommands: ReadonlyMap<string, InvokableCommand> = new Map<
+  string,
+  InvokableCommand
+>([
+  ['cors add', CorsAdd],
+  ['cors delete', CorsDelete],
+  ['cors list', CorsList],
+  ['projects list', ProjectsList],
```
jonahsnider: is there a way we can run commands programmatically without needing to define them all here? we want to specify the allowed commands in the MCP server, not the CLI.

mwritter: 🤔 These are 'invoke-able' commands meant to be called from outside the cli so they are run programmatically.  The agent would call the tool with an 'args' `string` that mimics the actual cli command.

```ts
runSanityCli({
    args: "cors list --projectId asdfjkl",
    token: "mcp-user-token" // extra.authInfo.token
})
```

IMO I actually think it would be the clis responsability to tell the MCP what it should or shouldn't be able to call 🤷‍♂️ - The cli and anyone working it would know when a breaking change is introduced (ie add a file system call to an already allowed command) and it would be their responsibility to make it work with mcp commands.

So maybe we need to distinguish mcp commands better? Right now we export `invokeSanityCli` which could technically be used by any project that wants to piggy back off of the cli functionality but maybe we can pass a `source` and the `mcp` could have its own dedicated command list (policy)?

This is also why the 'tokenizer' I think should live in the cli as well, because if another `source` wants to invoke a command it will have to create its own 'tokenizer' as well 🤷‍♂️

jonahsnider: mostly my gripe was just the fact that we would be maintaining a giant list of commands, which is a bit annoying to keep updated over time

packages/@sanity/cli/src/exports/invokeSanityCli/help.ts
```diff
@@ -0,0 +1,164 @@
...
+ * @internal
+ */
+export function isHelpRequest(argv: string[]): boolean {
+  if (argv[0] === 'help') return true
+  for (const token of argv) {
+    if (token === '--') return false
+    if (token === '--help' || token === '-h') return true
+  }
+  return false
+}
```
jonahsnider: oclif provides a util that can consume `Config` and give us the help flags:

```
import {getHelpFlagAdditions} from '@oclif/core/help'

getHelpFlagAdditions(config) // -> string[]
```

packages/@sanity/cli/src/util/tokenizeCliArgs.ts
jonahsnider: if we can i'd prefer to use something like [string-argv](<https://npmx.dev/package/string-argv>), so we don't have to own this parsing logic ourselves

packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/mcpPolicy.ts
```diff
@@ -0,0 +1,186 @@
...
+ * execution context). Missing project or dataset values may therefore produce
+ * a usage error, but cannot cause local filesystem access. Destructive remote
+ * operations are allowed and do not by themselves make a command unsafe.
+ *
+ * Every manifest command must have exactly one policy here:
+ * - allow: every valid invocation is safe
+ * - conditional: safety depends on parsed arguments or flags
+ * - deny: no invocation is safe
+ */
+export const mcpPolicy: CommandPolicySet = {
```
jonahsnider, line 30: i wonder if we could add a test that consumes the generated OCLIF CLI manifest and fails if there's a command missing from this? would be nice to have as a reminder to update this when commands are added/removed/renamed

.changeset/pr-1598.md
```diff
@@ -0,0 +1,7 @@
+<!-- auto-generated -->
+---
+'@sanity/cli-core': minor
+'@sanity/cli': minor
+---
+
+feat: AIGRO-4919 - Invoking allowed commands from outside the cli
```
jonahsnider: should update this to not mention the ticket number

packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/policy.ts
```diff
@@ -0,0 +1,55 @@
...
+interface Invocation {
+  args: Readonly<Record<string, unknown>>
+  flags: Readonly<Record<string, unknown>>
+}
+
+export type InvocationPolicy = (invocation: Invocation) => boolean
+
+export interface CommandPolicy {
+  kind: 'allow' | 'conditional' | 'deny'
+  validate: InvocationPolicy
```
jonahsnider: can we make it so `conditional` is the only one which has a `validate` field? either that or remove the `kind` thing. imo it's a bit confusing that you have both

packages/@sanity/cli/src/commands/cors/add.ts
```diff
@@ -72,14 +73,19 @@ export class Add extends SanityCommand<typeof Add> {
...
+    // filesystem.
+    if (!getCliExecutionContext()) {
+      try {
+        const isFile = fs.existsSync(path.join(process.cwd(), args.origin))
+        if (isFile) {
+          this.warn(`Origin "${args.origin}?" Remember to quote values (sanity cors add "*")`)
+        }
+      } catch {
+        // Ignore errors checking if it's a file
       }
```
mwritter, line 88: Check for other issues like this

packages/@sanity/cli/src/exports/invokeSanityCli/index.ts
```diff
@@ -0,0 +1,234 @@
...
+      exitCode: exitCodes.USAGE_ERROR,
+      output: err instanceof Error ? err.message : String(err),
+    }
+  }
+
+  if (!policy.validate(invocation)) {
+    const displayId = commandId.replaceAll(':', ' ')
+    const usedDeniedFlags = (policy.deniedFlags ?? []).filter(
+      (name) => invocation.flags[name] !== undefined && invocation.flags[name] !== false,
+    )
```
mwritter: Only needed for conditionals

packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/policy.ts
mwritter: I think the 'validate' method is fine here, it gives us an abstract way to validate `conditional` policies and just returns a hardcoded `true` or `false` for the other kinds - otherwise we'll need to check the `kind` everywhere.  And I think having the `kind` is a nice readable way to disengage the policy types so I think both are fine like this.

I do however think the 'deniedFlags' need to be pulled out, deniedFlags always indicate a `conditional` `kind` so we don't want ppl thinking they can have an 'allowed' policy with 'deniedFlags' - they should create a 'conditional' policy (since its conditional on the flags - and currently thats the only way to create 'conditionals' via `conditionalDenyFlags`

packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/mcpPolicy.ts
mwritter: Yes we have this test in `commands.test.ts`
`the mcp policy covers exactly this package’s visible commands`

jonahsnider (approved):

PR #1600, opened 2026-07-27
feat(workbench): derive federation shared dependencies from the app's package.json

gu-stav:
### Description

Every workbench remote bundles its own react and `@sanity/ui`, so a session with several remotes loads several copies, and a plugin module rendered inside another app's tree hits dual-React crashes. Nothing could redirect those imports: the build passed `shared: {}`, so the remotes compiled plain bundled imports with no share-scope lookup for a host to satisfy.

Remotes now consume the copies they can safely share from the host's federation share scope, falling back to their own when no host provides one. Versions negotiate rather than unify, so remotes deployed independently keep working when they sit on different versions.

A candidate is shared when the app declares it, it resolves, every copy in the tree is within one major, and it isn't a dependency of another selected candidate. The last two rules carry the weight:

- A share key is per specifier, not per resolved version, so every import collapses onto one virtual module built from the top-level copy. Two majors in the same tree means the other major's consumers lose exports they need, which fails the build (`@sanity/ui` v3 `ThemeProvider` under a v4 top-level copy). Module federation can express this per module through `include`/`exclude` version filters, but the vite plugin drops them as of 1.19.1.
- Module federation drops a share that another share depends on, but only in dev, which would leave dev and build with different maps. `@sanity/sdk` under `@sanity/sdk-react` hits exactly this.

`react-dom` is left out on purpose. Sharing it expands into a `react-dom/server` share through module federation's common-subpath table, and materialising that leaks a child-process handle: `sanity build` writes correct output, reports success, then never exits. Reproduced on the vite plugin 1.18.1 and 1.19.1 in any app whose graph reaches `react-dom/server`, which `sanity` does. It's worth ~177 KB per remote, so it gets its own follow-up once that's fixed upstream.

Nothing provides a share scope yet, so this stays inert until the workbench shell seeds one. Until then a core app's output grows about 5.7% (57 KB on `dev/drop-desk`) from the extracted fallback chunks; that reverses once a host provides and the fallbacks stop being fetched.

Part of SDK-2170.

### What to review

The four selection rules in `shared-dependencies.ts`, and the resolution walk. Resolution deliberately walks the `node_modules` ancestor chain by hand instead of using `require.resolve.paths`, which also appends the global and `NODE_PATH` folders — under pnpm that reaches the hidden hoisted `.pnpm/node_modules`, where every transitive package looks locally installed.

Derived maps from real builds, both exiting cleanly:

| App                         | Shares                                                           |
| --------------------------- | ---------------------------------------------------------------- |
| `fixtures/federated-studio` | `react` (exact 19.2.7), `styled-components`, plus react subpaths |
| workbench `dev/drop-desk`   | `react` (exact), `@sanity/sdk-react`, plus react subpaths        |

drop-desk declares `@sanity/ui` and it is correctly skipped: its top-level copy is `4.0.0-static.46` while `sanity` nests v3. `@sanity/sdk` is correctly dropped as a nested candidate.

react carries an exact `requiredVersion` rather than the default caret range, because its react-dom stays bundled and React requires the pair to be the same version.

### Testing

8 unit tests over the selection rules, plus the two real builds above.

### DIFF
--- a/.changeset/workbench-shared-dependencies.md
+++ b/.changeset/workbench-shared-dependencies.md
@@ -1,0 +1,5 @@
+---
+'@sanity/workbench-cli': patch
+---
+
+feat(workbench): share compatible react, `@sanity/ui`, `@sanity/sdk`, `@sanity/sdk-react` and `styled-components` copies from the host's federation share scope
--- a/packages/@sanity/workbench-cli/src/actions/build/vite/__tests__/shared-dependencies.test.ts
+++ b/packages/@sanity/workbench-cli/src/actions/build/vite/__tests__/shared-dependencies.test.ts
@@ -1,0 +1,113 @@
+import fs from 'node:fs'
+import os from 'node:os'
+import path from 'node:path'
+
+import {type PackageJson} from '@sanity/cli-core'
+import {afterEach, beforeEach, describe, expect, test} from 'vitest'
+
+import {deriveSharedDependencies} from '../shared-dependencies.js'
+
+describe('deriveSharedDependencies', () => {
+  let projectDir: string
+
+  beforeEach(() => {
+    projectDir = fs.mkdtempSync(path.join(os.tmpdir(), 'workbench-shared-deps-'))
+  })
+
+  afterEach(() => {
+    fs.rmSync(projectDir, {force: true, recursive: true})
+  })
+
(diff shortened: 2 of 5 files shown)


gu-stav:

- #1600 ◀

PR #1602, opened 2026-07-27
fix(cli): remove dead manifest flags from schemas deploy, list, and delete

runeb:
### Description

`--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` are declared on `sanity schemas deploy`, `schemas list`, and `schemas delete` but wired to nothing. They stopped having any effect when the three commands were refactored to read the schema from the live workspace instead of a manifest file on disk (#390, #406, #473) — the flag declarations were left behind. Passing them was accepted and silently ignored.

`schemas deploy` was the worst case: its two flags were not `hidden`, and the command description actively promised behavior that no longer existed:

> Regenerates a manifest file by default. To re-use an existing manifest, use `--no-extract-manifest`.

This came up as a secondary finding while triaging AIGRO-5192 (a user passed `--no-extract-manifest` expecting it to change the deploy path, and it did nothing).

I removed the flags rather than implementing them: `deploySchemas` has no read-manifest-from-disk path at all, so making `--no-extract-manifest` real would be a new feature, not a bug fix. Happy to switch to deprecate-with-warning instead if you'd rather not take the unknown-flag error.

### What to review

- `commands/schemas/deploy.ts` — both flags removed, plus the stale manifest sentence in the description.
- `commands/schemas/list.ts` — both flags removed (already `hidden`), plus the same stale sentence.
- `commands/schemas/delete.ts` — both flags removed (already `hidden`).
- Worth confirming the call you made in #390/#406/#473 was to drop these rather than keep them for compatibility — you hid them on `list`/`delete` at the time, so tell me if that was deliberate staging toward removal or intended to stay.

Note the `extractManifest` / `manifestDir` options on `actions/deploy/deployStudioSchemasAndManifests.ts` are a different, live code path (studio `deploy`) and are untouched.

### Testing

- `pnpm test:unit --project=@sanity/cli` — 3320 tests, 0 failures.
- `pnpm check:types`, `pnpm check:lint` (0 errors), `pnpm check:deps` all pass.
- Verified no test, e2e spec, or help snapshot referenced either flag, and that `Flags` is still imported/used in all three files.

### Notes for release

Remove the `--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` flags from `sanity schemas deploy`, `list`, and `delete`. The flags stopped doing anything when these commands moved to reading the schema from the live workspace instead of a manifest file, but were still accepted and silently ignored. Passing them now reports an unknown-flag error.

---

> [!NOTE]
> **Low Risk**
> CLI-only UX and flag parsing; schema deploy/list/delete behavior is unchanged aside from optional warnings for legacy flags.
> 
> **Overview**
> **Deprecates** `--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` on `sanity schemas deploy`, `list`, and `delete`. Those flags were still accepted but had no effect after the commands switched to reading schemas from the live workspace.
> 
> Shared helpers in `removedManifestFlags.ts` re-register the flags as **hidden** (no defaults, so usage can be detected) and **`warnOnRemovedManifestFlags`** prints a clear deprecation message when any are passed. The three commands spread `removedManifestFlags` and call the warning at the start of `run`.
> 
> **Removes** outdated command descriptions that still claimed manifest regeneration / `--no-extract-manifest` behavior on **deploy** and **list**. Unit tests cover warning copy for single flags, negated `--no-extract-manifest`, and combined usage.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67f5fd8387a619618d3267b721ec1e5253c02743. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/schemas-remove-dead-manifest-flags.md
+++ b/.changeset/schemas-remove-dead-manifest-flags.md
@@ -1,0 +1,7 @@
+---
+"@sanity/cli": patch
+---
+
+Remove the `--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` flags from `sanity schemas deploy`, `list`, and `delete`.
+
+The flags stopped doing anything when these commands moved to reading the schema from the live workspace instead of a manifest file, but they were still accepted and silently ignored — and `deploy` still advertised manifest re-use in its help text. Passing them now reports an unknown-flag error. Behavior is otherwise unchanged.
--- a/.changeset/schemas-remove-dead-manifest-flags.md
+++ b/.changeset/schemas-remove-dead-manifest-flags.md
@@ -2,6 +2,6 @@
 "@sanity/cli": patch
 ---
 
-Remove the `--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` flags from `sanity schemas deploy`, `list`, and `delete`.
+Deprecate the `--extract-manifest` / `--no-extract-manifest` and `--manifest-dir` flags on `sanity schemas deploy`, `list`, and `delete`.
 
-The flags stopped doing anything when these commands moved to reading the schema from the live workspace instead of a manifest file, but they were still accepted and silently ignored — and `deploy` still advertised manifest re-use in its help text. Passing them now reports an unknown-flag error. Behavior is otherwise unchanged.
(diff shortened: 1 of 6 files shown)


binoy14 (approved):
Hadn't removed them thinking it might be a breaking change because if you pass that flag now it will error

binoy14 (changes requested):
opps previous one was supposed to be a question

runeb:
> Hadn't removed them thinking it might be a breaking change because if you pass that flag now it will error

@claude this comment from binoy is a good point. Could we keep the change that removes the now void usage example, but instead of erroring on these flags just output a warning that they are no longer effective?

PR #1604, opened 2026-07-27
fix(cli): stop printing "null" as the user when authenticating with a token

runeb:
### Description

Closes AIGRO-5269. Found while verifying AIGRO-5071 (now closed).

An API token resolves to a user with `email: null` and `provider: 'sanity-token'` — confirmed against `/v2025-08-30/users/me` with a real `developer`-role robot token:

```
id:       'p-hN9kdnhZSVMU'
name:     '… (Robot)'
email:     null
provider: 'sanity-token'
```

But `SanityOrgUser.email` was typed as a non-optional `string`, so TypeScript actively concealed this at every site that formatted a user. On token auth — the path agents and CI take — that produced:

| where | before |
| -- | -- |
| `sanity init` success line | `✔ You are logged in as null using Sanity-token` |
| generated `package.json` | `"author": "Name <null>"` |
| `sanity debug` | `Email: null` |

The package.json one is the worst of the three: it persists into the user's project rather than scrolling past in a terminal, and `Name <null>` isn't valid npm author syntax.

### What to review

- `cli-core/src/types.ts` — `email: string | null`, and `provider` widened to include `'sanity-token'`. **This is the load-bearing change**; the rest only typechecks honestly because of it. Widening it surfaced one real consumer (`actions/debug/types.ts`), which had copied the same wrong assumption.
- `actions/auth/getProviderName.ts` — adds `getUserDisplayName()` (email → name → id). Also maps `sanity-token` to `an API token`, since the generic branch was title-casing it into `Sanity-token`.
- `util/getProjectDefaults.ts:68` — emits the bare name when there's no email, instead of `Name <null>`.
- `commands/debug.ts:234` — `(none)` rather than `null`. Note `debug --json` still emits a real `null` here, which seems right for machine output.

One judgment call worth a second opinion: I put `getUserDisplayName` in `getProviderName.ts` since both are small user-formatting helpers used together at the same call sites. Happy to split it into its own module if you'd rather keep that file single-purpose.

### Testing

- New `actions/auth/__tests__/getProviderName.test.ts` — 10 cases covering the provider map, the `sanity-token` mapping, the full fallback chain, and an explicit assertion that output never contains `null`. There was no prior coverage of either function.
- `pnpm test:unit --project=@sanity/cli --project=@sanity/cli-core` — 3330 tests, 0 failures.
- `pnpm check:types`, `pnpm check:lint`, `pnpm check:deps` all pass.

Not covered by automated tests: an end-to-end `sanity init` under a robot token to see the generated package.json. The formatting logic is unit-tested, but I haven't watched the file get written.

### Notes for release

Stop rendering a literal `null` when authenticating with an API token. Token auth resolves to a user with no email, so `sanity init` printed `You are logged in as null using Sanity-token`, wrote `"author": "Name <null>"` into the generated `package.json`, and `sanity debug` reported `Email: null`. User-facing output now falls back to the account name, and `SanityOrgUser.email` is correctly typed as nullable.

---

> [!NOTE]
> **Low Risk**
> Localized CLI formatting and type widening for token users; no auth or API behavior changes.
> 
> **Overview**
> **Fixes user-facing CLI output when auth uses an API token** (robot users have `email: null` and `provider: 'sanity-token'`).
> 
> `SanityOrgUser` in **cli-core** now types `email` as `string | null` and adds `'sanity-token'` to `provider`. New **`getUserDisplayName()`** (email → name → id) and **`sanity-token` → "an API token"** in `getProviderName` drive **`sanity init`** login success lines. **`getProjectDefaults`** writes the bare account name into `package.json` author when there is no email (avoids `Name <null>`). **`sanity debug`** shows `Email: (none)` instead of `null` in human-readable output.
> 
> Unit tests cover provider mapping and the display-name fallback chain.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 598a9f2a580a5653bf23148201be23432cb14681. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/user-display-name-token-auth.md
+++ b/.changeset/user-display-name-token-auth.md
@@ -1,0 +1,6 @@
+---
+"@sanity/cli": patch
+"@sanity/cli-core": patch
+---
+
+Stop rendering a literal `null` when authenticating with an API token. Token auth resolves to a user with no email, so `sanity init` printed `You are logged in as null using Sanity-token`, wrote `"author": "Name <null>"` into the generated `package.json`, and `sanity debug` reported `Email: null`. User-facing output now falls back to the account name, and `SanityOrgUser.email` is correctly typed as nullable.
--- a/.changeset/user-display-name-token-auth.md
+++ b/.changeset/user-display-name-token-auth.md
@@ -3,4 +3,4 @@
 "@sanity/cli-core": patch
 ---
 
-Stop rendering a literal `null` when authenticating with an API token. Token auth resolves to a user with no email, so `sanity init` printed `You are logged in as null using Sanity-token`, wrote `"author": "Name <null>"` into the generated `package.json`, and `sanity debug` reported `Email: null`. User-facing output now falls back to the account name, and `SanityOrgUser.email` is correctly typed as nullable.
+Show the account name instead of `null` when authenticating with an API token, which has no email address.
--- a/packages/@sanity/cli/src/actions/auth/__tests__/getProviderName.test.ts
+++ b/packages/@sanity/cli/src/actions/auth/__tests__/getProviderName.test.ts
@@ -14,15 +14,6 @@ function createUser(overrides: Partial<SanityOrgUser> = {}): SanityOrgUser {
 }
 
 describe('getProviderName', () => {
-  test.each([
-    ['google', 'Google'],
-    ['github', 'GitHub'],
-    ['sanity', 'Email'],
-    ['saml-acme', 'SAML'],
-  ])('maps %s to %s', (provider, expected) => {
-    expect(getProviderName(provider)).toBe(expected)
-  })
-
   test('reads as prose for API tokens, rather than "Sanity-token"', () => {
     expect(getProviderName('sanity-token')).toBe('an API token')
   })
@@ -48,9 +39,5 @@ describe('getUserDisplayName', () => {
 
   test('falls back to the id when neither email nor name is usable', () => {
     expect(getUserDisplayName(createUser({email: null, name: ''}))).toBe('abc123')
-  })
-
-  test('never renders a literal "null"', () => {
-    expect(getUserDisplayName(createUser({email: null}))).not.toMatch(/null/)
   })
 })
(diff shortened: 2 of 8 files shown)


.changeset/user-display-name-token-auth.md
```diff
@@ -0,0 +1,6 @@
+---
+"@sanity/cli": patch
+"@sanity/cli-core": patch
+---
+
+Stop rendering a literal `null` when authenticating with an API token. Token auth resolves to a user with no email, so `sanity init` printed `You are logged in as null using Sanity-token`, wrote `"author": "Name <null>"` into the generated `package.json`, and `sanity debug` reported `Email: null`. User-facing output now falls back to the account name, and `SanityOrgUser.email` is correctly typed as nullable.
```
jonahsnider: this can probably just be a one liner

packages/@sanity/cli/src/actions/auth/__tests__/getProviderName.test.ts
```diff
@@ -0,0 +1,56 @@
...
+
+describe('getProviderName', () => {
+  test.each([
+    ['google', 'Google'],
+    ['github', 'GitHub'],
+    ['sanity', 'Email'],
+    ['saml-acme', 'SAML'],
+  ])('maps %s to %s', (provider, expected) => {
+    expect(getProviderName(provider)).toBe(expected)
+  })
```
jonahsnider: does this test do anything? doesn't really seem valuable

jonahsnider (changes requested):

packages/@sanity/cli/src/actions/auth/__tests__/getProviderName.test.ts
runeb: Agreed, removed. That block asserted a hardcoded lookup table against hardcoded expectations, and those four mappings aren't part of this change anyway — it would only ever fail if someone deliberately edited the map.

Also dropped a `never renders a literal "null"` case that was redundant with the fallback test above it. Down to 5 tests, all covering behaviour this PR actually changes.

.changeset/user-display-name-token-auth.md
runeb: Trimmed to one line:

> Show the account name instead of `null` when authenticating with an API token, which has no email address.

runeb:
Good review @jonahsnider, thanks! Pushed changes

jonahsnider (approved):

PR #1608, opened 2026-07-28
feat: support providing sanityEnv to execution context

jonahsnider:
Make it so the MCP server running in staging can run the CLI also in staging mode

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>

---

> [!NOTE]
> **Medium Risk**
> Changes which API hosts, CDN URLs, and staging flags apply during programmatic CLI runs; incorrect `sanityEnv` could send MCP or embedded callers to the wrong environment, though the interactive CLI path is unchanged.
> 
> **Overview**
> Adds optional **`sanityEnv`** (`production` | `staging`) to **`CliExecutionContext`**, so hosts like an MCP server can run **`invokeSanityCli`** against staging without mutating **`SANITY_INTERNAL_ENV`**.
> 
> **`isStaging()`** now treats an active execution context as authoritative: staging only when **`context.sanityEnv === 'staging'`**; otherwise it falls back to **`SANITY_INTERNAL_ENV`**. Call sites that previously read the env directly (API clients, **`getSanityUrl`**, auto-update module CDN hosts, bridge script injection, Vite **`__SANITY_STAGING__`**, workbench runtime, manage URLs) route through **`isStaging()`** or **`getSanityEnv()`** so the same per-invocation override applies consistently.
> 
> **`invokeSanityCli`** accepts **`sanityEnv`** and passes it into **`runWithCliExecutionContext`** alongside token and output sinks, enabling isolated concurrent invocations (e.g. production vs staging API hosts in integration tests).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b456f0580696769a43416a0e34850d1d2486005. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/util/getSanityEnv.ts
+++ b/packages/@sanity/cli/src/util/getSanityEnv.ts
@@ -1,4 +1,9 @@
 import {getCliExecutionContext} from '@sanity/cli-core/executionContext'
 
-export const getSanityEnv = () =>
-  getCliExecutionContext()?.sanityEnv ?? process.env.SANITY_INTERNAL_ENV ?? 'production'
+export function getSanityEnv() {
+  const context = getCliExecutionContext()
+
+  if (context) return context.sanityEnv ?? 'production'
+
+  return process.env.SANITY_INTERNAL_ENV ?? 'production'
+}
--- a/packages/@sanity/cli/src/util/getSanityEnv.ts
+++ b/packages/@sanity/cli/src/util/getSanityEnv.ts
@@ -1,1 +1,4 @@
-export const getSanityEnv = () => process.env.SANITY_INTERNAL_ENV || 'production'
+import {getCliExecutionContext} from '@sanity/cli-core/executionContext'
+
+export const getSanityEnv = () =>
+  getCliExecutionContext()?.sanityEnv ?? process.env.SANITY_INTERNAL_ENV ?? 'production'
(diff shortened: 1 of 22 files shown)


packages/@sanity/cli/src/util/getSanityEnv.ts
```diff
@@ -1 +1,4 @@
-export const getSanityEnv = () => process.env.SANITY_INTERNAL_ENV || 'production'
+import {getCliExecutionContext} from '@sanity/cli-core/executionContext'
+
+export const getSanityEnv = () =>
+  getCliExecutionContext()?.sanityEnv ?? process.env.SANITY_INTERNAL_ENV ?? 'production'
```
mwritter: 🤔 When theres an 'execution context' the command is being invoked from some embedded system, if that embedded system doesn't pass in a `sanityEnv` in its context then this will try to read from the `process.env` of the host (mcp, other app, ect).

I think we should change the check to

```ts
export function getSanityEnv() {
  const context = getCliExecutionContext()

  if (context) return context.sanityEnv

  return process.env.SANITY_INTERNAL_ENV ?? 'production'
}
```

And we'd need to update `isStaging()` the same way.

mwritter (commented):
This looks good to me! One comment on reading from `process.env`

packages/@sanity/cli/src/util/getSanityEnv.ts
jonahsnider: good point, will update

mwritter (approved):

PR #1615, opened 2026-07-29
feat(cli): show synced Shopify metafields in the Shopify templates

madken:
### Description

Sanity Connect imports product and collection metafields into `store.metafields` (BIL-859, BIL-860), but neither Shopify template declared that field, so the Studio rendered it as an unknown field and editors could not read the values. Both templates now show it as a read-only list.

The importer writes an array of `{_key, namespace, key, type, value}`, with `_key` set to `<namespace>.<key>`. Two properties of that data drive the design:

- **Members carry a `_key` but no `_type`.** The `_type: 'shopifyMetafield'` change is still an open follow-up on BIL-842. Each array therefore declares exactly one member type: `resolveTypeName` returns `"object"` for an object with no `_type`, and `getItemType` then falls back to `of[0]` only when `of.length === 1`. Existing synced data validates with no migration, but adding a second member type to either array would break it.
- **`value` is deserialized from GraphQL `jsonValue`**, so one metafield may hold a string, number, boolean, list or object. No Sanity field type accepts all of those — declaring `value` as `string` renders an incompatible-type warning for every `number_integer`, `boolean`, `list.*` and `json` metafield. So `value` is not declared as a field, and a small read-only input reads it off the raw object value instead.

Since the synced `type` says what shape `value` has, `utils/formatMetafieldValue.ts` formats each type rather than dumping JSON:

| Metafield type | Synced value | Displayed as |
| --- | --- | --- |
| `dimension`, `weight`, `volume` | `{"value": 180, "unit": "CENTIMETERS"}` | `180 cm` |
| `money` | `{"amount": "649.00", "currency_code": "USD"}` | `649.00 USD` |
| `rating` | `{"value": "4.7", "scale_max": "5.0"}` | `4.7 / 5.0` |
| `list.*` | `["Poplar core", "Carbon stringers"]` | `Poplar core, Carbon stringers` |
| `json`, `rich_text_field` | `{"camber": "hybrid"}` | indented JSON |
| everything else | `"Advanced"`, `7`, `true` | as-is |

`json` and `rich_text_field` are the only values rendered in a code block; everything else formats to one line and renders as text. Measurement units missing from the unit map render unchanged rather than being guessed at, matching how the importer normalizes them (BIL-870). Dates keep their synced ISO form so they stay unambiguous.

Both templates gain a `shopifyMetafield` object type, a `components/inputs/ShopifyMetafield.tsx` renderer and the `formatMetafieldValue` helper. `store.metafields` is added to the product and collection `store` objects under a new collapsible Metafields fieldset. `templates/shopify/docs/features.md` gains a section describing the shape, the formatting table and the read-only behaviour.

Closes the Studio-template follow-up on [BIL-842](https://linear.app/sanity/issue/BIL-842), previously tracked as BIL-892.

### What to review

The type name must stay exactly `shopifyMetafield`. That is the contract with the importer's pending `_type` change.

To see it rendered, scaffold a studio from `templates/shopify` and import a product whose `store` holds metafields like these:

```json
"metafields": [
  {"_key": "custom.flex_rating", "namespace": "custom", "key": "flex_rating", "type": "number_integer", "value": 7},
  {"_key": "custom.materials", "namespace": "custom", "key": "materials", "type": "list.single_line_text_field", "value": ["Poplar core", "Carbon stringers"]},
  {"_key": "custom.snowboard_length", "namespace": "custom", "key": "snowboard_length", "type": "dimension", "value": {"value": 180, "unit": "CENTIMETERS"}},
  {"_key": "custom.msrp", "namespace": "custom", "key": "msrp", "type": "money", "value": {"amount": "649.00", "currency_code": "USD"}},
  {"_key": "custom.spec_sheet", "namespace": "custom", "key": "spec_sheet", "type": "json", "value": {"camber": "hybrid"}},
  {"_key": "custom.empty_note", "namespace": "custom", "key": "empty_note", "type": "single_line_text_field", "value": ""}
]
```

Then open Shopify sync -> Shopify -> Metafields and check that each value reads as the table above describes, that `custom.empty_note` shows "No value", and that a product with no `store.metafields` shows an empty fieldset rather than a warning. These members deliberately have no `_type`, which is the case worth confirming.

Scaffolding needs `sanity@5` today. The templates pin `sanity-plugin-hotspot-array@^3.0.2` (peer `^4 || ^5`) and `sanity-plugin-media@^4.1.1` (peer `^3.78 || ^4 || ^5`), so npm fails with `ERESOLVE` against `sanity@6`, which is what `latest` now resolves to. Pre-existing and filed separately.

Affected: `packages/@sanity/cli/templates/shopify` and `packages/@sanity/cli/templates/shopify-online-storefront`. No CLI source changes. Variant metafields and metaobjects are out of scope, because the importer does not emit them, so `productVariant` is unchanged.

### Testing

No automated tests. `templates/` has no test or type coverage in this repo: `check:types` is scoped to `@sanity/cli` `src` and `test`, and `test:unit --changed` matches zero tests for a template-only change. A type error in these files would not fail CI, which is worth knowing when reviewing.

Verified manually instead:

- `check:format`, `check:lint` (0 errors) and `check:types` pass. `pnpm test:unit --project=@sanity/cli` gives 3380 passed / 0 failed.
- Ran `formatMetafieldValue` against 31 assertions out-of-tree covering every branch: each measurement type, unknown units, a `0` value, money with and without a currency, ratings with and without a scale, `list.*` of both scalars and objects, an empty list, blank and empty-string values, and four type/value mismatches to confirm none of them throw.
- Typechecked the changed template files out-of-tree against real `sanity@6.7.0` and `@sanity/ui@3`, which are not installable in this repo. Clean, including `components.input` on an object type, `ObjectInputProps`, and the `preview.prepare` signature.
- Confirmed the `_type` fallback in the shipped `sanity@6.7.0` source rather than assuming it: `resolveTypeName` in `@sanity/util/lib/content.js` and the `of.length === 1` branch of `getItemType`.
- Ran a studio built from the modified `shopify` template against a real dataset, and confirmed by GROQ query that none of the imported metafield members carry a `_type`.

Not verified: the visual rendering of `ShopifyMetafield.tsx`. Layout, spacing and icon choices are unreviewed.

### DIFF
--- a/.changeset/pr-1615.md
+++ b/.changeset/pr-1615.md
@@ -3,4 +3,4 @@
 '@sanity/cli': minor
 ---
 
-feat(templates): show synced Shopify metafields in the Studio
+feat(cli): show synced Shopify metafields in the Shopify templates
--- a/.changeset/pr-1615.md
+++ b/.changeset/pr-1615.md
@@ -3,4 +3,4 @@
 '@sanity/cli': minor
 ---
 
-feat(cli): surface synced Shopify metafields in the Shopify studio templates
+feat(templates): show synced Shopify metafields in the Studio
--- a/.changeset/pr-1615.md
+++ b/.changeset/pr-1615.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+feat(cli): surface synced Shopify metafields in the Shopify studio templates
--- a/packages/@sanity/cli/templates/shopify-online-storefront/components/inputs/ShopifyMetafield.tsx
+++ b/packages/@sanity/cli/templates/shopify-online-storefront/components/inputs/ShopifyMetafield.tsx
@@ -1,0 +1,55 @@
+import {LockIcon} from '@sanity/icons'
+import {Box, Card, Code, Flex, Stack, Text} from '@sanity/ui'
+import {type ObjectInputProps} from 'sanity'
+import {formatMetafieldValue} from '../../utils/formatMetafieldValue'
+
+type Metafield = {
+  namespace?: string
+  key?: string
+  type?: string
+  value?: unknown
+}
+
+/**
+ * Read-only display for a single Shopify metafield.
+ *
+ * `value` is deliberately not declared as a field on the `shopifyMetafield` schema type: its shape
+ * depends on the Shopify metafield type and Sanity has no field type that accepts every variant. We
+ * read it off the raw object value instead and render it as text.
+ */
+const ShopifyMetafield = (props: ObjectInputProps) => {
+  const {namespace, key, type, value} = (props.value || {}) as Metafield
+
+  const formattedValue = formatMetafieldValue(value)
+
+  return (
+    <Card padding={3} radius={2} shadow={1} tone="transparent">
+      <Stack space={3}>
+        <Flex align="center" gap={2}>
+          <Text size={1} weight="semibold">
+            {[namespace, key].filter(Boolean).join('.') || 'Metafield'}
+          </Text>
+          {type && (
+            <Text muted size={1}>
+              <code>{type}</code>
+            </Text>
+          )}
+          <Text muted size={1}>
+            <LockIcon />
+          </Text>
+        </Flex>
+        {formattedValue ? (
+          <Box overflow="auto">
+            <Code size={1}>{formattedValue}</Code>
+          </Box>
+        ) : (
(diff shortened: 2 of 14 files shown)


madken:
Superseded by #1624. Same commits, recreated from `bil-947-studio-template-metafields` so Linear's git integration links the PR to BIL-947 instead of the already-closed BIL-842 epic. No review had started here.

PR #1622, opened 2026-07-29
feat(cli): add sanity new, sanity projects unclaimed

shapirodaniel:
### Description

Adds the first self-contained sanity new flow: mint an unclaimed project, optionally scaffold a connected Studio and Next.js app, and retain local recovery details.

### Scope

`sanity new` allows an agent pointed to `https://sanity.new` to build a project worth claiming, supported by skills geared towards pre-claim project content operations. `sanity projects unclaimed` allows the agent to discover unclaimed projects.

### What to review

Focus on the mint and recovery lifecycle, scaffold permutations, guarded environment handling, unclaimed projects behavior, and the clarity of the terminal flow.

### Testing

Added unit and integration coverage across minting, recovery, environment files, scaffolding, and output formatting. Also manually UAT’d the greenfield, existing-framework, --force, --no-scaffold, and unclaimed projects flows.

#### `sanity new` flow

sanity-new-flow-exemplar

#### `sanity projects unclaimed [--project-id]`

Screenshot 2026-07-30 at 4 08 06 PM

#### post-claim behavior, `sanity projects unclaimed [--project-id]`

Screenshot 2026-07-30 at 5 44 47 PM
Screenshot 2026-07-30 at 7 38 16 PM

---

> [!NOTE]
> **High Risk**
> Large new surface: unauthenticated provisioning, live tokens written to `.env.local` and user config, and claim-link handling—security- and data-loss sensitive despite extensive tests.
> 
> **Overview**
> Introduces **`sanity new`**: mint an unclaimed project via the provision API (no login), persist recovery metadata locally, then optionally scaffold `./sanity` and `./web` with matching `.env.local` files holding the robot token. The command uses a new **`createFlow`** railed terminal UX in `@sanity/cli-core`, blocks minting when `./sanity` or `./web` is non-empty, supports `--no-scaffold` and `--json`, and treats scaffold failures as best-effort while still returning project/claim details.
> 
> Adds **`sanity projects unclaimed`** to list or show (`--project-id`) locally stored claim URLs and tokens, prune expired or claimed projects (via `getProjectClaimStatus`), and avoid activating credentials. **`mintUnclaimedProject`**, **`scaffoldProject`** / **`createFrontend`**, and **`initStudio`**’s **`preclaim`** flag wire the mint → record → scaffold path; package installs gain **cancel/timeout** and throw **`CLIError`** instead of calling **`output.error`**. MCP policy **denies** `new` and **`projects:unclaimed`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d2ba6004feef94578f98836f13bf58049f1320c6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+          flow.note('Setup stopped. Your project was still created.')
+          flow.line('Claim it before the deadline to keep it.')
+          flow.gap()
+          flow.link(minted.claimUrl, {label: 'Claim your project:', outro: true})
+          throw err
+        }
+        flow.note(`Automatic setup did not finish: ${err instanceof Error ? err.message : err}`)
+        flow.line('The project is ready. You do not need to create another one.')
+        flow.gap()
+        printScaffoldRecipe()
```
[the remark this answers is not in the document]
shapirodaniel: def worth tailoring the scaffold recipe to the user's intended framework if we know it before we hit this failure branch; can be a fast follow fix up if there's no additional changes required here

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+      "and whoever opens it takes ownership: keep it out of git and shared channels. If you're " +
+      'an agent, surface the link to the end user immediately. Everything keeps working after ' +
+      "you've claimed, including the robot token.",
+    'The robot token has full content access to this project: create, edit, publish, and deploy ' +
+      'schemas. It cannot deploy a hosted Studio, create datasets, or manage settings. The dataset ' +
+      'is private until you claim, so frontend reads must run server-side; claiming makes it ' +
+      'public. Wherever it lives, the token must never sit under a browser-exposed prefix.',
+    `After the claim, run \`sanity login\` and remove SANITY_AUTH_TOKEN from ./.env or ${STUDIO_ENV_FILE} ` +
+      'to act as your own account; until then, CLI commands in this directory keep ' +
+      'authenticating as the robot.',
```
jwoods02: This is factually wrong and it's the one I'd fix first, because agents will act on it.

Nothing reads `./.env`. The prerun hook that loads `.env` files (`injectEnvVariables`) only fires once it finds a Sanity config, and the mint root has none, so `getCliToken` never sees that token. Removing `SANITY_AUTH_TOKEN` from `./.env` after claiming is a no-op, and "CLI commands in this directory keep authenticating as the robot" is false in the directory where this text is printed.

It is true one level down: `sanity/.env.local` plus `sanity/sanity.config.ts` does make the hook fire, so commands run from `./sanity` genuinely authenticate as the robot. So the copy is describing the right behaviour in the wrong place. Same for line 149's "run those from ./sanity" — that works, but only because of the Studio env file, which is worth saying out loud rather than implying the directory itself carries auth.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+        flow.highlight('Create your website')
+        flow.gap()
+        flow.command(frontendCommand)
+        flow.gap()
+        flow.line('then add:')
+        flow.gap()
+        printFrontendEnv()
+      }
+
+      if (!scaffold.studioEnvWritten) {
```
jwoods02: This is the only place the raw token is printed, and it only fires when the Studio env write **failed**. So on every successful run the token never appears in full anywhere except URL-encoded inside the `#token=` link.

I'd print it unconditionally, as a copyable value, alongside the project ID and dataset in the early handoff. Two reasons. It's what the launch-minimum spec asks for. And more practically, right now if someone loses the terminal output there is no way to get the token back: the registry has it, nothing reads it, and no command surfaces it.

Also missing and worth adding next to it: the `SANITY_AUTH_TOKEN="<token>" sanity <command>` example the spec mandates. Grep says it appears nowhere in the source.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+  )
+
+  public async run(): Promise<MintProjectResult> {
+    const {output} = this
+    const json = this.jsonEnabled()
+    const flow = createFlow(output.log)
+    const invoked = [this.config.bin, ...(this.id?.split(':') ?? [])].join(' ')
+    const cwd = process.cwd()
+    const envPath = path.join(cwd, '.env')
+    const existingEnvFiles = this.guardExistingProject(
```
jwoods02: This runs before the `json` branch, so `--json` reads `.env` from disk and can hard-refuse to mint. An agent calling `sanity new --json` in a repo that happens to have `SANITY_PROJECT_ID` in `.env` gets `EXISTING_SANITY_ENV_VALUES` instead of a project, which breaks the headless contract.

Looks like a one-line move inside `if (!json)`. The `--json` test seeds an empty inspection so it doesn't catch this.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+      [
+        {displayPath: './.env', path: envPath},
+        {displayPath: './.env.local', path: path.join(cwd, '.env.local')},
```
jwoods02: The guard inspects `./.env.local`, but mint never writes that file. So the refusal can fire on someone's own long-lived token for a completely different project, and the message tells them "this directory already has Sanity project values" when it doesn't have anything of ours.

I'm in favour of keeping the guard generally, for what it's worth — with 58 of 79 test projects expiring, something that stops a silent second mint earns its place, and the spec's alternative (just make the side effect clear in help copy) feels thin. But I'd scope it to the files we actually write.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+        flow.gap()
+        printEnvValues()
+        flow.gap()
+        if (this.flags.scaffold) printScaffoldRecipe()
+        rootEnvRecoveryPrinted = true
+      }
+      flow.gap()
+    }
+
+    const shouldScaffold = !json && this.flags.scaffold && !hasExistingKeys && setupValuesWritten
```
jwoods02: `--force` sets `hasExistingKeys`, so it silently turns off scaffolding, which is on by default. There's a test asserting this so I assume it's deliberate, but nothing in the flag description or the example at line 183 says so — someone passing `--force` in a folder with no `sanity/` or `web/` gets a bare mint and no explanation.

Either say it in the flag help or let `--force` scaffold.

packages/@sanity/cli/src/commands/projects/mint.ts
```diff
@@ -0,0 +1,689 @@
...
+            flow.highlight(`Update ${key} in ./.env`)
+            flow.line('The existing value was left unchanged. Set it to:')
+            flow.gap()
+            printValues({[key]: envValues[key as keyof typeof envValues]})
+            flow.gap()
+          }
+          rootEnvRecoveryPrinted = true
+        }
+
+        setupValuesWritten = SCAFFOLD_REQUIRED_ENV_KEYS.every((key) =>
```
jwoods02: Scaffolding is gated on the `./.env` write having succeeded, but the scaffold needs nothing from that file — `initStudio` gets the project ID and dataset as arguments. So an unwritable or shadowed `.env` cancels a Studio and frontend that would otherwise work fine.

Worth decoupling regardless of what you decide about `.env` itself.

packages/@sanity/cli/src/actions/scaffold/scaffoldProject.ts
```diff
@@ -0,0 +1,269 @@
...
+        frontendEnvWritten: false,
+        studioEnvWritten,
+        studioPath,
+      }
+    }
+    throw err
+  }
+
+  const frontendPath = path.join(workDir, FRONTEND_DIR)
+  const frontendEnvWritten = writeScaffoldEnv(
```
jwoods02: Raising this as a gap rather than a defect, and not blocking.

`web/.env.local` gets the project ID and dataset but no token, which is right on security grounds. The consequence is that the scaffolded frontend can't read anything: the dataset is private pre-claim, and no client code is written either, so `web/` knows where to point and nothing more. Whoever comes next has to remember to add a server-side token and keep it server-side, and that's exactly the step the private-dataset behaviour already makes easy to get wrong (200 with empty results rather than a 401).

The skill will prompt agents to add `SANITY_API_READ_TOKEN` themselves, so this isn't broken. But given how sharp the private-dataset edge already is, writing a server-only token here with a loud comment might be safer than leaving a hole people have to notice. If you do, please make sure `web/.env.local` is definitely gitignored — `create-next-app` ships `.env*` in its own `.gitignore`, and the root `.env*` entry covers it too, but I'd want that pinned by a test rather than inherited by luck.

Either way, worth a decision rather than a default.

packages/@sanity/cli-core/src/ux/flowOutput.ts
```diff
@@ -0,0 +1,145 @@
...
+}
+
+function wrapText(text: string): string[] {
+  const terminalWidth =
+    process.stdout.isTTY && Number.isFinite(process.stdout.columns) && process.stdout.columns > 0
+      ? process.stdout.columns
+      : MAX_RENDERED_LINE_WIDTH + 1
+  const renderedWidth = Math.min(terminalWidth - 1, MAX_RENDERED_LINE_WIDTH)
+  const width = Math.max(renderedWidth - RAIL_PREFIX_WIDTH, 1)
+  return wrapAnsi(text, width, {hard: false, wordWrap: true}).split('\n')
```
jwoods02, line 25: `wrap-ansi`'s `trim` option defaults to `true`, so this strips leading whitespace from every wrapped line. The effect is that indented values printed through the flow (the `printValues` / env-line helpers in mint) lose their indent and render flush against the rail, looking like prose rather than something you're meant to copy. `command()` skips `wrapText` so it keeps its indent, which makes the output inconsistent with itself.

`{trim: false}` should do it. Nothing pins indentation today — the tests use `toContain` and an unindented fixture.

packages/@sanity/cli-core/src/ux/flowOutput.ts
```diff
@@ -0,0 +1,145 @@
...
+  link(url: string, options?: {label?: string; outro?: boolean}): void
+  /** `●` — a step happening behind the scenes, or a tip. */
+  note(text: string): void
+  /** `└` — closing line of the story. */
+  outro(text: string): void
+  /** `◇` — a completed step or produced value. */
+  result(text: string): void
+  /**
+   * An in-flight step. Resolve it with `succeed(text)` (persists as a `◇` line) or `fail(text)`;
+   * renders on stderr so machine-readable stdout is never corrupted.
```
jwoods02, line 90: The doc comment promises "renders on stderr so machine-readable stdout is never corrupted", but the non-interactive fallback routes `note` / `fail` / `succeed` through `log`, which goes to stdout.

Harmless today because mint never calls `spin` under `--json`, but this module is general-purpose now and that comment is exactly what the next adopter will trust. Worth either routing the fallback to stderr or softening the promise.

packages/@sanity/cli-core/src/ux/flowOutput.ts
```diff
@@ -0,0 +1,145 @@
...
+   */
+  spin(text: string): {fail(text: string): void; succeed(text: string): void}
+}
+
+/** Create a {@link Flow} that writes rail lines through `log`. */
+export function createFlow(log: LogFn): Flow {
+  return {
+    command(command: string) {
```
jwoods02, line 98: The URL path in this module is properly hardened (`isSafeTerminalLink`, `encodeTerminalControls`, with a sanitised bare-string fallback) — nice work. But `command()` and the spinner text aren't sanitised, and the spinner prints `displayName`, which is user input validated only for length.

`sanity new $'\e]8;;http://evil\aclick\e]8;;\a'` emits a live hyperlink. Self-inflicted so low severity, but it undercuts the guarantee the module otherwise makes, and running the same `encodeTerminalControls` over both would close it.

packages/@sanity/cli/src/actions/init/initStudio.ts
```diff
@@ -138,6 +140,9 @@ export async function initStudio({
     output.log(`  ${styleText('cyan', `npx sanity dataset create <name>`)}\n`)
   }
 
+  // `sanity new` owns the combined Studio/frontend outro.
+  if (preclaim) return
```
jwoods02, line 146: This is a clean seam and it fixes the duplicate "Get started" block nicely.

Separate thing Opus noticed while tracing this path: `updateProjectInitializedAt` earlier in this function goes through `getProjectCliClient({requireUser: true})`, and a mint flow has no logged-in user, so it throws on every single `sanity new` run and gets swallowed as "non-critical". Harmless but it's one guaranteed-failing request per mint, and `preclaim` is the obvious signal to skip it.

packages/@sanity/cli/src/util/unclaimedProjects.ts
```diff
@@ -0,0 +1,58 @@
...
+    ? (value as Record<string, unknown>)
+    : {}
+}
+
+/**
+ * Record a mint response in the passive, project-ID-keyed user-config registry.
+ *
+ * This storage is deliberately not connected to CLI credential resolution.
+ */
+export function recordUnclaimedProject(
```
jwoods02, line 31: Non-blocking suggestion,/

This record already holds everything needed for recovery (token, claimToken, claimUrl, expiresAt), and right now nothing can read it back, so a lost terminal line means a lost project even though we have the details on disk. A small read-only command would close that: something like `sanity projects unclaimed` to list and `sanity projects unclaimed <id>` to show the token / claim URL for one

Happy for it to be a follow-up or to discuss command structure more

packages/@sanity/cli/src/commands/projects/__tests__/mint.test.ts
```diff
@@ -0,0 +1,556 @@
...
+
+  test('mints, records, writes setup values, and scaffolds the project', async () => {
+    await expect(MintProjectCommand.run(['My New Project'])).resolves.toEqual(result)
+
+    expect(mockMintUnclaimedProject).toHaveBeenCalledWith({displayName: 'My New Project'})
+    expect(mockRecordUnclaimedProject).toHaveBeenCalledWith(minted)
+    expect(mockAppendEnvValues).toHaveBeenCalledWith(
+      expect.stringMatching(/\.env$/),
+      {
+        SANITY_AUTH_TOKEN: minted.token,
```
jwoods02: Flagging this one because of what it locks in rather than because it's wrong: it asserts the token **is** written to `./.env`, which is the opposite of what the launch-minimum spec's required coverage asks for ("minting writes no robot token or claim secret into the project tree"). Whichever way that decision goes, this test decides it, so it's worth being deliberate about.

While you're in here, the spec names 16 assertions and a few of the load-bearing ones are missing: that the handoff prints before scaffolding starts (no ordering assertion anywhere), that a registry write doesn't touch the active credential or token cache, and that a command run from the mint root or a generated descendant gains no mint credential. That last one can't pass for `./sanity` today, which is itself the useful finding.

jwoods02 (commented):
Me and opus found a few things, but like how far this is scoped down.

Only comment is if we could rename [‎packages/@sanity/cli-core/src/ux/flowOutput.ts](https://github.com/sanity-io/cli/pull/1622/changes#diff-554f2ff87ed654888f87453ec6dfcf0bc36b500f78deb1940c61c59ca2039e5e) to be more obvious that it's a clack style output that should be sued for specific interactive flows like this, not general purpose API flows, it's a more stylised approach we're adding, don't want AI reaching for it when doing other CLI work.

packages/@sanity/cli/src/commands/projects/mint.ts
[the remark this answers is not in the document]
shapirodaniel: fixed in b59b081c

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: nice call out, updated copy to give corrected info in d6f7d67ab1cce97a6

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: great suggestion, fixed up in 5a6f37f739e91df25715a644669063c87432d7cd

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: thanks for this, took a closer look and ended up implementing a check like this in fe0f01caf1e4a15aa72e14c95007240f35568028

```
const existingEnvFiles = json ? [] : this.guardExistingProject(...);
```

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: nice call out, `--force` semantics elsewhere mean "do this unconditionally", which imo should include scaffolding. fixed up in 7694bbde1fc0867231f3817e0d63e8f9b41797b4

packages/@sanity/cli/src/commands/projects/mint.ts
shapirodaniel: i agree, keeping the guard here makes sense, this is addressed in the next comment's commit (7694bbde1fc0867231f3817e0d63e8f9b41797b4)

packages/@sanity/cli-core/src/ux/flowOutput.ts
shapirodaniel: great catch, added a stderr writer to use in `spin`'s `error` callback in b59122a95d6d87367c1ed34958519a3cd97f0f24

packages/@sanity/cli-core/src/ux/flowOutput.ts
shapirodaniel: def agree on having sanitized outputs by default in `flow` methods, thanks for this -- fixed up in e7300ceb9b6299a18361925cdac71f1bd3b0aa40

packages/@sanity/cli/src/actions/init/initStudio.ts
shapirodaniel: nice one, yeah agree we can 

```
if (!preclaim) // update
else // pre-claim, non-critical update debug stmt
```

fixed in 5cdd083f6f18d41cef3cc47eea4dfdd64702215a

packages/@sanity/cli/src/util/unclaimedProjects.ts
shapirodaniel: love this idea! and it plays nicely with [your suggestion](https://github.com/sanity-io/cli/pull/1622#discussion_r3682087445) to reconcile launch minimum guidance of "no token writes to `.env`" with what's currently happening in this PR

if we implement `sanity projects unclaimed [projectId]` now, we can stop writing the token to `.env` and `.env.local` and get better deterministic credential use -- no more ambient `SANITY_AUTH_TOKEN` pickup from `.env*`, which is where a lot of the undesired complexity in https://github.com/sanity-io/cli/pull/1592 came from

packages/@sanity/cli/src/commands/projects/__tests__/mint.test.ts
shapirodaniel: great call out, i'd propose we take your suggestion to implement `sanity projects unclaimed [projectId]`, please see [this comment](https://github.com/sanity-io/cli/pull/1622#discussion_r3683121489) above for details

packages/@sanity/cli/src/commands/new.ts
```diff
@@ -0,0 +1,406 @@
...
+      if (scaffold?.frontendPath) {
+        flow.note(`Your access token is in ./${STUDIO_ENV_FILE} and ./${FRONTEND_ENV_FILE}`)
+        flow.line('Keep both files out of version control.')
+      } else if (scaffold) {
+        flow.note(`Your access token is in ./${STUDIO_ENV_FILE}`)
+        flow.line('Keep that file out of version control.')
+      } else {
+        flow.note('Your access token is shown above.')
+        flow.line('Keep it out of version control.')
+      }
```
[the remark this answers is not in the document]
shapirodaniel: this falls under the class of partial failures that we could solve for, but are pretty complex: i'm inclined to punt on this

packages/@sanity/cli/src/commands/projects/__tests__/mint.test.ts
shapirodaniel: update, after offline sync, we aligned on a few core changes here

- `sanity projects mint` goes away (mint is an internal concept)
- `sanity new` is not an alias, it's the top-level mint-and-scaffold command
- `sanity projects unclaimed [--project-id]` allows users/agents to recover minted projects from cli config
- `--force` has more power over scaffolding behavior
- `.env*` writes are consistent across top-level and `.env.local` files; humans/agents manage their contents

the `unclaimed` subcommand is supported by agent skill copy in https://github.com/sanity-io/sanity-new/pull/4

shapirodaniel:
let's wait and see on [this bugbot finding for potential inconsistency](https://github.com/sanity-io/cli/pull/1622#issuecomment-5137884822) if `removeUnclaimedProject` fails to modify cli config. it's a valid issue, but transient: the next cli command will generally succeed in reconciling unclaimed projects state

packages/@sanity/cli/src/hooks/prerun/claimReminder.ts
```diff
@@ -0,0 +1,103 @@
...
+  return [
+    claimPrompt,
+    claimInstructions,
+    deletionWarning,
+    allProjectsInstructions,
+    agentInstructions,
+  ].join('\n\n')
+}
+
+export const claimReminder: Hook.Prerun = async function ({argv, Command}) {
```
jwoods02: The reminder only checks two things before it fires: is the project in the local ledger, and has it not expired yet. It never checks whether the project has been claimed, and it never checks which project you are actually working on.

That gives us two problems:

- After someone claims their project, every command still prints "Claim project X now" in bold red for the rest of the 72 hours. The only thing that clears the record is running `sanity projects unclaimed`.
- The reminder fires for any project on the machine, from any folder, even when you are logged in and working on something completely unrelated. There is no cwd check, no config check and no token check.

Checking claim state means calling `getProjectClaimStatus`, which has no timeout today, so we would be adding a network round trip to every single command including `sanity --version`. Scoping the nudge to the project you are actually in means tying a folder or a token back to a ledger entry, and then this starts snowballing again.

Suggest we take the reminder out of this PR then bring it back in the reminder follow-up with a timeout budget and a proper answer for scoping unless there's a simple solution I'm not thinking of

Everything else in the flow looks good to me.

packages/@sanity/cli/src/commands/new.ts
```diff
@@ -0,0 +1,389 @@
...
+      flow.gap()
+    }
+
+    if (!json) {
+      if (!scaffold && !this.flags.scaffold) {
+        flow.result('Project created without scaffolding')
+        flow.line(
+          'No folders or env files were created. Use the project ID and dataset above in your own setup.',
+        )
+        flow.gap()
```
jwoods02, line 356: Small one

On the `--no-scaffold` path we never print the Studio sign-in link, because that only prints inside the scaffold branch further down. Suggested fix:

```suggestion
        flow.result('Project created without scaffolding')
        flow.line(
          'No folders or env files were created. Use the project ID and dataset above in your own setup.',
        )
        flow.gap()
        flow.link(`http://localhost:3333/#token=${encodeURIComponent(project.token)}`, {
          label: 'Once a Studio is running on http://localhost:3333, sign in with:',
        })
        flow.gap()
        flow.line('The token signs you in: there is no account yet.')
        flow.gap()
```

Less important now the token is now printed in copyable form above on every path, and `sanity projects unclaimed --project-id` prints it again, so nothing is unrecoverable. All that is missing is the ready made URL.

packages/@sanity/cli/src/hooks/prerun/claimReminder.ts
```diff
@@ -0,0 +1,103 @@
...
+    "There's no downside to claiming early. Open this link to claim your project:",
+    claimUrl,
+  ].join('\n')
+  const deadline = formatClaimDeadline(record.expiresAt)
+  const deletionWarning = wrapText(
+    `Project ${record.projectId} and all its content will be permanently deleted on ${deadline} ` +
+      'unless you claim it. Other projects may also require action.',
+  )
+    .replace(record.projectId, styleText('cyan', record.projectId))
+    .replace(deadline, styleText('yellow', deadline))
```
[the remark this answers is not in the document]
shapirodaniel: nice catch bugbot, we can remove this highlight and let the wrapped date be the default color

packages/@sanity/cli/src/actions/init/initStudio.ts
```diff
@@ -82,11 +84,13 @@ export async function initStudio({
...
-  } catch (err) {
-    // Non-critical update
-    debug('Failed to update cliInitializedAt metadata', err)
+  if (!preclaim) {
+    try {
+      await updateProjectInitializedAt(projectId)
+    } catch (err) {
+      // Non-critical update
+      debug('Failed to update cliInitializedAt metadata', err)
+    }
```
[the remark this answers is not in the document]
shapirodaniel: this finding doesn't seem valid, looks like it's running against a prior commit that did reference `updateInitialTemplateMetadata`? happy to be wrong here but i'm not seeing that method in the current diff

jwoods02 (commented):
Couple really minor things and 1 copy run then LGTM!

For copy I think it's a little wordy atm and lots of links around, not obvious what is important. The "Run a CLI command:" pre-scaffold is confusing as the token is so long it's difficult to understand.

Pre-scaffold I would print the claim link and token then quickly explain them e.g.:

Token: token
Claim link: link

Claim your project by 3 August 2026, 10:34 UTC to avoid losing it

To use the Sanity CLI before claiming use your token by setting SANITY_AUTH_TOKEN="token" sanity "command".


Then after scaffold change the token text to something like:

Your content is private until you claim, to read it, you need the token. Treat your token as a password and don't expose it publicly in your app.

The claim link and token are printed above but you can recover them with `sanity projects unclaimed` before the timer run out on "time".

For more information on how claiming works, or how to build your app visit https://sanity.new"

So we're not duplicating any knowledge, non scaffolded projects still get their info and it's more scannable IMO

Current state:
Image

packages/@sanity/cli/src/hooks/prerun/claimReminder.ts
shapirodaniel: great call outs all around, happy to defer reminders for now. my sense after going through a few rounds on reminders, it's prob better to instruct the agent to reminder the person they're working for that 

- their current project, and possibly others, are about to expire, so they should surface a claim url 
- `sanity projects unclaimed [--project-id]` is the right tool

fixed up in [4c191c0](https://github.com/sanity-io/cli/pull/1622/commits/4c191c0860386cee2da32f7f801adbc384d02903)

shapirodaniel:
> For copy I think it's a little wordy atm and lots of links around, not obvious what is important.

great edits! for viz, i ran `sanity new` after rebuilding local cli to point to [ae9a162](https://github.com/sanity-io/cli/pull/1622/commits/ae9a162df700d4927ce109b8c88b9848f985be1f)

Screenshot 2026-07-31 at 8 32 36 AM

packages/@sanity/cli/src/commands/new.ts
```diff
@@ -0,0 +1,376 @@
...
+      flow.result(`Dataset: ${styleText('cyan', project.datasetName)} (where your content lives)`)
+      flow.result(`Token: ${styleText('cyan', project.token)}`)
+      flow.link(project.claimUrl, {label: 'Claim link:'})
+      flow.gap()
+      flow.highlight(
+        `Claim your project by ${styleText('yellow', formatClaimDeadline(project.expiresAt))} to avoid losing it`,
+      )
+      flow.gap()
+      flow.line('To use the Sanity CLI before claiming, set SANITY_AUTH_TOKEN:')
+      flow.command('SANITY_AUTH_TOKEN="token" sanity "command"')
```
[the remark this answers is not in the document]
shapirodaniel: that's expected, agents/humans will interpolate the real auth token value: we're aiming to cut down on visual noise

packages/@sanity/cli/src/commands/new.ts
shapirodaniel: nice catch! fixed up in a9bd53229b0f33ed857479b01026fcb6d23f7f26

packages/@sanity/cli/src/commands/projects/unclaimed.ts
```diff
@@ -0,0 +1,177 @@
...
+        }
+        continue
+      }
+      if (status === 'unknown') {
+        this.output.warn(
+          `Could not verify whether project "${record.projectId}" is claimed. Local recovery details were kept. Run this command again to retry.`,
+        )
+      }
+      records.push(record)
+    }
```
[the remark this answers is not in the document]
shapirodaniel: nice catch, fixed up in https://github.com/sanity-io/cli/pull/1622/commits/d2ba6004feef94578f98836f13bf58049f1320c6 by expiring projects we know have passed the deadline

jwoods02 (approved):
Nice!!! Looks great to me, LGTM 🎉

PR #1624, opened 2026-07-30
feat(cli): show synced Shopify metafields in the Shopify templates

madken:
### Description

Sanity Connect imports product and collection metafields into `store.metafields` (BIL-859, BIL-860). Neither Shopify template declared that field, so the values synced but never appeared, and the Studio treated them as unknown fields. Both templates now show them in a read-only, collapsed **Metafields** fieldset.

Two decisions worth knowing about, since neither is obvious from the diff:

**`value` is not declared as a field.** Shopify's metafield type determines its shape, so one array holds strings, numbers, booleans, lists and objects. Sanity has no field type that accepts all of those, and a declared field is type-checked even when hidden, so declaring it reports `Expected type "String", got "Number"` on most products. Undeclared fields are only flagged outside the Studio. `components/inputs/ShopifyMetafield.tsx` reads the value from the raw object instead.

**Display is extensible rather than exhaustive.** `utils/formatMetafieldValue.ts` holds a map keyed by metafield type, with six entries as examples: `dimension`, `weight`, `volume`, `money`, `rating` and `list.single_line_text_field`. Anything not listed falls back to the raw value, so unlisted metafields still display. Customers add their own with one line, and the file links to Shopify's list of data types. The six chosen are the ones the fallback handles badly; scalars like `single_line_text_field` and `number_integer` already read correctly without an entry.

The storefront template also gains a `shop` object type. Three of its types declare a `shop` field but no such type existed, so `sanity schemas validate` reported three `Unknown type: shop` errors on a freshly scaffolded studio. Unrelated to metafields — split it out if you prefer.

The studio init e2e suite now scaffolds both templates, which it never did before.

Closes the Studio-template follow-up on [BIL-842](https://linear.app/sanity/issue/BIL-842), tracked in [BIL-947](https://linear.app/sanity/issue/BIL-947).

### What to review

Whether an undeclared field plus a small input component is the right trade for not shipping validation errors. The alternative — declaring `value` — is simpler to read but produces 12 errors against real data.

Each `metafields` array must keep exactly one member type. Adding a second one stops existing synced data from resolving. Commented in place.

### Testing

`src/actions/init/templates/__tests__/formatMetafieldValue.test.ts` — 11 tests covering the six mapped types, the raw-value fallback, `null` and `undefined`, and a check that the two template copies stay byte-identical. Run with `pnpm test:unit formatMetafieldValue`.

It sits outside `templates/` because `copy()` copies that directory recursively with no filtering, so a colocated test would land in every scaffolded studio and ship in the package tarball. Importing it from the test tree also pulls that one file into `check:types`, which `templates/` is otherwise excluded from. The schema types and the input component stay untested; a component test would need React testing, which this repo does not have.

`init.studio.shared.ts` now scaffolds both Shopify templates, which the e2e suite never built before. That is the guard for dependency resolution — the plugin pins that broke `sanity init` would have failed here. It adds four real `npm install` runs, since the shared helper runs for both the `-y` and unattended modes. It does not cover schema validity: no e2e test runs `sanity schemas validate`, so the missing `shop` type would still have slipped through.

Also verified manually by scaffolding both templates with the locally built CLI against a real project, using a plain `npm install`: both `init` runs exit 0, `sanity schemas validate` reports 0 errors, and `sanity documents validate --level error` reports 4 valid documents and 0 errors against a dataset holding 14 metafields across every value shape.


madken:
@indrekkarner Can you take a look at this when you have the time? Any thoughts on the solution?

indrekkarner:
> @indrekkarner Can you take a look at this when you have the time? Any thoughts on the solution?

One solution is to add different types and array include all:
shopifyMetafieldString
shopifyMetafieldNumber
...
And
const commonFields = [
// Namespace
    defineField({
      title: 'Namespace',
      name: 'namespace',
      type: 'string',
    }),
    // Key
    defineField({
      title: 'Key',
      name: 'key',
      type: 'string',
    }),
]

const shopifyMetafieldString = defineType({
  name: 'shopifyMetafield',
  title: 'Shopify Metafield',
  type: 'object',
  fields: [
    ...commonFields,
    defineField({ name: 'value', title: 'Value', type: 'string' }),
  ],
})


and usage
of: [{type: 'shopifyMetafieldString'}, {type: 'shopifyMetafieldNumber'}],

madken:
@indrekkarner Thanks. I built your suggestion and tested it, and it does work, with one condition and one limit.

I made `shopifyMetafieldString`, `Number`, `Boolean` and `Dimension` sharing `commonFields`, listed all four in `of`, and validated a document whose metafields each carry a matching `_type`. Result: 0 errors, 0 warnings. `value` is properly typed per shape, and you get real TypeGen types, which the generic version cannot give.

The condition is that every member needs a `_type` matching one of the declared types. Synced metafields have a `_key` but no `_type`, and Sanity only falls back to the single member type when `of` has exactly one entry. So with the data as it is written today, a multi-type array resolves nothing. That also explains a misleading result: with two member types and no `_type`, `documents validate` reports 0 errors and 0 warnings, but only because the items were skipped. With one member type the same data reports 21 warnings, which is the proof it resolved.

The limit is `json` and `rich_text_field`. A typed variant needs `value` declared as an object, and Sanity rejects that:

```
[ERROR] [shopifyMetafieldJson]
  value  ✖ Object should have at least one field
```

Those values have arbitrary keys, so there is nothing to declare.

I also tried to get both at once, a catch-all member alongside typed ones, by registering the generic type under the name `object`, which is what Sanity resolves a `_type`-less object to. That is explicitly blocked: `Array members can not be given the same name as a built-in type`.

What I have done instead is take the display half of your idea. `utils/formatMetafieldValue.ts` has a map keyed by metafield type with five examples, `dimension`, `weight`, `money`, `rating` and `list.single_line_text_field`, and anything not listed falls back to the raw value. Adding a type is one line, and the file links to Shopify's list of data types. So `{"value":180,"unit":"CENTIMETERS"}` now shows as `180 CENTIMETERS` without needing a schema type per shape.

Typed schema fields are still the better end state for everything except `json`. That needs the importer to emit a per-shape `_type` first, so I have left it out of this PR.

indrekkarner (dismissed):

indrekkarner (approved):

madken:
@indrekkarner Your approval was dismissed automatically when I pushed more commits — nothing in the design changed. You have already reviewed that part, so here is just what landed since:

- Unit tests for `formatMetafieldValue`.
- Both Shopify templates added to the studio init e2e suite. CI never built them before, which is why the plugin pins that broke `sanity init` reached review.
- Removed the comment above each `metafields` array.
- Merges with main. The plugin pins now match what main already landed in #1643.

Nothing here touches the schema shape or the value display. Thanks for the re-approval.

PR #1633, opened 2026-07-30
feat(tokens): add expiry option and migrate token management to the Access API

runeb:
### Description

`sanity tokens create` previously prompted for a label and role but offered no way to set an expiration, and `tokens list` gave no visibility into whether a token expires. This PR adds expiry support to the token commands and, to enable it, migrates all token management from the deprecated Projects API tokens endpoints to the Access API robots endpoints.

**Expiry**

- `tokens create` now includes an expiration step in the interactive flow, with the same choices and default as the Manage token creation flow: No expiration (default), 30/60/90 days (labeled with the resolved date), or a custom date
- A new `--expires-at YYYY-MM-DD` flag sets the expiry non-interactively (validated as today or later); unattended mode defaults to no expiration
- `tokens create` output includes `Expires: <date>` (or `Never`), and `tokens list` gains an `Expires` column

**Access API migration**

- `services/tokens.ts` now uses `access/project/{id}/robots` and `access/project/{id}/roles` (with cursor pagination) instead of `projects/{id}/tokens` and `projects/{id}/roles`
- Listing filters robots to those managed by the project, matching what the old endpoint returned (robots managed by other resources can hold project memberships but are not project tokens)
- The Access API identifies robots by a robot ID, while the old Projects API exposed what is now the robot's `tokenId`. Tokens carry both: `id` (robot ID, displayed and used for operations) and `tokenId` (legacy ID) in `--json` output
- `tokens delete` accepts either form: it tries the given ID as a robot ID first, and on 404 resolves it as a legacy token ID against the robot list — so IDs saved from earlier CLI versions keep working
- If the project roles cannot be fetched (used only to resolve role display titles), `tokens list` degrades to showing role names instead of failing

Note: `--json` output shape changed accordingly — it now includes `id`, `tokenId`, `expiresAt`, and `roles`, and no longer includes Projects-API-only fields such as `projectUserId` and `permissions`.

### What to review

- `packages/@sanity/cli/src/services/tokens.ts` — the endpoint migration, pagination, `managedBy` filtering, and the delete ID-resolution fallback
- `packages/@sanity/cli/src/actions/tokens/expiry.ts` — expiry choice/validation/formatting helpers
- `packages/@sanity/cli/src/commands/tokens/{create,list,delete}.ts` — prompt flow, flag handling, and output changes
- Manual check: run `sanity tokens create`, `tokens list`, and `tokens delete` against a project; try deleting by both the listed ID and a legacy token ID from an older CLI version

### Testing

- Rewrote the three command test suites against the Access API endpoints, including new coverage for the expiry prompt choices, preset/custom/flag expiry paths, `--expires-at` validation, pagination, `managedBy` filtering, legacy-ID delete resolution, and the roles-fetch fallback
- Added unit tests for the expiry helpers
- Full `@sanity/cli` unit suite passes (the only failures are network-simulation tests that also fail on `main` in the sandbox environment used for development)

### Notes for release

`sanity tokens create` now offers an expiration setting for new tokens — interactively or via the `--expires-at YYYY-MM-DD` flag — and `sanity tokens list` shows when each token expires.

---

https://claude.ai/code/session_019UvuCLVE5Rw17DfrRQhyVH

### DIFF
--- a/.changeset/pr-1633.md
+++ b/.changeset/pr-1633.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+feat(tokens): add expiry option and migrate token management to the Access API
--- a/packages/@sanity/cli/src/actions/tokens/__tests__/expiry.test.ts
+++ b/packages/@sanity/cli/src/actions/tokens/__tests__/expiry.test.ts
@@ -1,0 +1,69 @@
+import {describe, expect, test} from 'vitest'
+
+import {
+  formatExpiryChoiceDate,
+  formatTokenExpiry,
+  getPresetExpiryDate,
+  TOKEN_EXPIRY_PRESET_DAYS,
+  validateExpiryDate,
+} from '../expiry.js'
+
+describe('tokens expiry helpers', () => {
+  test('exposes the same preset choices as the Manage token creation flow', () => {
+    expect(TOKEN_EXPIRY_PRESET_DAYS).toEqual([30, 60, 90])
+  })
+
+  test('getPresetExpiryDate resolves days from now to a yyyy-MM-dd date', () => {
+    const now = new Date('2026-07-30T12:00:00Z')
+    expect(getPresetExpiryDate(30, now)).toBe('2026-08-29')
+    expect(getPresetExpiryDate(60, now)).toBe('2026-09-28')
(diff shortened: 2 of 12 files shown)


runeb:
Superseded by #1646, which lands the Access API migration together with expiry support and a new `tokens rotate` command.

PR #1635, opened 2026-07-31
fix(deploy): align federated studio deploys with the workbench app path

mttdnt:
### Description

Federated studios (`unstable_defineApp` + a `sanity.config.*`) reach the workbench deploy path without several behaviours core apps already had. This closes those gaps so a basic federated studio deploys the same way an app does. Part of SDK-1045.

- **Auto-updates are silently disabled.** A federated studio takes `sanity` from the workbench shell's shared runtime, so it can't resolve its own version. The build previously ran the whole auto-updates preamble and threw the result away — resolving an import map it discards, and prompting about version drift for a feature that never applies (which could abort the build). It now skips the preamble, and the deployment reports `isAutoUpdating: false` instead of telling Brett it auto-updates. A `warn` check notes it isn't supported yet, with a TODO to confirm the long-term story with the studio team.
- **Slug collisions report the existing app id.** `resolveWorkbenchStudio` now does the same `listApplications` → `slug-taken` lookup `resolveWorkbenchApp` has, so a first deploy onto a taken slug names the existing app and the `deployment.appId` edit to reuse it, rather than failing with a raw server error.
- **`visibility` is honoured.** Declared on `unstable_defineApp`, it was read by the config resolver and then dropped by the studio path. It's now forwarded on create and patched on redeploy, matching the app path.
- **Output is accurate.** The deploy spinner no longer says "Deploying to sanity.studio" while printing a `sanity.run` URL, and the dry run says "Would create a new studio … with slug …" instead of calling a slug a hostname (and no longer claims availability is checked later, since the resolver now checks it).

### What to review

All changes are scoped to the workbench branch of the deploy path. Legacy studio and external (`--external`) deploys are untouched — worth confirming, since `describeStudioTarget` and `resolveDeployTarget` are shared:

- `deployStudio.ts` — the `if (workbench && !isExternal)` branch only
- `resolveDeployTarget.ts` — `resolveWorkbenchStudio` gains `organizationId` + the collision lookup; the `slug-taken` variant added to `StudioDeployTargetResolution`
- `deployChecks.ts` — new `slug-taken` case and an `isWorkbench` branch in `would-create`. The legacy/external branch is behaviourally identical (`url` was already `studioUrl(appHost)` whenever `isWorkbench` was false)
- `deployWorkbenchApp.ts` — `createStudio` accepts `visibility`, mirroring `createCoreApp`
- `buildStudio.ts` — one condition; the later `&& !options.isWorkbenchApp` guard became redundant and was removed

### Testing

Unit tests added alongside the existing app-path equivalents:

- `buildStudio.test.ts` — 4 cases: builds pinned with no import map, preamble skipped (no `checkAppId`, no import-map resolution), never prompts on version drift, no auto-updates build log
- `resolveDeployTarget.test.ts` — 5 cases for `resolveWorkbenchStudio`: `appId` short-circuits without listing, slug free → `would-create`, slug taken → `slug-taken` with the existing id, a `coreApp` holding the slug also yields `slug-taken` (apps and studios share one slug namespace), no org → no listing
- `deployChecks.test.ts` — the `slug-taken` render, and the workbench `would-create` message/target
- `deployWorkbenchApp.test.ts` — 2 cases mirroring `createCoreApp`: `visibility` forwarded when declared, omitted when not

`check:types`, `check:lint`, `check:deps` and the unit suite pass. The `build.studio` integration tests (including the `federated-studio` fixture, which sets `autoUpdates: true`) also pass. Verified manually by deploying the fixture.

Not covered: `deployStudio.ts` still has no unit coverage of its flow — neither does `deployApp.ts` (18%, only its two exported helpers), so this doesn't regress parity. Left for follow-up.

---

> [!NOTE]
> **Medium Risk**
> Changes the federated studio deploy and auto-update resolution paths; scope is limited to the workbench branch, but deploy failures and build pinning behavior are user-visible.
> 
> **Overview**
> Brings **federated studio** deploy/build behavior in line with workbench **core apps** on the workbench-hosted path (legacy `sanity.studio` and `--external` flows are unchanged).
> 
> **Auto-updates:** Workbench apps no longer silently pretend auto-updates apply. `resolveAutoUpdates` treats them as **unsupported**, warns when configured, and build/deploy run with **`autoUpdatesEnabled` / `isAutoUpdating: false`** so installed package versions are used.
> 
> **Slug collisions:** `resolveWorkbenchStudio` now lists org applications and returns **`slug-taken`** (same as apps). Deploy checks and dry-run use a shared **`describeSlugTaken`** diagnosis with the existing app id and `deployment.appId` fix; taken slugs block deploy before create/build.
> 
> **Visibility:** `unstable_defineApp` **`visibility`** is passed through **`createStudio`** and **`deployWorkbenchApp`** on federated studio create/redeploy.
> 
> **Reporting:** Federated studio messaging uses slug-based “would create” copy (not hostname), deploy label says “Deploying studio”, JSON **`target`** includes **`slug`** on first create, and dry-run plans ignore **`target`** from **failing** checks.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f16dda9f3a42fe9211ceecc30e2b52ee2219e7d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


packages/@sanity/workbench-cli/src/services/applications.ts
joshuaellis, line 1: this isn't true, because studios and coreApps currently still run on the `*.sanity.stuidio` domain it needs to be a global check imo

joshuaellis (dismissed):

packages/@sanity/workbench-cli/src/services/applications.ts
gu-stav: You're right, and Brett confirms it: `NonSingletonSlugUnique` is a global unique index, so an org-scoped scan cleared slugs a create would still be rejected for.

Added `isSlugAvailable`, which asks `GET /studiohosts/:slug/availability` — the same rule both of Brett's create paths are held to (live studio via `findStudioByHost`, or registry slug via `hostedSlugExists`, singletons excluded).

The org scan stays and runs first, for two reasons. It names the existing app id so you get `deployment: {appId: '...'}` to paste rather than just being told the slug is gone. And singleton slugs are unique per org, not globally (`SingletonSlugUniquePerOrg`), so the global check reports them available — the org scan is what still catches a Media Library redeployed without an appId.

So: holder in your org gives the id, taken anywhere else fails with `The slug "x" is already taken`.

joshuaellis: 😢

gu-stav: Sorry Claude was too eager - didn't see it was still looking at the bugbot reports in the background and included yours

gu-stav:
@mttdnt I've advanced the PR a bit further and mainly moved more of the code into the `workbench-cli` package and removed the changes in the build package, which were imo not necessary.

packages/@sanity/workbench-cli/src/services/applications.ts
gu-stav: Reverted - checks are global again.

packages/@sanity/cli/src/actions/deploy/resolveDeployTarget.ts
```diff
@@ -184,6 +184,7 @@ export async function resolveWorkbenchApp({
     const existing = (await listApplications(organizationId)).find((app) => app.slug === slug)
     if (existing) {
       return {
+        appHost: existing.slug ?? '',
```
gu-stav: Shouldn't need a fallback, because `slug` is mandatory?

gu-stav (approved):

PR #1642, opened 2026-07-31
feat(cli): show terms of service in sanity new

jwoods02:
### Description

`sanity new` creates a real, working project without an account, so there's no signup step where anyone accepts our terms. Consent is implicit in using the project, which only holds up if we actually say so in the output.

The provisioning API now returns a terms of service notice and URL alongside a new project, and `sanity new` prints both just above the closing link. `--json` carries the same thing under a new `terms` key, since that path prints nothing else and is what agents read.

The CLI treats the field as optional rather than required. If the API doesn't send it, the notice falls back to a bundled copy, so it can't silently vanish from the output and a CLI release doesn't have to line up with an API deploy.

### What to review

- Placement in `new.ts`: the notice sits after the token and claim caveats, above the `visit sanity.new` outro.
- `parseProvisionResponse` in `services/mintProject.ts`: `terms` is deliberately outside the required-field check, so a response without it still mints.
- `terms` is a new field on the `NewProjectResult` that `--json` prints, so anything consuming that output sees a new key.

### Testing

Unit tests cover the printed lines for the scaffolded and `--no-scaffold` paths, the `--json` shape, and the fallback when the response omits `terms`.

To see it by hand without minting a real project, point the CLI at a local stub of the provision endpoint with `SANITY_API_HOST` and set `SANITY_CLI_CONFIG_PATH` to a throwaway file, then run `sanity new "Test" --no-scaffold --yes` and again with `--json`. I ran all four combinations that way (terms present and absent, both flags).

### Notes for release

`sanity new` now prints the terms of service you accept by using an unclaimed project, and includes them in `--json` output.

---

> [!NOTE]
> **Low Risk**
> Additive CLI output and an optional JSON field with safe API fallbacks; no changes to auth, provisioning requirements, or stored recovery records.
> 
> **Overview**
> **`sanity new`** now surfaces Terms of Service after the token/claim guidance and before the `sanity.new` outro: the notice text plus a labeled link, for both scaffolded and `--no-scaffold` flows.
> 
> Provisioning responses can supply `terms.notice` and `terms.url`; the mint layer maps these onto **`MintedProject`** and uses **bundled fallbacks** when the API omits or sends invalid/empty terms, without blocking project creation. **`--json`** adds a **`terms`** object (`notice`, `url`) on **`NewProjectResult`** for the same data agents and scripts consume.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7e92351c202973c4e201d9db28d50c24c87654a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1642.md
+++ b/.changeset/pr-1642.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': minor
+---
+
+feat(cli): show terms of service in sanity new
--- a/packages/@sanity/cli/src/commands/__tests__/new.test.ts
+++ b/packages/@sanity/cli/src/commands/__tests__/new.test.ts
@@ -37,6 +37,8 @@ const project = {
   datasetName: 'production',
   expiresAt: '2026-08-01T00:00:00.000Z',
   resourceId: 'abc123',
+  termsNotice: 'By continuing to use this project you accept the Sanity Terms of Service.',
+  termsUrl: 'https://www.sanity.io/legal/tos',
   token: 'sk-robot-token',
 }
 
@@ -48,6 +50,7 @@ const result = {
   dataset: project.datasetName,
   expiresAt: project.expiresAt,
   projectId: project.resourceId,
+  terms: {notice: project.termsNotice, url: project.termsUrl},
   token: project.token,
 }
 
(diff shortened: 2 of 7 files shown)


shapirodaniel (approved):
looks great!

PR #1567, opened 2026-07-20
fix: link to project members page when projectUserNotFoundError is thrown

jonahsnider:
When a `projectUserNotFoundError` error is thrown, we now include a link to the page where you can update what members have access to the project. This makes it clearer what the next steps are to resolve the error

---

> [!NOTE]
> **Low Risk**
> Localized CLI error-message handling with a type guard; no auth or API behavior changes beyond clearer 401 messaging for one error type.
> 
> **Overview**
> **Project-scoped CLI API errors** now distinguish `projectUserNotFoundError` from a generic 401: instead of suggesting `sanity login`, the message points users to the **project members** page for that `projectId` (via `getSanityUrl`).
> 
> The `authErrors` middleware takes an optional `projectId` (passed from `getProjectCliClient` only); global clients keep the existing login/help flow. **401 detection** is aligned with `@sanity/client` by reading `err.statusCode` rather than `response.body.statusCode`. Tests cover the new message and updated error shapes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dbb868a54ddb757cb1ef9ee6fe52142bfae83d4d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #1644, opened 2026-07-31
fix(cli-core): explain missing project membership in Content Lake 401 errors

runeb:
### Description

Solves [AIGRO-5281](https://linear.app/sanity/issue/AIGRO-5281/project-creators-user-id-mapping-never-reached-redis-content-lake-401s).

The Content Lake rejects requests from authenticated users who are not explicit project members with a bare 401: `project user not found for user ID "..." in project "..."`. As the discovery on the ticket showed, this is a legitimate (and confusing) situation: the management API honors organization-level roles (e.g. org administrator), while the data plane requires explicit project membership — so a user can list projects, manage CORS, and deploy schemas, yet be denied on every content request with no hint of why or what to do.

PR #1567 added a members-page link for this case, but it only matched the structured `{error: {type: 'projectUserNotFoundError'}}` body, and only on project-scoped clients. This PR extends the `authErrors` middleware in `@sanity/cli-core` to:

- Detect the membership case on all clients: the structured `projectUserNotFoundError` body type first, with the Content Lake's plain-text sentence as an anchored fallback (`project user not found for user ID "..." in project "..."`). Unrelated 401 prose cannot trigger it; if the wording ever changes, detection degrades to the existing generic 401 guidance rather than misfiring. If the platform confirms the structured type is always present on this error, the sentence fallback can be deleted and detection is purely contract-based.
- Explain the actual situation and give a CLI-native fix first: `sanity users invite` goes through the management plane, which honors org-level roles, so the affected persona can self-serve without a browser (important for agents). The members-page link remains as the second path. All real occurrences arrive on project-scoped clients, which know the project id; the hypothetical no-id case uses placeholders and the Manage root.

Resulting message:

```
project user not found for user ID "gZmyPK60G" in project "0zvlb7pj". This account is not a member of the project. Organization-level roles do not grant access to project content. Invite this account with sanity users invite <email> --project-id 0zvlb7pj --role <role> or add it as a project member at https://www.sanity.io/manage/project/0zvlb7pj/members.
```

This also reaches MCP-driven invocations via `invokeSanityCli`, which relays the enriched error message. Note the MCP command policy currently denies `users:invite`, so MCP-hosted agents can read the guidance but not execute it in-process — whether to allow it there is a separate product/security decision.

An active-probe design (on 401: check session validity, then project membership via the management API) was considered and rejected: the typed 401 from the data plane already states both facts, the probes themselves require grants/session types the affected personas may lack, and asking populus to confirm membership reintroduces the exact management-plane/data-plane divergence this incident is about. A follow-up worth its own ticket: detecting the member-but-not-yet-propagated case (the original sync-lag theory) genuinely requires comparing planes.

### What to review

- `packages/@sanity/cli-core/src/apiClient.ts` — the `authErrors` middleware: `isProjectUserNotFoundError` (structured type or anchored sentence) and the new copy.
- Whether the anchored sentence fallback is acceptable until the data plane's typed contract is confirmed.
- Whether pointing at `sanity users invite` is the right first remediation (it sends an email invitation; direct member addition is only available in Manage).

### Testing

Unit tests in `packages/@sanity/cli-core/src/__tests__/apiClient.test.ts` cover: the structured error type without the message text, the plain-text Content Lake sentence on project-scoped and global clients, the Manage-root fallback when no project id is known, and a negative case proving unrelated 401 prose still gets the login guidance. Full `@sanity/cli-core` unit suite passes (353/353); `pnpm test:unit --changed` passes apart from three pre-existing sandbox-environment failures in network-error simulation tests, which fail identically on `main`.

### Notes for release

`sanity` CLI errors from the Content Lake now explain when access was denied because the account is not a project member (for example, when it only has organization-level roles), suggest `sanity users invite` to fix it from the terminal, and link to the project's members page in Manage.

### DIFF
--- a/packages/@sanity/cli-core/src/apiClient.ts
+++ b/packages/@sanity/cli-core/src/apiClient.ts
@@ -165,8 +165,9 @@ function authErrors(projectId?: string) {
 
       const statusCode = isHttpError(err) && err.statusCode
       if (statusCode === 401) {
-        if (isProjectUserNotFoundError(err.response.body, err.message)) {
-          const membersProjectId = projectId ?? getProjectIdFromMessage(err.message)
+        const membership = detectMissingProjectMembership(err)
+        if (membership) {
+          const membersProjectId = projectId ?? membership.projectId
           const membersUrl = membersProjectId
             ? getSanityUrl(`/manage/project/${encodeURIComponent(membersProjectId)}/members`)
             : getSanityUrl('/manage')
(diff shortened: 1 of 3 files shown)


packages/@sanity/cli-core/src/apiClient.ts
```diff
@@ -178,9 +183,40 @@ function authErrors(projectId?: string) {
...
-function isProjectUserNotFoundError(
-  body: unknown,
-): body is {error: {type: 'projectUserNotFoundError'}} {
+// Detection fallback for Content Lake responses that carry the
+// missing-membership 401 only as prose. Anchored to the full sentence so
+// unrelated 401s can't trigger; if the wording ever changes, detection
+// degrades to the generic 401 guidance below rather than misfiring.
+const PROJECT_USER_NOT_FOUND_MESSAGE =
+  /\bproject user not found for user ID "[^"]*" in project "[a-zA-Z0-9-]+"/
+
```
runeb, line 194: I dont like this matching, but there doesnt appear to be a more structured way to detect this

PR #1646, opened 2026-08-01
feat(tokens)!: migrate token management to the Access API, add expiry and rotate

runeb:
### Description

Migrates the `sanity tokens` commands (`create`, `list`, `delete`) from the Projects API tokens endpoints to the Access API, and adds the capabilities that unblocks:

- **Token expiry**: `tokens create --expires-at <date|timestamp>`, an interactive expiry step with the same presets as sanity.io/manage (never/30/60/90 days/custom date), and an Expires column in `tokens list`.
- **`sanity tokens rotate`**: replaces a token's secret while keeping its roles and expiry. The token is read from stdin (`echo "$SANITY_TOKEN" | sanity tokens rotate`) to keep secrets out of shell history; `-t, --token` is also accepted. The previous secret is revoked immediately.
- Token listing follows the Access API's cursor pagination, and excludes org-managed tokens since they cannot be managed at project scope.

**Breaking**: `--json` output now returns the Access API responses verbatim instead of the legacy Projects API shapes (see release notes below). Human-readable output is unchanged, except `tokens list` shows role names instead of titles and the "Token ID" column is now "ID".

Fixes [AIGRO-5284](https://linear.app/sanity/issue/AIGRO-5284/add-expiry-option-to-sanity-tokens-create). Supersedes #1633.

### What to review

- `services/tokens.ts` — all Access API calls (endpoints, pagination, rotate) live here.
- `commands/tokens/` — `create` (expiry flag + prompt), `list` (columns), `rotate` (new), `delete` (unchanged flow, robot ids).
- `actions/init/bootstrapRemoteTemplate.ts` — uses the same `createToken` service; call sites updated.

### Testing

- Unit tests cover the exact request bodies, pagination across cursors, response passthrough, the org-managed filter, expiry validation, the interactive expiry/rotate flows, and 401/403 error mappings for rotate. Coverage on the migrated files is 100%.
- Smoke-tested the full lifecycle against staging: create (with and without expiry) → authenticate → list → rotate (old secret rejected, new accepted, personal token correctly refused) → delete.

### Notes for release

The `sanity tokens` commands are now backed by the Access API and gain token expiry support and a `sanity tokens rotate` command.

- `tokens create --expires-at <date>` sets an expiry; interactive runs offer the same presets as sanity.io/manage. `tokens list` shows expiry in a new Expires column.
- `echo "$SANITY_TOKEN" | sanity tokens rotate` replaces a token's secret in place, revoking the previous secret immediately.
- **Breaking**: `--json` output now returns the Access API token shape verbatim. For `tokens create` the secret moved from `key` to `token`; for `tokens list`, `roles`/`createdBy`/`permissions`/`projectUserId`/`lastUsedAt` are replaced by `memberships`, and `id` is the token's stable identifier (used by `tokens delete`). `tokens list` also shows role names instead of display titles and excludes organization-managed tokens.

---

> [!NOTE]
> **High Risk**
> Major breaking change to token JSON contracts and credential identifiers, plus new rotate/revoke behavior that can immediately invalidate production API tokens if misused.
> 
> **Overview**
> **Breaking:** `tokens create`, `list`, and `delete` now call the **Access API** (`/access/project/.../robots` and roles) instead of the Projects API. `--json` echoes the new **robot** shape (secret on `.token`, roles in `.memberships[].roleNames`; several legacy fields dropped). Human list output uses role **names**, an **ID** column, and omits org-managed tokens. Init/bootstrap still uses `createToken` but reads `.token` instead of `.key`.
> 
> **New:** `tokens create --expires-at` plus interactive expiry presets (never / 30–90 days / custom). `tokens list` adds an **Expires** column. **`sanity tokens rotate`** posts to `/access/robots/me/rotate` with the current secret from stdin or `--token`, revoking the old secret immediately; MCP policy denies it like other token mutators.
> 
> Shared **`readTokenFromStdin`** replaces duplicated login stdin handling. API version constant switches to `v2025-07-11` with cursor pagination for list/roles.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1c5a9ad7bb1063aea0112b07a9084268e859f9a9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/pr-1646.md
+++ b/.changeset/pr-1646.md
@@ -1,6 +1,0 @@
-<!-- auto-generated -->
----
-'@sanity/cli': major
----
-
-feat(tokens)!: migrate token management to the Access API, add expiry and rotate
--- a/.changeset/pr-1646.md
+++ b/.changeset/pr-1646.md
@@ -1,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': major
+---
+
+feat(tokens)!: migrate token management to the Access API, add expiry and rotate
--- a/packages/@sanity/cli/src/commands/tokens/rotate.ts
+++ b/packages/@sanity/cli/src/commands/tokens/rotate.ts
@@ -1,5 +1,6 @@
 import {Flags} from '@oclif/core'
 import {exitCodes, SanityCommand, subdebug} from '@sanity/cli-core'
+import {getErrorMessage} from '@sanity/cli-core/errors'
 import {isHttpError} from '@sanity/client'
 
 import {rotateToken} from '../../services/tokens.js'
@@ -83,8 +84,9 @@ export class RotateTokenCommand extends SanityCommand<typeof RotateTokenCommand>
         )
       }
 
-      const err = error as Error
-      this.error(`Token rotation failed:\n${err.message}`, {exit: exitCodes.RUNTIME_ERROR})
+      this.error(`Token rotation failed:\n${getErrorMessage(error)}`, {
+        exit: exitCodes.RUNTIME_ERROR,
+      })
     }
   }
 
@@ -93,8 +95,7 @@ export class RotateTokenCommand extends SanityCommand<typeof RotateTokenCommand>
     try {
       token = await readTokenFromStdin(TOKEN_USAGE_HINT)
     } catch (error) {
-      const err = error as Error
-      return this.error(err.message, {exit: exitCodes.USAGE_ERROR})
+      return this.error(getErrorMessage(error), {exit: exitCodes.USAGE_ERROR})
     }
 
     if (!token) {
--- a/packages/@sanity/cli/src/commands/tokens/rotate.ts
+++ b/packages/@sanity/cli/src/commands/tokens/rotate.ts
@@ -1,0 +1,106 @@
+import {Flags} from '@oclif/core'
+import {exitCodes, SanityCommand, subdebug} from '@sanity/cli-core'
+import {isHttpError} from '@sanity/client'
+
+import {rotateToken} from '../../services/tokens.js'
+import {readTokenFromStdin} from '../../util/readTokenFromStdin.js'
+
+const tokensRotateDebug = subdebug('tokens:rotate')
+
+const TOKEN_USAGE_HINT =
+  'Token is required. Pipe it on standard input (`echo "$SANITY_TOKEN" | sanity tokens rotate`) to keep it out of shell history, or pass it with `--token`.'
+
+export class RotateTokenCommand extends SanityCommand<typeof RotateTokenCommand> {
+  static override description = 'Rotate an API token, replacing its secret with a new one'
+
+  static override examples = [
+    {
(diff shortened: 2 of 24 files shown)


.changeset/pr-1646.md
```diff
@@ -0,0 +1,6 @@
+<!-- auto-generated -->
+---
+'@sanity/cli': major
+---
+
+feat(tokens)!: migrate token management to the Access API, add expiry and rotate
```
jonahsnider: we should add some more context on what the breaking changes are and how to migrate

the two additions, adding expiry and token rotation should be separate changeset files so they get their own items in the changelog

packages/@sanity/cli/src/commands/tokens/rotate.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      if (isHttpError(error) && error.statusCode === 403) {
+        this.error(
+          'The token is not an API token and cannot be rotated. Personal tokens are renewed by logging in again.',
+          {exit: exitCodes.RUNTIME_ERROR},
+        )
+      }
+
+      const err = error as Error
+      this.error(`Token rotation failed:\n${err.message}`, {exit: exitCodes.RUNTIME_ERROR})
```
jonahsnider: instead of casting, can use the `getErrorMessage` helper here

packages/@sanity/cli/src/commands/tokens/rotate.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+  }
+
+  private async readTokenArg(): Promise<string> {
+    let token: string
+    try {
+      token = await readTokenFromStdin(TOKEN_USAGE_HINT)
+    } catch (error) {
+      const err = error as Error
```
jonahsnider: same note for `getErrorMessage`

jonahsnider (changes requested):

.changeset/pr-1646.md
runeb: Replaced the auto-generated changeset with three hand-authored ones in 1c5a9ad7b: a major for the migration with how-to-migrate notes (`key` → `token`, `memberships`, ids), and separate minors for expiry and rotate so each gets its own changelog entry. Manual changesets make the generate-changeset workflow skip, so the bot won't re-add its file.

packages/@sanity/cli/src/commands/tokens/rotate.ts
runeb: Done in 1c5a9ad7b.

packages/@sanity/cli/src/commands/tokens/rotate.ts
runeb: Done in 1c5a9ad7b.

jonahsnider (approved):

runeb:
@jonahsnider thanks for the review, both addressed now, appreciate another pass

jonahsnider (approved):

PR #1667, opened 2026-08-04
fix(workbench): use slug as app id in development

gu-stav:
### Description

A workbench app's bus identity and its registry entry came from two different ports. `__SANITY_APP_ID__` was compiled from the port the app *asks* for, while the registry advertised the port it *bound*. Apps attaching to a running workbench all reach for `workbenchPort + 1`, so whoever loses that race gets moved by Vite and keeps the old id in its bundle. **It doesn't just carry a stale id, it speaks on the bus as the app that won the port.**

The port can't be read in time: Vite patches `httpServer.listen` to run `initServer(true)` — and with it `depsOptimizer.init()` — *before* the socket binds, and `bus.ts` lives in those optimized deps. By the time a port exists the id is already inlined.

**For local apps the slug is the id now.** Applications clashing on the same slug are not send to the workbench remote.

---

> [!NOTE]
> **Medium Risk**
> Changes workbench federation bus identity and dev registry addressing; misconfigured duplicate slugs only warn and skip registration, but wrong-id bugs in multi-app dev were the motivation for the fix.
> 
> **Overview**
> **Workbench dev apps now use the configured `slug` as `__SANITY_APP_ID__` and as the dev-server registry id**, instead of a `host-port` string from `resolveAppId`. That keeps the Vite-inlined bus identity aligned with what the workbench advertises, avoids wrong-app collisions when ports shift, and stays stable across restarts.
> 
> `resolveAppId` is removed from `@sanity/workbench-cli` exports and from `getDevServerConfig` / `startDevServerRegistration`. **`registerDevServer` rejects duplicate slug ids** via `DevServerIdTakenError`; registration catches that, **warns**, and leaves the dev server running without a registry entry. Tests and the workbench e2e assert the slug in `/@vite/env`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aed29d40ffe9f0a3c444f2d5025af6dd3a627ee3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/dev-app-id-slug.md
+++ b/.changeset/dev-app-id-slug.md
@@ -1,0 +1,6 @@
+---
+'@sanity/cli': minor
+'@sanity/workbench-cli': minor
+---
+
+fix(workbench): key a dev app's id on its slug
--- a/packages/@sanity/cli-e2e/__tests__/dev/dev.workbench.test.ts
+++ b/packages/@sanity/cli-e2e/__tests__/dev/dev.workbench.test.ts
@@ -45,15 +45,15 @@ describe.skipIf(isRegistryMode)('sanity dev (workbench/federation)', {timeout: 1
     expect(output).toContain(`app on port ${port + 1}`)
 
     // The app server carries the app's bus identity (`__SANITY_APP_ID__`) for
-    // `@sanity/runtime` — in dev, the address it's served at (`<host>-<port>`),
-    // so a running app never collides with its deployed twin. Vite delivers
-    // `define` entries through the per-server `/@vite/env` module rather than by
-    // text replacement, so that module is where the id is observable.
+    // `@sanity/runtime` — its slug, so a running app never collides with its
+    // deployed twin. Vite delivers `define` entries through the per-server
+    // `/@vite/env` module rather than by text replacement, so that module is
+    // where the id is observable.
     const env = await fetch(`http://localhost:${port + 1}/@vite/env`)
     expect(env.ok).toBe(true)
     const envSource = await env.text()
     expect(envSource).toContain('__SANITY_APP_ID__')
-    expect(envSource).toContain(`"localhost-${port + 1}"`)
+    expect(envSource).toContain('"federated-studio"')
 
     session.sendControl('c')
     await session.waitForExit(15_000).catch(() => session.kill())
--- a/packages/@sanity/cli/src/actions/dev/servers/__tests__/getDevServerConfig.test.ts
+++ b/packages/@sanity/cli/src/actions/dev/servers/__tests__/getDevServerConfig.test.ts
@@ -17,7 +17,7 @@ describe('getDevServerConfig', () => {
     vi.unstubAllEnvs()
   })
 
-  test('inlines the workbench app id as `${host}-${port}`, and omits it for a plain project', () => {
+  test("inlines the workbench app's slug as its id, and omits it for a plain project", () => {
     const workbench = getDevServerConfig({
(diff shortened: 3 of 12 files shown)


joshuaellis:
I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?

gu-stav:
> I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?

The root of the issue is that an app can get a different port assigned by Vite than the one that has been inlined in the bundle which is why sometimes panels don't work, because their ids don't match.

I've tried to keep the port in the id but have not found a good way to do that.

joshuaellis:
> > I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?
> 
> The root of the issue is that an app can get a different port assigned by Vite than the one that has been inlined in the bundle which is why sometimes panels don't work, because their ids don't match.
> 
> I've tried to keep the port in the id but have not found a good way to do that.

rightttt, can we use the PID? or do you want it to be consistent between turning on / off servers (which im isn't that important, we could make this better one day)

gu-stav:
> > > I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?
> 
> > 
> 
> > The root of the issue is that an app can get a different port assigned by Vite than the one that has been inlined in the bundle which is why sometimes panels don't work, because their ids don't match.
> 
> > 
> 
> > I've tried to keep the port in the id but have not found a good way to do that.
> 
> 
> 
> rightttt, can we use the PID? or do you want it to be consistent between turning on / off servers (which im isn't that important, we could make this better one day)

We could use the pid if I think (need to check how Vite handles processes) if we are ok with having a different URL every time? I'm not quite sure what the benefit of that would be though, if we could use a stable identifier. I know we want to do a more complete migration at some point - if you prefer researching the pid idea I can look into that tomorrow?

mttdnt:
> > I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?
> 
> The root of the issue is that an app can get a different port assigned by Vite than the one that has been inlined in the bundle which is why sometimes panels don't work, because their ids don't match.
> 
> I've tried to keep the port in the id but have not found a good way to do that.

I looked at some potential other ways to deal with this:

1. Runtime injection: Similar to `__SANITY_STAGING__`, but that spreads across the runtime and the build path to. Too much plumbing for what it buys.
2. Pre allocate the port: We could try to lock the port, then generate the id. Though there is still a race condition as we would need to release it before vite would bind it.

I think your solution is the best for now.

joshuaellis:
> > > > I'm sorry, I don't understand the problem you're fixing. What's the root of the issue here?
> > 
> > 
> > > 
> > 
> > 
> > > The root of the issue is that an app can get a different port assigned by Vite than the one that has been inlined in the bundle which is why sometimes panels don't work, because their ids don't match.
> > 
> > 
> > > 
> > 
> > 
> > > I've tried to keep the port in the id but have not found a good way to do that.
> > 
> > 
> > rightttt, can we use the PID? or do you want it to be consistent between turning on / off servers (which im isn't that important, we could make this better one day)
> 
> We could use the pid if I think (need to check how Vite handles processes) if we are ok with having a different URL every time? I'm not quite sure what the benefit of that would be though, if we could use a stable identifier. I know we want to do a more complete migration at some point - if you prefer researching the pid idea I can look into that tomorrow?

well we want to move to slugs over ids in URLs anyway, so it wouldn't be different in the long run?

joshuaellis (approved):

PR #1675, opened 2026-08-05
fix(tokens): transparently resolve legacy token ids in tokens delete

runeb:
### Description

Token ids issued before the Access API migration (#1646) live on as the robot's `tokenId` — verified against the API. `tokens delete` now resolves them transparently: on a 404 it looks the id up in the robots list and deletes the matching robot, so ids stored from earlier CLI versions keep working. Also trims the now-moot id caveat from the unreleased migration changeset.

### What to review

`deleteToken` in `services/tokens.ts` — the 404 fallback path.

### Testing

Service and command tests cover the fallback sequence, no-match rethrow, and non-404 passthrough. Verified live against staging with a token created through the legacy Projects API and deleted by its legacy id.

---

> [!NOTE]
> **Medium Risk**
> Deletes API tokens via an extra list-and-match path on 404; wrong matching logic could delete the wrong robot, though scope is limited to failed direct deletes.
> 
> **Overview**
> **`tokens delete`** now accepts **pre–Access API token ids** without breaking scripts that still pass the old identifier.
> 
> `deleteToken` still deletes by the given id first; on **404** it loads the project robots, finds the robot whose **`tokenId`** matches, and deletes that robot instead. Non-404 errors skip the lookup. The unreleased migration changeset no longer warns that delete ids change with the new API.
> 
> Service and CLI delete tests cover the happy path, no matching robot (404 rethrown), and non-404 passthrough.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6df8e3c65b2ca50e981868a7b0496109c8d8cf72. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tokens-access-api.md
+++ b/.changeset/tokens-access-api.md
@@ -8,5 +8,4 @@ The `tokens create`, `tokens list` and `tokens delete` commands are now backed b
 
 - `tokens create --json`: read the secret from `.token` instead of `.key`
 - `tokens list --json`: roles now live in `.memberships[].roleNames` instead of `.roles[].name`; `createdBy`, `permissions`, `projectUserId` and `lastUsedAt` are no longer returned
-- `.id` is still the identifier to pass to `tokens delete`, but existing ids change with the backing API
 - `tokens list` shows role names instead of display titles, and no longer includes organization-managed tokens, which cannot be managed at project scope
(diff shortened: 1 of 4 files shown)


runeb:
A suggestion. The edge case it handles is someone who saved the old form of token ids and need to pass them to the cli. Using the prior major would also be an unblock for that situation.

jonahsnider (approved):

PR #1684, opened 2026-08-10
feat(assets): add local asset upload command

shapirodaniel:
### Description

Agents can create and edit Sanity content but cannot finish workflows that depend on a local or generated image without a manual Studio upload.

This adds `sanity assets upload` as a local image and file upload primitive. It streams one asset with the user's CLI authentication, reports bounded progress on stderr, and returns the uploaded asset ID plus ready-to-use reference JSON on stdout. Interactive terminals update progress in place; non-interactive callers receive a small number of meaningful progress states. Ctrl+C cancels both upload and asset-document creation and exits with code 130.

The command supports a feature-flagged MCP handoff while a presigned upload solution goes through API and security review. The command-owned telemetry policy redacts values for both `--file` and `--filename` while preserving useful flag signals such as asset type and content type. Upload errors are sanitized, and `assets:upload` is explicitly denied in MCP-hosted CLI execution.

Context: [AIGRO-5250](https://linear.app/sanity/issue/AIGRO-5250/mcp-managed-studio-decide-what-capability-gaps-to-support)

Privacy prerequisite: [CLI #1694](https://github.com/sanity-io/cli/pull/1694) / [AIGRO-5379](https://linear.app/sanity/issue/AIGRO-5379/redact-file-paths-and-filenames-from-cli-telemetry)

### What to review

- The `assets upload --file` command surface and JSON response.
- TTY and non-interactive progress behavior.
- Ctrl+C cancellation during upload and asset-document creation.
- Authentication, dataset selection, error redaction, command-owned telemetry, and MCP policy boundaries.

### Testing

- Upload command unit tests: 12 passed, including telemetry redaction coverage.
- Real image upload with telemetry debugging confirmed that `--file` and `--filename` values are absent while non-sensitive flags remain visible.
- CLI unit suite: 3,725 passed and 4 skipped. Two ANSI snapshot assertions also fail unchanged on `origin/main` under Node 26.
- CLI integration suite: 497 passed and 10 skipped.
- Format, type, lint, dependency, build, and diff checks passed.

#### happy, interactive progress
shapiro assets-upload cli

#### happy, with redacted telemetry
cli assets-upload success-with-telemetry-redacted

#### happy, with result payload
Screenshot 2026-08-12 at 6 35 25 PM

#### happy, non-interactive agent output shows progress checkpoints
Screenshot 2026-08-12 at 6 40 36 PM

#### sad, SIGINT during upload
Screenshot 2026-08-12 at 6 33 52 PM
 / document creation steps

### DIFF
--- a/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
+++ b/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
@@ -52,6 +52,16 @@ export async function uploadAssetWithProgress(options: UploadAssetWithProgressOp
   const handlesInterrupt = !executionContext
   if (handlesInterrupt) process.once('SIGINT', interruptUpload)
 
+  // Ora uses raw mode for Ctrl+C, but its input listener does not always make stdin flow.
+  const resumesInterruptInput =
+    handlesInterrupt &&
+    isTTY &&
+    uploadProgress.isSpinning &&
+    process.stdin.isTTY &&
+    process.stdin.isRaw &&
+    process.stdin.readableFlowing !== true
+  if (resumesInterruptInput) process.stdin.resume()
+
   try {
     const asset = await uploadAssetFromFile({
       ...uploadOptions,
@@ -95,5 +105,6 @@ export async function uploadAssetWithProgress(options: UploadAssetWithProgressOp
     throw error
   } finally {
     if (handlesInterrupt) process.off('SIGINT', interruptUpload)
+    if (resumesInterruptInput) process.stdin.pause()
   }
 }
--- a/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
+++ b/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
@@ -99,9 +99,7 @@ export async function uploadAssetWithProgress(options: UploadAssetWithProgressOp
   } catch (error) {
     const activeProgress = assetDocumentProgress ?? uploadProgress
     activeProgress.stop()
-    if (uploadController.signal.aborted) {
-      throw uploadController.signal.reason
-    }
+    uploadController.signal.throwIfAborted()
     throw error
   } finally {
     if (handlesInterrupt) process.off('SIGINT', interruptUpload)
--- a/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
+++ b/packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
@@ -1,0 +1,99 @@
+import {getCliExecutionContext} from '@sanity/cli-core/executionContext'
+import {spinner} from '@sanity/cli-core/ux'
+
+import {type AssetType} from '../../services/assets.js'
+import {uploadAssetFromFile} from './uploadAssetFromFile.js'
+
+interface UploadAssetWithProgressOptions {
+  assetType: AssetType
+  dataset: string
+  filename: string
+  filePath: string
+  isInteractive: boolean
+  logToStderr: (message: string) => void
+  projectId: string
+
+  contentType?: string
+}
+
+export async function uploadAssetWithProgress(options: UploadAssetWithProgressOptions) {
+  const {assetType, isInteractive, logToStderr, ...uploadOptions} = options
+  const uploadMessage = `Uploading ${assetType} asset`
+  const assetDocumentMessage = `Creating ${assetType} asset document`
+  const executionContext = getCliExecutionContext()
+  const isTTY = !executionContext && isInteractive
+  const uploadProgress = spinner(
+    isTTY
+      ? `${uploadMessage} [0%]. Large uploads may take several minutes.`
+      : `${uploadMessage}. Large uploads may take several minutes.`,
+  ).start()
+  if (executionContext) logToStderr(`${uploadMessage}. Large uploads may take several minutes.`)
+
+  let assetDocumentProgress: ReturnType<typeof spinner> | undefined
+  let assetDocumentStarted = false
+  let lastReportedProgress = 0
+  let nextNonInteractiveCheckpoint = 25
+
+  const startAssetDocument = () => {
+    if (assetDocumentStarted) return
+    assetDocumentStarted = true
+    if (!isTTY) {
+      uploadProgress.text = assetDocumentMessage
+      logToStderr(assetDocumentMessage)
+      return
+    }
+
+    uploadProgress.succeed(`${uploadMessage} [100%]`)
+    assetDocumentProgress = spinner(assetDocumentMessage).start()
+  }
+
+  const uploadController = new AbortController()
+  const interruptUpload = () => uploadController.abort(new Error('SIGINT'))
+  const handlesInterrupt = !executionContext
+  if (handlesInterrupt) process.once('SIGINT', interruptUpload)
+
+  try {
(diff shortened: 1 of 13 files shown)


packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/mcpPolicy.ts
```diff
@@ -65,6 +65,10 @@ export const mcpPolicy: CommandPolicySet = {
     validate: apiValidator,
   }),
 
+  // Reads a file from the machine running the command. The MCP asset bridge
+  // returns this command for a client to run locally; Mellon must not run it.
```
jonahsnider: this shouldn't mention the internal name of the MCP service. also nitpick but this should be a one line comment 🙃

packages/@sanity/cli/src/commands/assets/upload.ts
```diff
@@ -0,0 +1,230 @@
...
+      this.error(
+        'Asset upload failed: Cannot read the local file. Check that --file points to a readable file, then retry.',
+        {exit: exitCodes.USAGE_ERROR},
+      )
+    }
+    if (!fileStat.isFile()) {
+      this.error('Asset upload failed: --file must point to a file, not a directory.', {
+        exit: exitCodes.USAGE_ERROR,
+      })
+    }
```
jonahsnider: not to be That Guy, but could this be changed to start the upload immediately and catch any fs errors there? technically it's a race condition to check file existence and then start the upload, since it could be modified/deleted in between

packages/@sanity/cli/src/commands/assets/upload.ts
```diff
@@ -0,0 +1,230 @@
...
+    const handlesInterrupt = !executionContext
+    const readsInterruptInput = handlesInterrupt && process.stdin.isTTY
+    if (handlesInterrupt) {
+      process.once('SIGINT', interruptUpload)
+      if (readsInterruptInput) {
+        process.stdin.on('data', interruptOnInput)
+        process.stdin.resume()
+      }
+    }
+
```
jonahsnider: not a fan of how complex all this logic is for interactivity & interrupting the upload. a few comments:

* using process.exit() directly means we aren't able to run any cleanup logic
  * we should just use the AbortController to stop the upload and have the rest of the CLI code handle exiting with an error code
* ora is already re-emitting SIGINT, the manual stdin listener for it is not needed
* progress reporting is ugly since it touches every part of the upload flow here
  * probably should be moved to some kind of helper function thing that keeps all the state encapsulated, so that this command isn't the one owning all the TTY state and upload logic
* use `this.resolveIsInteractive()` instead of directly accessing `process.stderr.isTTY`
* probably the logic for opening the file handle should be moved to a helper function which is called by the command
  * this will also make the tests a lot cleaner, since we can just mock that function instead of all the individual fs calls

jonahsnider (dismissed):

packages/@sanity/cli/src/exports/invokeSanityCli/commandPolicies/mcpPolicy.ts
shapirodaniel: great catch! fixed up in 55751772

packages/@sanity/cli/src/commands/assets/upload.ts
shapirodaniel: great call, yeah this is def dicey in current state if the file changes out from under us. pivoted to proper file handle open/close behavior via `finally` block in https://github.com/sanity-io/cli/pull/1684/commits/d4c23031bda416b3918bf810f6261c52a0890480

packages/@sanity/cli/src/commands/assets/upload.ts
shapirodaniel: yeah this one was tricky, glad you called out the wonky implementation though because we can def tighten it up. high level, there’s a wrinkle with ora: bit of a wall of text incoming, but it’s interesting 🙂

ora puts `stdin` in raw mode, so Ctrl+C arrives as the control byte `0x03` instead of a native SIGINT. it then attaches a data listener to catch that byte and re-emit the interrupt, but node tracks "listenability" here through `stdin.readableFlowing`, which remains `null` because ... ora never tells node to start flowing `stdin` data

so node never starts reading, so the control byte never reaches it. feels like an ora bug that we have to call `stdin.resume()` explicitly to get around this, which is why things are a little more verbose than they should be here

tldr, i helper-ized upload progress + interruption, dropped `stderr.isTTY` in favor of `this.resolveIsInteractive()`, and scoped down `stdin.resume()` for ora's raw mode issue above

fixed up in d6a684f5, 0634b8a2

jonahsnider: wow, weird + bummer with ora but interesting. thanks for the explanation!

packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
```diff
@@ -0,0 +1,110 @@
...
+      logToStderr(`Uploaded ${assetType} asset: ${asset._id}`)
+    } else {
+      uploadProgress.succeed(`Uploaded ${assetType} asset: ${asset._id}`)
+    }
+
+    return asset
+  } catch (error) {
+    const activeProgress = assetDocumentProgress ?? uploadProgress
+    activeProgress.stop()
+    if (uploadController.signal.aborted) {
```
jonahsnider: can't do throwIfAborted?

packages/@sanity/cli/test/integration/commands/assets/upload.test.ts
```diff
@@ -0,0 +1,91 @@
...
+
+describe('#assets:upload integration', () => {
+  const projectId = 'test-project'
+  const dataset = 'production'
+  const bytes = Buffer.from([0, 1, 2, 3, 255])
+  let fixtureDirectory: string
+  let fixturePath: string
+
+  beforeAll(async () => {
+    fixtureDirectory = await mkdtemp(join(tmpdir(), 'sanity-assets-upload-'))
```
jonahsnider, line 20: trailing hyphen?

jonahsnider (dismissed):
nice!

packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
shapirodaniel: yeah this def seems doable, i'll give this a pass!

packages/@sanity/cli/test/integration/commands/assets/upload.test.ts
shapirodaniel: that's def suspect! i'll poke around

shapirodaniel: ah yeah it's a contrivance we're using to generate names like `sanity-assets-upload-${randomId}`, here and elsewhere, so i'll prob stick with it

packages/@sanity/cli/src/actions/assets/uploadAssetWithProgress.ts
shapirodaniel: yup! much nicer to use `throwIfAborted`, fixed in https://github.com/sanity-io/cli/pull/1684/commits/429f14fe03fd42f231b92bc711d2d49883b83ff0

jonahsnider (approved):

PR #1685, opened 2026-08-10
fix(cli): preserve responsive table borders

runeb:
### Description

Follow-up to #1680 for AIGRO-5358. This addresses the unresolved [Cursor Bugbot review](https://github.com/sanity-io/cli/pull/1680#discussion_r3737883018).

The responsive table helper wrapped the complete `console-table-printer` output. When `string-width` and the table printer disagree about the width of a multi-code-point character, the rendered table can be wider than the helper's estimate. Passing that completed output through `wrap-ansi` then splits rows and horizontal borders.

This change renders the table rows once without a title, derives the printer-styled title from empty tables, wraps only that title segment, and appends the untouched table. Keeping title extraction row-free prevents filter, sort, transform, and computed-column callbacks from running a second time. It keeps the table printer's existing title styling while ensuring finished borders are never reflowed. It adds no dependencies or installed weight.

Synthetic 20-column example:

Before:

```text
┌──────────┬─────
──┐
│ Emoji    │
Value │
├──────────┼─────
──┤
│ 👨‍👩‍👧‍👦 │ Cafe  │
└──────────┴─────
──┘
```

After:

```text
┌──────────┬───────┐
│ Emoji    │ Value │
├──────────┼───────┤
│ 👨‍👩‍👧‍👦 │ Cafe  │
└──────────┴───────┘
```

Note that the data row still looks short above, and that is expected rather than a leftover bug. `console-table-printer` measures this ZWJ sequence as 8 columns and pads the cell to match, while `string-width` measures it as 2, so the row renders narrow anywhere the sequence composes into a single glyph (GitHub, iTerm2, Ghostty, WezTerm). Terminals that draw the four faces separately line it up exactly.

That underlying disagreement is the cause of the bug, so it cannot be demonstrated with a character that avoids it: `string-width` and the table printer agree on CJK text and on single-code-point emoji, and those inputs never trigger the reflow. The scope here is strictly that finished borders stay on one line. Column padding for such characters, and the fact that the printer can size a table wider than this helper's estimate, remain open and predate this change.

### What to review

- Confirm only the title segment is passed through `wrap-ansi` after rendering.
- Confirm titled tables invoke stateful row callbacks only once per row.
- Confirm tables without titles and tables with wrapped titles retain their existing output.
- Confirm multi-code-point emoji cannot split the surrounding borders.
- Note the regression test pins the exact border lines rather than an overall width bound: the emoji table renders 20 columns wide against a 17-column estimate, so a width assertion would pass by coincidence.

### Testing

- Responsive table unit tests with coverage: 6 passed; 100% statements, functions, and lines.
- Affected unit tests: 122 passed across 11 files.
- Projects list integration tests: 4 passed.
- `pnpm check:format`, `pnpm check:types`, `pnpm check:lint`, `pnpm check:deps`, and repeated `pnpm build:cli` runs passed.
- The broader unit run reached 592 passing tests before the known, unchanged ANSI snapshot mismatch in `formatDocumentValidation.test.ts`.

---

> [!NOTE]
> **Low Risk**
> Display-only change in the CLI table helper with no auth, data, or API impact; behavior is narrowed to border preservation and titled-table rendering.
> 
> **Overview**
> **CLI responsive tables** no longer run the full `console-table-printer` output through `wrap-ansi`, which could break horizontal rules when `string-width` and the printer disagree on wide characters (e.g. ZWJ emoji).
> 
> Rendering now leaves the data table untouched; only the optional title is wrapped to the computed table width, using a separate printer pass so title styling stays the same without re-running row filters. Tables without a title skip post-render wrapping entirely.
> 
> Regression coverage adds a multi-code-point emoji case (borders stay on one line) and asserts titled tables with a `filter` invoke that callback only once per row.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5521380397aff922364ba2a00b079d03ffd87f26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/users-list-table-width.md
+++ b/.changeset/users-list-table-width.md
@@ -4,4 +4,4 @@
 
 fix(cli): keep table output within the terminal width
 
-CLI tables now wrap long values instead of overflowing the terminal or truncating content.
+CLI tables now wrap long values instead of truncating content and preserve borders when Unicode width estimates differ.
--- a/packages/@sanity/cli/src/util/__tests__/responsiveTable.test.ts
+++ b/packages/@sanity/cli/src/util/__tests__/responsiveTable.test.ts
@@ -73,12 +73,23 @@ describe('responsiveTable', () => {
     const lines = output.split('\n')
     expect(output).toContain('👨‍👩‍👧‍👦')
     expect(output).toContain('Cafe')
-    expect(lines[0]).toMatch(/^┌.*┐$/u)
-    expect(lines.at(-1)).toMatch(/^└.*┘$/u)
-    for (const line of lines.slice(1, -1)) {
-      expect(line).toMatch(/^(?:│.*│|├.*┤)$/u)
-      expect(stringWidth(line)).toBeLessThanOrEqual(20)
-    }
+
+    // The family emoji is a ZWJ sequence that `string-width` measures as 2 columns and
+    // `console-table-printer` as 8. That disagreement pushes the rendered table past the
+    // width this helper estimated, which is what previously let `wrap-ansi` reflow the
+    // finished output and split the horizontal rules across lines.
+    //
+    // The guarantee under test is that every border survives on a single line. Note that
+    // the table is 20 columns wide while the helper estimated 17, so the disagreement is
+    // still visible: the data row pads to the printer's 8-column reading of the emoji and
+    // renders short wherever the sequence composes into a single glyph. Asserting an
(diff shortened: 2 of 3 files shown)


jonahsnider (approved):

runeb:
Thanks jonah. updating the PR body a bit for posterity but otherwise no need to re-review this

PR #1686, opened 2026-08-10
fix(ci): publish behind a holding tag until packages are installable

rexxars:
### Description

**Before:** the release workflow publishes straight to `latest`. npm [scans new versions for malware at publish time](https://github.blog/changelog/2026-07-28-npm-publish-time-malware-scanning-and-dual-use-metadata/) and holds each one for roughly 5-15 minutes before it resolves on the registry. Our packages pin each other, so when one package's scan clears before a dependency's, `latest` points at a version that cannot be installed. `create-sanity` pins `@sanity/cli` exactly (`workspace:*`), so this breaks `npm create sanity`, and the `@sanity/cli*` packages pin each other with caret ranges. This is what bit the sanity monorepo during v6.8.0 (sanity-io/sanity#13823).

**After:** `release.yml` publishes everything behind a temporary `pending` dist-tag, polls until every published version resolves, then moves `latest`. During the hold `latest` keeps pointing at the previous, mutually consistent set of versions. If the wait times out, `latest` is never moved: the release is invisible rather than broken.

This is deliberately not a straight port of sanity-io/sanity#13823. That PR gates downstream jobs so the release is not *announced* before it is installable, which works there because announcement lives in separate jobs. Here `changesets/action` creates the GitHub releases inside its own step, so the only downstream job is `post-release` (Linear issue bookkeeping). A gate alone would just hold the workflow open without protecting anyone, so this moves the protection to the dist-tag instead.

Three commits:

- `20bfc978c` publish behind the holding tag, wait, then promote
- `3c7436392` add `promote-latest.yml` and extract the shared logic into a composite action
- `fab229b5f` document both in CONTRIBUTING.md

**Why the recovery workflow:** if the release job dies after publishing, the versions are on npm but still parked on `pending`, and re-running `release.yml` does *not* finish the job. Changesets sees them as already published, reports nothing as newly published, and skips promotion. `promote-latest.yml` is a `workflow_dispatch` that reads the versions back off the `pending` tag and promotes them. It takes no input, so it cannot promote anything the release workflow did not park there.

**Why a token instead of OIDC:** npm's trusted publishing [only covers `npm publish`](https://docs.npmjs.com/trusted-publishers/), not `npm dist-tag`. That gap is [npm/cli#8547](https://github.com/npm/cli/issues/8547), open since September 2025 with no response from npm; the workarounds people tried (`electron/npm-trusted-auth-action`, calling the dist-tags endpoint directly) are archived or return 401. So the promote step uses `NPM_PUBLISH_TOKEN`, the same secret `snapshot-release.yml` already publishes with. Publishing itself still uses OIDC, and the token is only written to `.npmrc` after the publish step so it cannot interfere with the OIDC exchange.

### What to review

`.github/workflows/release.yml` is the important one. Worth checking:

- **The `--tag` forwarding.** `publish:` is now `pnpm run publish-packages --tag pending`, built by the new "Determine publish mode" step. I first wrote it with a `--` separator, which pnpm passes through literally, so changesets' meow parser would have treated `--tag pending` as positional input and silently published to `latest` anyway. Verified the current form forwards cleanly on pnpm 10.34.4, the version pinned in `packageManager`.
- **The prerelease guard.** In prerelease mode changesets picks its own dist-tag from `.changeset/pre.json`, so forcing `--tag pending` would clobber it and promoting would put a prerelease on `latest`. The step skips both. `pre.json` has never appeared in this repo's history, so this path is untested by construction, but the failure mode is bad enough to guard.
- **Holding tag choice.** `pending`, picked after checking existing dist-tags on these packages. `next`, `canary`, `stable`, `alpha`, `beta` and `rc` are all in use.
- **Whether GitHub releases should also be gated.** They are still created by `changesets/action` inside its own step, so for a few minutes they announce a version that is not on `latest` yet. Gating that means `createGithubReleases: false` plus a separate step, which I left out as a separate decision.

Known caveats: the release job now holds a runner for the length of the scan hold, and npm sets `latest` on a package's first-ever version regardless of `--tag` (harmless, nothing depends on a prior version).

### Testing

Manual. This is CI workflow bash, not practical to unit test. I extracted each `run:` block and executed it in isolation:

- **Wait loop, live registry:** all specs resolve and it exits 0; with an injected `@sanity/cli@99.99.99` and a shortened deadline it retries, then fails with the missing-package list and exits 1. Empty input is rejected.
- **Publish mode:** both branches produce the expected `$GITHUB_OUTPUT`, and arg forwarding verified on both pnpm 10.34.4 and 11.5.2 with a script that prints raw argv.
- **Promote, stubbed `npm`:** correct `dist-tag` commands, scoped-name parsing works (`@sanity/cli@7.19.0` to `@sanity/cli`), a failed `dist-tag rm` warns without failing the step, a failed `dist-tag add` exits 1. `.npmrc` ends up containing the literal `${NODE_AUTH_TOKEN}` placeholder, not the token value.
- **Discovery in `promote-latest.yml`:** against the real registry today it correctly reports nothing to promote and exits 1; with a stub it picks up exactly the parked packages.
- All three YAML files parse; step order and `if` conditions confirmed.
- `pnpm check:format` passes. `pnpm check:deps` fails identically on `main` (unlisted `get-it` in a `dist/` file), so it is pre-existing and unrelated.

Not verified: an end-to-end run against a real release, since that needs an actual publish inside the scan window.

### Notes for release

N/A - Internal only (CI)

---

> [!NOTE]
> **Medium Risk**
> Changes npm release semantics and dist-tags for all public packages; misconfiguration could leave `latest` stale or promote wrong versions, though prerelease guards and registry-only recovery limit blast radius.
> 
> **Overview**
> **Release CI** no longer publishes straight to `latest`. Stable releases go out under a temporary **`pending`** dist-tag while npm finishes publish-time malware scanning; **`latest` stays on the previous release** until every published `name@version` resolves on the registry, then promotion moves `latest` and removes `pending`.
> 
> `release.yml` adds **Determine publish mode** (prerelease via `.changeset/pre.json` skips `--tag pending` and promotion), forwards **`--tag pending`** into `pnpm run publish-packages`, and after changesets publishes runs a new **Promote packages to latest** step. Promotion uses **`NPM_PUBLISH_TOKEN`** because OIDC trusted publishing does not cover `npm dist-tag`.
> 
> Shared logic lives in **`.github/actions/promote-npm-packages`**: poll `npm view` until specs are installable (default 20 min), then `dist-tag add … latest` and `dist-tag rm` the holding tag.
> 
> **`promote-latest.yml`** is a manual **break-glass** workflow: discovers versions parked on `pending` from the registry and runs the same composite action—needed when release fails after publish (re-running release skips promotion because changesets reports nothing new).
> 
> **CONTRIBUTING.md** documents the `pending` tag, token requirement, and recovery steps.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fab229b5f6a2c4eb16cc3dc9f21bc79cf72b73e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/actions/promote-npm-packages/action.yml
+++ b/.github/actions/promote-npm-packages/action.yml
@@ -1,0 +1,95 @@
+name: 'Promote npm packages'
+description: 'Waits for published versions to clear npm scanning, then moves the latest dist-tag onto them'
+
+inputs:
+  specs:
+    description: 'Newline-separated name@version specs to promote'
+    required: true
+  holding-tag:
+    description: 'Dist-tag the versions were published under, removed once they are promoted'
+    required: true
+  npm-token:
+    description: >-
+      npm token with publish rights. OIDC cannot be used here: trusted publishing only
+      covers `npm publish`, not `npm dist-tag` (https://github.com/npm/cli/issues/8547).
+    required: true
+  timeout-minutes:
+    description: 'How long to wait for every version to become installable'
+    default: '20'
+
+runs:
+  using: composite
+  steps:
+    - name: Wait for packages to be installable
+      # npm holds newly published versions for malware scanning (usually 5-15
+      # minutes, sometimes longer) before they resolve on the registry. Poll
+      # until the whole set resolves, so `latest` is only moved onto versions
+      # that can actually be installed.
+      # https://github.blog/changelog/2026-07-28-npm-publish-time-malware-scanning-and-dual-use-metadata/
+      shell: bash
+      env:
+        SPECS: ${{ inputs.specs }}
+        TIMEOUT_MINUTES: ${{ inputs.timeout-minutes }}
+      run: |
+        mapfile -t pending < <(grep -v '^[[:space:]]*$' <<<"$SPECS")
+
+        if [ "${#pending[@]}" -eq 0 ]; then
+          echo "Error: no package specs were given"
+          exit 1
+        fi
+
+        echo "Waiting for ${#pending[@]} package(s) to resolve on the registry: ${pending[*]}"
+
+        deadline=$((SECONDS + TIMEOUT_MINUTES * 60))
+        while true; do
+          still_pending=()
+          for spec in "${pending[@]}"; do
+            if resolved=$(npm view "$spec" version --prefer-online 2>/dev/null) && [ -n "$resolved" ]; then
+              echo "✅ $spec is installable"
+            else
+              still_pending+=("$spec")
+            fi
+          done
+          pending=("${still_pending[@]}")
+
+          if [ "${#pending[@]}" -eq 0 ]; then
+            echo "✅ All packages are installable"
+            break
+          fi
+
+          if [ "$SECONDS" -ge "$deadline" ]; then
+            echo "Error: timed out waiting for these packages to become installable:"
+            printf ' - %s\n' "${pending[@]}"
+            exit 1
+          fi
+
+          echo "⏳ Still waiting on ${#pending[@]} package(s): ${pending[*]} (retrying in 15s)"
+          sleep 15
+        done
+
+    - name: Move the latest tag
+      shell: bash
+      env:
+        SPECS: ${{ inputs.specs }}
+        HOLDING_TAG: ${{ inputs.holding-tag }}
+        NODE_AUTH_TOKEN: ${{ inputs.npm-token }}
+      run: |
+        # Single-quoted so the token is resolved from the environment at run
+        # time rather than written to disk
+        echo '//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}' >> .npmrc
+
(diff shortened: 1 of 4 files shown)


mwritter:
I think this fixes our side of it, but the e2e init failures we're looking at is come from third-party packages mid-scan, not ours.

The one we caught was `@typescript-eslint/scope-manager@8.67.0`, published about 3 minutes after`typescript-eslint@8.67.0`, and `sanity init` pulls that chain in through `@sanity/eslint-config-studio`. Nothing in our release pipeline can help there, so those will keep happening after this lands.

its like we need `typescript-eslint` to implement this same fix 😆

mwritter (approved):
I think we need this for our side - but we may need to look safeguarding `installPackages.js` (if possible)

PR #1687, opened 2026-08-10
test: isolate env vars from interfering with tests

jonahsnider:
I noticed that tests for me would fail despite working fine in CI. Turns out it's because I have `SANITY_INTERNAL_ENV` set to `staging` in my shell, which was breaking a bunch of tests here.

This adds a setup script to all the Vitest configs to remove certain Sanity env vars that could be problematic and interfere with tests.

---

> [!NOTE]
> **Low Risk**
> Test-only harness changes with no production runtime impact; minor risk that tests relied on host env vars no longer being present.
> 
> **Overview**
> Adds a shared **`test/vitest/setup.ts`** that runs before tests: it **removes** a fixed list of Sanity-related `process.env` keys that can leak from a developer shell, then sets **`SANITY_INTERNAL_ENV`** to **`production`** so runs match CI.
> 
> That setup file is wired into **`setupFiles`** on Vitest configs across CLI-related packages (unit, integration, and e2e), including **`create-sanity`**. **`@sanity/cli`** keeps its local **`test/setup.ts`** but now loads the shared setup **first**.
> 
> The **`injectEnvVariables`** integration test clears **`SANITY_INTERNAL_ENV`** before asserting it is loaded from **`.env`**, so the new default does not mask the hook under test.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ef27a9939c661eed0d75fa7ee520bfc390bd27d5. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@repo/coverage-delta/vitest.config.ts
+++ b/packages/@repo/coverage-delta/vitest.config.ts
@@ -4,5 +4,6 @@ export default defineConfig({
   test: {
     environment: 'node',
     globals: false,
+    setupFiles: ['../../../test/vitest/setup.ts'],
   },
 })
--- a/packages/@sanity/cli-build/vitest.config.ts
+++ b/packages/@sanity/cli-build/vitest.config.ts
@@ -18,5 +18,6 @@ export default defineConfig({
     exclude: ['**/node_modules/**', '**/dist/**', 'test/integration/**'],
     globals: false,
     name: '@sanity/cli-build/unit',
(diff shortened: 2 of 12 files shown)


shapirodaniel (dismissed):
nice!

shapirodaniel (approved):
looks great!

PR #1689, opened 2026-08-10
fix(cli): resolve command topic aliases in programmatic CLI invocations

jonahsnider:
`invokeSanityCLI` (used by `run_sanity_cli`) was not using the same command topic alias resolution as the regular CLI. This meant commands that would work locally were failing when called via `run_sanity_cli`

This change:
- Centralizes alias resolution to a single file
- Resolves topic aliases in `invokeSanityCli`

---

> [!NOTE]
> **Low Risk**
> Localized CLI argument normalization with broad test coverage; no changes to auth, data handling, or command execution semantics beyond correct routing for aliases.
> 
> **Overview**
> **Programmatic CLI** (`invokeSanityCli` / MCP) now rewrites singular topic aliases (e.g. `hook`, `hook:list`) to canonical names before command dispatch and help, matching the interactive CLI behavior that previously only ran through the `command_not_found` hook.
> 
> Alias logic is **centralized** in `topicAliases.ts`: new `resolveTopicAlias` handles bare topics and colon-separated command IDs; `resolveTopicAliasInArgv` moves there from `SanityHelp`. The `command_not_found` hook drops its local alias map and calls `resolveTopicAlias` instead. Policy-scoped help (`InvokableHelp.showHelp`) applies the same argv rewrite.
> 
> Tests move to `topicAliases.test.ts` (including colon-form argv) and integration cases cover alias resolution for MCP policy and help.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fb8e54f7910f897e5c465d85b5bd6e3f6c5a47d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli/src/exports/invokeSanityCli/index.ts
+++ b/packages/@sanity/cli/src/exports/invokeSanityCli/index.ts
@@ -27,6 +27,7 @@ import {runWithCliExecutionContext, type SanityEnvironment} from '@sanity/cli-co
 import {type SanityCommand} from '@sanity/cli-core/SanityCommand'
 import {parseArgsStringToArgv} from 'string-argv'
 
+import {resolveTopicAliasInArgv} from '../../topicAliases.js'
 import {commandPolicies} from './commandPolicies/index.js'
 import {
   type CommandPolicySet,
@@ -201,6 +202,7 @@ async function invokeSanityCliInContext(
       ? parseArgsStringToArgv(args).map((t) => stripFlagQuotes(t))
       : [...args]
   if (argv[0] === 'sanity') argv = argv.slice(1)
+  argv = resolveTopicAliasInArgv(argv)
 
   // Help requests are routed through oclif's help system, scoped to the
   // source's policy: root help for a bare request, topic/command help when a
--- a/packages/@sanity/cli/src/hooks/commandNotFound/topicAliases.ts
+++ b/packages/@sanity/cli/src/hooks/commandNotFound/topicAliases.ts
@@ -26,16 +26,16 @@ const hook: Hook.CommandNotFound = async function (opts) {
   // Standardize the ID (handles topic separator differences)
   const standardId = toStandardizedId(id, config)
   const parts = standardId.split(':')
+  const hasTopic = parts.length === 1
   const rewrittenId = resolveTopicAlias(standardId)
 
   const aliasExists =
-    rewrittenId &&
-    (parts.length === 1 ? config.findTopic(rewrittenId) : config.findCommand(rewrittenId))
+    rewrittenId && (hasTopic ? config.findTopic(rewrittenId) : config.findCommand(rewrittenId))
   if (aliasExists) {
     debug('Rewriting topic: %s -> %s', parts[0], rewrittenId.split(':')[0])
 
     // Bare topic (eg "sanity dataset" with no subcommand) -> show topic help
-    if (parts.length === 1) {
+    if (hasTopic) {
       return config.runCommand('help', [rewrittenId])
     }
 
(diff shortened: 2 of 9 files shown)


packages/@sanity/cli/src/exports/invokeSanityCli/index.ts
```diff
@@ -201,6 +202,7 @@ async function invokeSanityCliInContext(
       ? parseArgsStringToArgv(args).map((t) => stripFlagQuotes(t))
       : [...args]
   if (argv[0] === 'sanity') argv = argv.slice(1)
+  argv = resolveTopicAliasInArgv(argv)
```
shapirodaniel, line 205: 🔥

packages/@sanity/cli/src/hooks/commandNotFound/topicAliases.ts
```diff
@@ -16,55 +16,31 @@
...
-  if (candidates) {
-    // Find which candidate topic actually has commands registered
-    const resolvedTopic = candidates.find((candidate) =>
-      parts.length === 1
-        ? config.findTopic(candidate)
-        : config.findCommand([candidate, ...parts.slice(1)].join(':')),
-    )
+  const aliasExists =
+    rewrittenId &&
+    (parts.length === 1 ? config.findTopic(rewrittenId) : config.findCommand(rewrittenId))
```
shapirodaniel: super nit, maybe `const hasTopic = parts.length === 1` since we use it in a few spots? seems pre-existing so no strong pref

packages/@sanity/cli/src/topicAliases.ts
```diff
@@ -30,3 +32,47 @@ export const topicAliases: Record<string, string[]> = {
...
+
+const canonicalTopicByAlias = new Map(
+  Object.entries(topicAliases).flatMap(([canonical, aliases]) =>
+    aliases.map((alias) => [alias, canonical]),
+  ),
+)
+
+/**
+ * Resolve the topic portion of a command ID from an alias to its canonical name.
+ * Handles both bare topics (`hook`) and colon-separated command IDs (`hook:list`).
```
shapirodaniel, line 44: nice!

packages/@sanity/cli/src/topicAliases.ts
```diff
@@ -30,3 +32,47 @@ export const topicAliases: Record<string, string[]> = {
...
+  const topic = separatorIndex === -1 ? id : id.slice(0, separatorIndex)
+  const canonical = canonicalTopicByAlias.get(topic)
+  if (!canonical) return undefined
+
+  const suffix = separatorIndex === -1 ? '' : id.slice(separatorIndex)
+  return `${canonical}${suffix}`
+}
+
+/**
+ * Resolve the first positional topic in argv, preserving flags and arguments.
```
shapirodaniel, line 59: seems like a good parser, curious if there's a lib for argv parsing? though otoh we're trying to shrink the package size overall, so this is better from a dependency bloat perspective

shapirodaniel (dismissed):
looks great! minor qs / nits

packages/@sanity/cli/src/topicAliases.ts
jonahsnider: i'm being told no, not really. we don't know the flags ahead of time, so we can't use `node:util` `parseArgs()`. oclif has some parser utils but they also need us to specify what the args are ahead of time

allegedly we can alleviate some of this complexity by leveraging native oclif alias features (i have not investigated this at all but seems reasonable that this is some new-ish feature we could be using). i think keep this as-is for now regrettably

shapirodaniel: makes sense! def agree on holding off on oclif aliases for now, sounds interesting / future nice-to-have 😄

shapirodaniel (approved):
nice!

PR #1690, opened 2026-08-10
feat: include source of auth token in `sanity debug`

jonahsnider:
`sanity debug` now includes the source of the auth token (ex. env var, config file, etc.)

---

> [!NOTE]
> **Low Risk**
> Auth resolution behavior for `getCliToken()` is preserved; the change mainly adds metadata for debugging and extends the internal cache shape with matching test coverage.
> 
> **Overview**
> **`sanity debug`** now prints a **Token source** line in the Authentication section so you can see whether the active credential came from the execution context, `SANITY_AUTH_TOKEN`, or a config file path (including `SANITY_CLI_CONFIG_PATH`).
> 
> In **`@sanity/cli-core`**, token resolution is refactored around a new **`getCliTokenInfo()`** API and **`CliTokenInfo`** (`{ token, source }`). **`getCliToken()`** still returns only the string and now delegates to `getCliTokenInfo()`. The in-process token cache stores the full `CliTokenInfo` so repeated lookups keep the original source label.
> 
> Both symbols are exported from the public **`@sanity/cli-core`** entry points; tests cover source labeling and debug output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fa5a032a4d80a56f82abc557e421a9f35f3091fa. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/@sanity/cli-core/src/config/cli/cliTokenCache.ts
+++ b/packages/@sanity/cli-core/src/config/cli/cliTokenCache.ts
@@ -1,22 +1,27 @@
 /**
- * Module-level cache for the CLI auth token, shared between
- * `getCliToken` (reads/writes) and `setCliUserConfig` (invalidates).
+ * Module-level cache for the resolved CLI auth token, shared between
+ * `getCliTokenInfo` (reads/writes) and `setCliUserConfig` (invalidates).
  *
  * Extracted into its own module to avoid a circular dependency
  * between `cliUserConfig.ts` and `getCliToken.ts`.
  */
-let cachedToken: string | undefined
+export interface CliTokenInfo {
+  source: string
+  token: string
+}
 
-export function getCachedToken(): string | undefined {
+let cachedToken: CliTokenInfo | undefined
+
+export function getCachedToken(): CliTokenInfo | undefined {
   return cachedToken
 }
 
-export function setCachedToken(token: string | undefined): void {
+export function setCachedToken(token: CliTokenInfo | undefined): void {
   cachedToken = token
 }
 
 /**
- * Clear the in-process token cache so the next `getCliToken()` call
+ * Clear the in-process token cache so the next token resolution
  * re-reads from disk or the environment.
  *
  * Called automatically by `setCliUserConfig('authToken', ...)`.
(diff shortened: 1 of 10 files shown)


packages/@sanity/cli-core/src/config/cli/cliTokenCache.ts
```diff
@@ -1,22 +1,27 @@
...
- * Module-level cache for the CLI auth token, shared between
- * `getCliToken` (reads/writes) and `setCliUserConfig` (invalidates).
+ * Module-level cache for the resolved CLI auth token, shared between
+ * `getCliTokenInfo` (reads/writes) and `setCliUserConfig` (invalidates).
  *
  * Extracted into its own module to avoid a circular dependency
  * between `cliUserConfig.ts` and `getCliToken.ts`.
  */
-let cachedToken: string | undefined
+export interface CliTokenInfo {
```
shapirodaniel: is it possible to decouple the addition of token source provenance from the token itself so we don't have to change the contract of what `cachedToken` is?

packages/@sanity/cli-core/src/config/cli/cliUserConfig.ts
```diff
@@ -40,14 +60,19 @@ export async function getCliToken(): Promise<string | undefined> {
...
   const token = process.env.SANITY_AUTH_TOKEN
   if (token) {
-    const trimmed = token.trim()
-    setCachedToken(trimmed)
-    return trimmed
+    const tokenInfo = {
+      source: 'SANITY_AUTH_TOKEN environment variable',
+      token: token.trim(),
+    }
+    setCachedToken(tokenInfo)
```
shapirodaniel: similar to the comment above, `setCachedToken` with `tokenInfo` is a bit difficult to reason through, if we're unable to decouple token provenance from the token itself, maybe we update getter/setter methods and global vars to indicate they now handle token info rather than a token?

packages/@sanity/cli-core/src/config/cli/cliTokenCache.ts
jonahsnider: this was actually my original approach, but i was spooked by the idea of the cached token becoming out of sync with its source. IMO keeping them coupled is the right thing to do, it makes working with it a single atomic thing

shapirodaniel: great call out!

shapirodaniel (dismissed):
looks great!

shapirodaniel (approved):
nice!

PR #1694, opened 2026-08-11
fix(cli): add command-owned telemetry redaction

shapirodaniel:
### Description

Add an opt-in, command-owned way to redact selected flag values from CLI telemetry. No existing command opts in, so current telemetry remains unchanged.

This prepares for assets upload: that command can redact its own local-file inputs without changing telemetry for unrelated commands that use the same flag name.

Issue: [AIGRO-5379](https://linear.app/sanity/issue/AIGRO-5379/redact-file-paths-and-filenames-from-cli-telemetry)  
Related: [CLI #1684](https://github.com/sanity-io/cli/pull/1684), [Mellon #669](https://github.com/sanity-io/mellon/pull/669)

### Examples

In a future assets upload command, we might define a `--filename` flag like:

```ts
const flags = {
  file: Flags.string({
    aliases: ['filename'],
    char: 'f',
    description: 'Path to the asset',
  }),
  format: Flags.string({
    options: ['json', 'text'],
  }),
}
```

When we create the command's `telemetry` recorder, we'd supply the `redact` option with a list of in-scope flags:

```ts
export class AssetsUploadCommand extends SanityCommand {
  static flags = flags

  static telemetry = defineCommandTelemetry(flags, {
    redact: ['file'],
  })
}
```

When we run:

```sh
sanity assets upload \
  --file=/Users/jane/Clients/Acme/private.png \
  --format=json
```

Without `redact`, we'd see this telemetry output:

```json
{
  "extraArguments": [
    "--file=/Users/jane/Clients/Acme/private.png",
    "--format=json"
  ]
}
```

After supplying `{ redact: ['file'] }`, we'd see this output instead:

```json
{
  "extraArguments": ["--file", "--format=json"]
}
```

Redaction is command-specific. Other commands that use the same flag name aren't affected:

```sh
sanity documents validate --file=/Users/jane/private.ndjson
```

Unless `documents validate` independently opts in, its telemetry remains:

```json
{
  "extraArguments": ["--file=/Users/jane/private.ndjson"]
}
```

### What to review

- `defineCommandTelemetry` type-checks redacted names and resolves canonical names, long aliases, and short forms from the owning command's flag definitions.
- `parseArguments` consumes the resolved telemetry contract rather than reconciling flag definitions and redaction names independently.
- Commands without a redaction declaration preserve the existing telemetry payload.

### Testing

- Unit tests cover unchanged defaults, canonical names, long aliases, `char`, `charAliases`, single-character aliases, short/long option boundaries, and forwarded arguments.
- Telemetry integration tests cover unchanged defaults and explicit command-owned redaction.
- Format, types, lint, dependency checks, and CLI build pass.

### Notes for release

Add command-owned configuration for redacting CLI flag values from telemetry.

### DIFF
--- a/packages/@sanity/cli/src/util/parseArguments.ts
+++ b/packages/@sanity/cli/src/util/parseArguments.ts
@@ -36,6 +36,52 @@ interface ParsedArguments<F = Record<string, string>> {
   groupOrCommand: string
 }
 
+const SENSITIVE_TELEMETRY_OPTIONS = ['--file', '--filename'] as const
+
+function getSensitiveTelemetryOption(argument: string): string | undefined {
+  return SENSITIVE_TELEMETRY_OPTIONS.find(
+    (option) => argument === option || argument.startsWith(`${option}=`),
+  )
+}
+
+function redactSensitiveTelemetryValues(arguments_: string[]): string[] {
+  const redactedArguments: string[] = []
+
+  for (let index = 0; index < arguments_.length; index += 1) {
+    const argument = arguments_[index]
+    const sensitiveOption = getSensitiveTelemetryOption(argument)
+
+    if (!sensitiveOption) {
+      redactedArguments.push(argument)
+      continue
+    }
+
+    redactedArguments.push(sensitiveOption)
+    if (argument === sensitiveOption) index += 1
+  }
+
+  return redactedArguments
+}
+
+function collectOptionArguments(arguments_: string[]): string[] {
+  const optionArguments: string[] = []
+
+  for (let index = 0; index < arguments_.length; index += 1) {
+    const argument = arguments_[index]
+    const sensitiveOption = getSensitiveTelemetryOption(argument)
+
+    if (sensitiveOption) {
+      optionArguments.push(sensitiveOption)
+      if (argument === sensitiveOption) index += 1
+      continue
+    }
+
+    if (argument.startsWith('-')) optionArguments.push(argument)
+  }
+
+  return optionArguments
+}
+
 /**
  * Parse the arguments from the command line
  *
@@ -58,9 +104,12 @@ export function parseArguments(argv = process.argv): ParsedArguments {
 
   const [groupOrCommand, ...argsWithoutOptions] = _
 
+  const argumentSeparatorIndex = args.indexOf('--')
+  const argumentsBeforeSeparator =
+    argumentSeparatorIndex === -1 ? args : args.slice(0, argumentSeparatorIndex)
   const finalExtraArguments = [
-    ...(extraArguments || []),
-    ...argv.filter((arg) => arg.startsWith('-')),
+    ...redactSensitiveTelemetryValues(extraArguments || []),
+    ...collectOptionArguments(argumentsBeforeSeparator),
   ]
 
   // oclif allows to run `sanity help` or `sanity help <command>`
--- a/packages/@sanity/cli/src/util/parseArguments.ts
+++ b/packages/@sanity/cli/src/util/parseArguments.ts
@@ -1,7 +1,6 @@
-import {type Interfaces, Parser} from '@oclif/core'
 import minimist from 'minimist'
 
-import {type CommandTelemetry} from './telemetry/commandTelemetry.js'
+import {type CommandTelemetry, redactTelemetryArguments} from './telemetry/commandTelemetry.js'
 
 interface ParsedArguments<F = Record<string, string>> {
   /**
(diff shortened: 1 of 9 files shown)


packages/@sanity/cli/src/util/parseArguments.ts
```diff
@@ -36,6 +36,52 @@ interface ParsedArguments<F = Record<string, string>> {
   groupOrCommand: string
 }
 
+const SENSITIVE_TELEMETRY_OPTIONS = ['--file', '--filename'] as const
```
jonahsnider: are there not other sensitive parameters anywhere in the CLI? i'd be surprised if this was the first instance of something that's especially linked to user data. feels like we should either be tracking this like normal, or ignoring a whole bunch of other params

jonahsnider: instead of declaring params as sensitive here, is it possible to add something to the command registration itself which defines the sensitive flags? that way we can handle things like aliases easier, plus it scales better for lots of commands

shapirodaniel: great call out! there's def pre-existing sensitive params in current state: `sanity api` and `sanity tokens` are good examples. i agree with your take here + below, better to solve the whole problem now and put good AGENTS.md guidance in place to allow teams to opt-in to capturing inputs in the future for triage / debugging needs

i'll push a follow up commit to get telemetry redaction in place as a framework capability

shapirodaniel: love this idea, and the tradeoff of wider telemetry redaction seems justified given the first pass was too targeted to become a framework capability

updated in https://github.com/sanity-io/cli/pull/1694/commits/b3e2d29a915b812078837b5e90825fe6d275d4ff

shapirodaniel: following up here, in offline sync we're aligned that an opt-in approach is safer for a first pass here: it'll let us redact inputs to the assets-upload-specific api we're about to introduce, and gives teams who depend on these telemetry inputs time to weigh in. i'll have a commit to swap the machinery shortly

shapirodaniel: updated to make command telemetry redaction opt-in in https://github.com/sanity-io/cli/pull/1694/commits/d513bc78fd5766871d8806b32cb8218280ec8013

packages/@sanity/cli/src/util/parseArguments.ts
```diff
@@ -36,15 +39,75 @@ interface ParsedArguments<F = Record<string, string>> {
...
+  flags: Interfaces.FlagInput,
+  commandTelemetry: CommandTelemetry,
+): RedactedFlagSyntax[] {
+  return (commandTelemetry.redact ?? []).map((name) => {
+    const flag = flags[name]
+
+    return {
+      canonical: `--${name}`,
+      long: [name, ...(flag.aliases ?? [])].map((alias) => `--${alias}`),
+      short: [flag.char, ...(flag.charAliases ?? [])].filter(Boolean).map((alias) => `-${alias}`),
```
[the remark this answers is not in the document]
shapirodaniel: nice call out, better to be consistent with oclif short flag expectations. fixed in https://github.com/sanity-io/cli/pull/1694/commits/eb7501dadb1b488eb46ab3f6ad472395f4f930ba

packages/@sanity/cli/src/util/telemetry/commandTelemetry.ts
```diff
@@ -0,0 +1,72 @@
+import {type Interfaces} from '@oclif/core'
+
+interface RedactedFlag {
+  canonical: string
+  long: readonly string[]
+  short: readonly string[]
```
jonahsnider: for the sake of simplifying this file's logic, could we lump `long` and `short` into a single `flags: string[]`? does it change any of the redaction logic to do that? seems like we'd be able to trim some of the stuff that sorts them into one or the other, without impacting the actual end result

shapirodaniel: love this idea! i think it's possible, let me give it a pass and see what shakes out

shapirodaniel: def nicer without sorting by short/long option, great suggestion! fixed up in https://github.com/sanity-io/cli/pull/1694/commits/5c50829954f86bd1c8bc85069a5603b22fa42b2c

jonahsnider: nice!

jonahsnider (approved):

